PR #255, opened 2026-06-23
Fix OMH quality-loop routing intent

rlaope:
## Summary

This PR fixes the routing-quality regression found by the OMH self-improvement loop after PR #254 was merged and installed locally.

The live finding was that a request about improving OMH router/context/coding-handoff quality was routed primarily to `feedback-triage` because it contained the Korean word `버그`. That exposed a deeper issue: routing treated generic bug keywords as primary selectors instead of semantic evidence inside the whole request intent.

What changed:
- Adds deterministic `classify_omh_quality_intent` in `src/routing/intent.py`.
- Connects the same semantic intent layer to plugin/context route hints and package `omh recommend` guardrails.
- Routes OMH/router/context/progress/coding-handoff quality loops to the coding/process lane (`ultraprocess`) even when bug/failure words appear.
- Keeps actual customer/product bug reports routed to `feedback-triage`.
- Avoids substring false positives such as treating `OMHM` as `OMH`.
- Adds Korean/English paraphrase regressions plus customer-bug counterexamples.

## Why

The fix is intentionally not a keyword-ordering patch. Keywords like `bug` / `버그` remain useful evidence, but they should not override the semantic subject and goal of the request. OMH self-improvement loops need to be recognized as router/context/handoff quality work; customer bug reports still need feedback triage.

## Verification

Observed by Codex:

- `UV_CACHE_DIR=/tmp/omhm-routing-safety-uv-cache PYTHONPATH=tests uv run python -m unittest discover -s tests -v` — passed, 713 tests
- `UV_CACHE_DIR=/tmp/omhm-routing-safety-uv-cache uv run python -m compileall -q src tests` — passed
- `git diff --check` — passed

## Remaining risk

Live installed Hermes/OMH plugin reload has not been tested after this commit. After merge, run `omh setup --default-executor codex --force`, `omh doctor`, and re-run the same live route/context prompt.

### DIFF
--- a/src/plugin_bundle/omh/awareness.py
+++ b/src/plugin_bundle/omh/awareness.py
@@ -50,6 +50,31 @@ except ImportError:  # pragma: no cover - exercised by standalone plugin hosts.
                 *(f"handoff:{cue}" for cue in self.handoff_cues),
             )
 
+    def _fallback_contains_non_ascii(value: str) -> bool:
+        return any(ord(char) > 127 for char in value)
+
+    def _fallback_contains_bounded_english_cue(text: str, cue: str) -> bool:
+        parts = re.findall(r"[a-z0-9]+", cue)
+        if not parts:
+            return False
+        separator = r"[\s_-]+"
+        pattern = r"(?<![a-z0-9])" + separator.join(re.escape(part) for part in parts) + r"(?![a-z0-9])"
+        return re.search(pattern, text) is not None
+
+    def _fallback_matched_omh_quality_cues(cues: tuple[str, ...], text: str, compact: str) -> tuple[str, ...]:
+        matches = []
+        for cue in cues:
+            normalized_cue = unicodedata.normalize("NFKC", cue).casefold()
+            if not normalized_cue:
+                continue
+            if _fallback_contains_non_ascii(normalized_cue):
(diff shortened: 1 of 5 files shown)


rlaope (commented):
Reviewed by Friren bot.

Finding (medium): `classify_omh_quality_intent()` still treats short cue substrings as semantic evidence, so ordinary OMH/router reference questions can be misrouted to `ultraprocess`.

Evidence: `_matched_cues()` uses substring matching (`normalized_cue in normalized` / compact contains) for every cue, including the short improvement cue `fix`. That makes words like `prefix` count as `fix`.

Repro from this branch:

```py
from omh.routing.intent import classify_omh_quality_intent
classify_omh_quality_intent('What is the OMH router prefix syntax?').applies
# True
```

That request is a router/reference question, not an OMH quality-improvement loop. Because it has OMH subject + router domain + accidental `fix` inside `prefix`, the new guard can steal it into `ultraprocess`.

Suggested fix: require token/phrase-boundary matching for short English cues such as `fix`, `bug`, `loss`, `keep`, etc., or give `classify_omh_quality_intent` its own cue matcher that treats single-word English cues as tokens rather than substrings. Add a counter-regression like `What is the OMH router prefix syntax?` that must not produce `omh_quality_improvement_loop`.

PR #771, opened 2026-08-02
test: make POSIX-permission and symlink tests Windows-compatible

mrmixx-max:
### What Changed

Two test-only fixes make the OMH test suite compatible with Windows and
non-elevated environments where POSIX-only capabilities are unavailable:

1. **`tests/test_awareness_delivery.py`** — `test_ledger_file_and_directory_are_private`
   now skips cleanly on non-POSIX platforms (`os.name != "posix"`). The
   0o700/0o600 permission assertions are unchanged and still enforced on
   Linux/macOS.
2. **`tests/test_codegraph.py`** — `test_scanner_skips_python_symlink_files`
   now probes symlink creation capability first and skips when the OS
   denies it (e.g. WinError 1314 / missing SeCreateSymbolicLinkPrivilege).
   The full assertion path (scanner skips symlinked files, warning emitted)
   still runs on platforms that allow symlinks.

### Why This Exists

Running the suite on Windows produced two hard failures that are not
product bugs — they are tests asserting POSIX-only platform semantics:

- NTFS reports `0o777` for files and directories regardless of ACLs, so
  `stat.S_IMODE(...) == 0o700/0o600` can never pass on Windows even when
  the runtime correctly creates private files.
- `Path.symlink_to(...)` raises `OSError: [WinError 1314]` for non-elevated
  Windows processes because creating symlinks requires a privilege that
  standard developer machines do not grant.

Without the fix, `python -m unittest` on Windows reports 2 failures and the
suite is never green on that platform, which hides genuine regressions and
discourages Windows contributions.

### User / Operator Impact

- Windows and non-elevated developers can run the affected test modules
  and get a clean `OK (skipped=2)` instead of hard failures.
- POSIX CI behavior is identical: the original assertions are untouched and
  both tests still run (not skipped) on Linux/macOS.

### How It Works

- POSIX permission test: a one-line guard before the assertions. On Windows
  the platform cannot express POSIX mode bits, so the test documents why it
  cannot run rather than failing spuriously.
- Symlink test: a small probe creates a temporary target and a symlink; if
  creation raises `OSError`/`NotImplementedError`, the test skips with a
  clear reason. On capable platforms the probe succeeds and the original
  test body executes unchanged.

- `tests/test_awareness_delivery.py` (+3 lines: `import os`, skip guard)
- `tests/test_codegraph.py` (+16 lines: `import os`, `import shutil`,
  `import tempfile`, symlink capability probe)

No product code, no contracts, no generated files, no catalog changes.

## Validation

- [x] `PYTHONPATH=tests python -m unittest tests/test_awareness_delivery.py tests/test_codegraph.py`
      → `Ran 27 tests, OK (skipped=2)`
- [x] `python -m compileall -q tests/test_codegraph.py tests/test_awareness_delivery.py` → clean
- [x] Targeted ad-hoc verification of both skip-guard behaviors (Temp dir, `hermes-verify-` prefix, removed after run) → `OK (skipped=2)`
- [ ] `uv run --group lint ruff check src tests` — not run (uv/ruff not set up in this checkout)
- [ ] `python3 -m unittest discover -s tests` — full suite on Windows has
      pre-existing failures in `test_cli.py` (20 failures/errors, unrelated
      to this change) and long runtimes; out of scope for this PR
- [ ] `python3 -m omh.cli docs workflows --check` — not run (Windows venv toolchain)
- [ ] `python3 -m omh.cli harness validate` — not run
- [ ] `python3 -m omh.cli release checklist --json` — not run
- [ ] Manual Hermes/TUI check: not applicable (test-only change)

## Risk

- Low. Test-only edits; no runtime code paths touched.
- On POSIX, both tests still execute their original assertions (guards are
  `os.name`/capability-conditional, so behavior is unchanged in CI).

## Compatibility / Rollout

- No interface, schema, or CLI changes.
- Safe to merge independently; no migration needed.

## Release / Claims

- [x] Release-channel impact considered: none (test-only, ships with any channel)
- [x] Runtime/native capability claims: none made
- [x] Known manual Hermes checks listed: not applicable

## Follow-Up

- Consider a follow-up pass over `tests/test_cli.py` Windows failures
  (venv-python path resolution, setup path expectations, JSON escape-hatch
  rendering) to get the full suite green on Windows.
- Optionally add a Windows CI job (e.g. `windows-latest`) so these
  compatibility fixes are enforced in CI rather than discovered manually.

### DIFF
--- a/tests/test_awareness_delivery.py
+++ b/tests/test_awareness_delivery.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import json
+import os
 import stat
 import threading
 import unittest
@@ -202,6 +203,8 @@ class AwarenessDeliveryLedgerTests(unittest.TestCase):
             self.assertEqual(legacy.severity, "ok")
 
     def test_ledger_file_and_directory_are_private(self) -> None:
+        if os.name != "posix":
+            self.skipTest("POSIX permission bits are not meaningful on Windows/NTFS")
         with TemporaryDirectory() as tmp:
             record_awareness_delivery(
                 delivered=False,
--- a/tests/test_codegraph.py
+++ b/tests/test_codegraph.py
@@ -1,6 +1,9 @@
 from __future__ import annotations
 
 import json
+import os
+import shutil
+import tempfile
 import unittest
 from pathlib import Path
 from tempfile import TemporaryDirectory
@@ -269,6 +272,19 @@ omh = "src"
         )
 
     def test_scanner_skips_python_symlink_files(self) -> None:
+        # Symlink creation requires SeCreateSymbolicLinkPrivilege on Windows
+        # (absent for non-elevated processes), so skip there.
+        try:
+            tmp_dir = tempfile.mkdtemp()
+            probe = os.path.join(tmp_dir, "probe_target.py")
+            open(probe, "w").close()
+            os.symlink(probe, os.path.join(tmp_dir, "probe_link.py"))
+            os.remove(os.path.join(tmp_dir, "probe_link.py"))
+        except (OSError, NotImplementedError):
+            self.skipTest("symlink creation is not permitted on this platform")
+        finally:
+            shutil.rmtree(tmp_dir, ignore_errors=True)
+
         with TemporaryDirectory() as tmp, TemporaryDirectory() as outside_tmp:
             repo = Path(tmp)
             outside = Path(outside_tmp) / "outside.py"


tests/test_codegraph.py
```diff
@@ -269,6 +272,19 @@ def test_package_dir_name_is_canonical_before_src_alias(self) -> None:
...
+            tmp_dir = tempfile.mkdtemp()
+            probe = os.path.join(tmp_dir, "probe_target.py")
+            open(probe, "w").close()
+            os.symlink(probe, os.path.join(tmp_dir, "probe_link.py"))
+            os.remove(os.path.join(tmp_dir, "probe_link.py"))
+        except (OSError, NotImplementedError):
+            self.skipTest("symlink creation is not permitted on this platform")
+        finally:
+            shutil.rmtree(tmp_dir, ignore_errors=True)
+
```
sionic-khope, line 287: **Blocking.** Two problems here, and the fix for the second removes the first.

**The `finally` can mask the skip.** `tmp_dir` is bound inside the `try`, so if `tempfile.mkdtemp()` raises `OSError` (unwritable `TMPDIR`, disk full, sandbox), `skipTest()` fires and then `finally` hits an unbound name — `UnboundLocalError` replaces `SkipTest` and the test **errors instead of skipping**. Reproduced against this exact block.

**The repo already has a pattern for this**, and it guards the real call instead of a proxy — see `tests/test_web_visual_qa_capture_file_safety.py:32-35` and `tests/test_dynamic_workflow.py:345,365,420`. That matters beyond style: this probe symlinks inside one `mkdtemp` dir, but the real call at line 292 targets a path in a *separate* `TemporaryDirectory`. A passing probe doesn't prove the real call works.

Suggest deleting the probe entirely:

```suggestion
```

...and wrapping the actual call on line 292 (outside this hunk, so I can't attach a suggestion):

```python
            try:
                (repo / "linked.py").symlink_to(outside)
            except (NotImplementedError, OSError) as exc:
                self.skipTest(f"symlink creation unavailable: {exc}")
```

With the probe gone, the new `os`, `shutil`, and `tempfile` imports can all come back out.

tests/test_awareness_delivery.py
```diff
@@ -202,6 +203,8 @@ def test_a_delivery_or_legacy_record_never_false_ages(self) -> None:
             self.assertEqual(legacy.severity, "ok")
 
     def test_ledger_file_and_directory_are_private(self) -> None:
+        if os.name != "posix":
+            self.skipTest("POSIX permission bits are not meaningful on Windows/NTFS")
```
sionic-khope, line 207: **Non-blocking, consistency.** The repo guards POSIX-only semantics with a `unittest.skipUnless` decorator rather than an in-body `if` — see the named guards at `tests/test_runtime_reader_filesystem_faults.py:48-55` (used at `:116`, `:220`, …) and the inline form at `tests/test_runtime_artifacts.py:493`.

The decorator reports the skip without entering the test body, and keeps the precondition visible at the signature:

```suggestion
    @unittest.skipUnless(os.name == "posix", "POSIX permission bits are not meaningful on Windows/NTFS")
    def test_ledger_file_and_directory_are_private(self) -> None:
```

`unittest` is already imported in this module, and your `import os` is still needed.

sionic-khope (commented):
Thanks for this — running the suite on Windows is a real gap, and both failures you found are genuine platform-semantics problems rather than product bugs. The diagnosis is correct on both counts: NTFS reports `0o777` regardless of the underlying ACL, and `SeCreateSymbolicLinkPrivilege` is absent for non-elevated processes. The commit trailers are complete and correctly ordered, which is appreciated.

## Verification I ran

I checked out `a6fe9af` (based on current `main` @ `ce91cfa3`) on macOS and filled in the gates your checklist left unrun:

| Gate | Result |
| --- | --- |
| `unittest tests/test_awareness_delivery.py tests/test_codegraph.py` | `Ran 27 tests — OK`, **0 skipped** |
| `unittest discover -s tests` (full suite) | `Ran 2946 tests — OK (skipped=1)` |
| `uv run --group lint ruff check src tests` | `All checks passed!` |
| `omh.cli docs workflows --check` | `ok: true` (tap_skills 113/113) |
| `omh.cli docs roles --check` | `ok: true` |
| `omh.cli docs capability-families --check` | `ok: true` |
| `compileall -q src tests` | clean |

The single full-suite skip is pre-existing (`test_local_store_locking.test_file_lock_degrades_gracefully_without_fcntl`) and unrelated to this PR.

The important result: **0 skips in the two touched modules on POSIX.** Over-skipping on Linux/macOS was the main risk in a change like this, and it's clean — your guards do stay inert on POSIX exactly as claimed.

## Two things to fix before merge

Both are in the `test_codegraph.py` symlink probe, and fixing the second dissolves the first.

### 1. The probe can turn a skip into an ERROR (reproduced)

`tmp_dir` is assigned *inside* the `try` but cleaned up in `finally`. If `tempfile.mkdtemp()` itself raises `OSError` — unwritable `TMPDIR`, disk full, restrictive sandbox — the `except` fires `skipTest(...)`, and then `finally` dereferences an unbound `tmp_dir`. Reproduced locally against your exact block:

```
unittest.case.SkipTest: symlink creation is not permitted on this platform

During handling of the above exception, another exception occurred:
    shutil.rmtree(tmp_dir, ignore_errors=True)
                  ^^^^^^^
UnboundLocalError: cannot access local variable 'tmp_dir' where it is not associated with a value

FAILED (errors=1)
```

The `UnboundLocalError` replaces the `SkipTest`, so the test **errors instead of skipping** — the exact failure mode this PR exists to remove.

### 2. The repo already has conventions for both guards

This is the larger point, and adopting them also deletes the code path in #1.

**Symlink capability** — the established pattern guards *the actual symlink call*, not a separate probe. See `tests/test_web_visual_qa_capture_file_safety.py:32-35` and three sites in `tests/test_dynamic_workflow.py` (`:345`, `:365`, `:420`):

```python
try:
    (captures_root / "checkout-qa").symlink_to(outside, target_is_directory=True)
except (NotImplementedError, OSError) as exc:
    self.skipTest(f"symlink creation unavailable: {exc}")
```

**POSIX-only semantics** — a `unittest.skipUnless` decorator. `tests/test_runtime_reader_filesystem_faults.py:48-55` defines named module-level guards (`requires_posix_permissions`, `requires_symlinks`) used at `:116`, `:220` and elsewhere; `tests/test_runtime_artifacts.py:493` uses the inline decorator form.

Beyond consistency, guarding the real call is more *correct* here. Your probe symlinks within a single `mkdtemp` directory, but the test's real call is `(repo / "linked.py").symlink_to(outside)` where `outside` lives in a **separate** `TemporaryDirectory`. A passing probe does not prove the real call succeeds, and the two can drift apart. Guarding the real call cannot drift.

Adopting the existing pattern also makes all three new imports (`os`, `shutil`, `tempfile`) unnecessary in `test_codegraph.py` — the file already imports `Path` and `TemporaryDirectory`. Net diff shrinks to roughly +4 lines instead of +16.

## Non-blocking

- **Nothing enforces this.** `.github/workflows/ci.yml:10` is `ubuntu-latest` only, so the Windows compatibility this PR adds can regress silently the next time someone touches these tests. Your Follow-Up section already flags this — worth filing as a tracked issue, since a `windows-latest` job is what makes these guards load-bearing rather than decorative.
- Treating the `test_cli.py` Windows failures as out of scope is the right call for this PR.

Happy to re-run the full gate set once the probe is reworked.

sionic-khope:
Merged — thank you. This is a genuinely useful contribution.

Nobody here had run the suite on Windows, so the two failures you surfaced were invisible to us. Your diagnosis was right on both counts: NTFS reports `0o777` regardless of the underlying ACL, so `stat.S_IMODE(...) == 0o700` can never hold there, and `SeCreateSymbolicLinkPrivilege` is absent for non-elevated processes. Neither is a product bug, and both were quietly making the suite un-greenable on a whole platform. The PR body and the commit trailers were complete and correctly ordered, which made this fast to review.

## Verification before merge

`main` had moved three merges ahead (`ce91cfa3` → `714bb073`) since your CI run, so I re-verified the actual merge result on macOS rather than trusting the older green:

| Gate | Result |
| --- | --- |
| `unittest discover -s tests` (merge result) | `Ran 3189 tests — OK (skipped=1)` |
| `unittest tests/test_awareness_delivery.py tests/test_codegraph.py` | `Ran 27 — OK`, **0 skipped on POSIX** |
| `uv run --group lint ruff check src tests` | `All checks passed!` |
| `docs workflows --check` | `ok: true` (tap_skills 115/115) |
| `docs roles --check` / `docs capability-families --check` | `ok: true` |

The single skip is pre-existing (`test_local_store_locking.test_file_lock_degrades_gracefully_without_fcntl`) and unrelated. The important number is the second row: **0 skips in the two modules you touched.** Over-skipping on Linux/macOS was the main risk in a change like this, and your guards stay inert on POSIX exactly as you claimed.

## About the two review points

I asked for two changes earlier — I've applied them myself rather than making you round-trip, so this could land now. Follow-up PR: #778.

- The symlink probe's `tmp_dir` was assigned inside the `try` and cleaned in `finally`, so a failing `tempfile.mkdtemp()` raised `UnboundLocalError` out of `finally` and replaced `SkipTest` with an ERROR — the exact failure mode the guard exists to prevent.
- Guarding the real `symlink_to` call instead of a probe matches `test_web_visual_qa_capture_file_safety.py` and three sites in `test_dynamic_workflow.py`, and it can't drift: the probe symlinked inside one `mkdtemp` directory while the real call crosses two separate `TemporaryDirectory` roots.
- The POSIX check became a `unittest.skipUnless` decorator, matching `test_runtime_artifacts.py` and the named guards in `test_runtime_reader_filesystem_faults.py`.

Both guards are now mutation-tested — with `Path.symlink_to` raising `OSError(1314)` and with `os.name` patched to `"nt"`, each test skips with 0 errors and 0 failures. That proves they fire, and fire *as skips*; a green POSIX run only proves they stay inert.

These are conventions you had no way to know from the outside. The diagnosis — the part that actually required running an unsupported platform and reading the failures correctly — was yours.

## Follow-up

Your point about CI stands and is still open: `.github/workflows/ci.yml` is `ubuntu-latest` only, so nothing enforces what you just fixed and it can regress the next time someone edits these tests. A `windows-latest` job is what makes these guards load-bearing instead of decorative — that is worth tracking as its own issue.

Treating the `test_cli.py` Windows failures as out of scope was the right call. If you want to take that on next, it's a real gap and the venv-python path resolution is the first thread to pull.

Thanks again.

Issue #781, opened 2026-08-04
Windows test suite: 120 failures + 288 errors (3225 tests) — tracker for Windows-green


mrmixx-max:
## Summary

The full test suite is not Windows-green. On a real Windows host (non-elevated), `python -m unittest discover -s tests` reports:

```
Ran 3225 tests in 517.644s

FAILED (failures=120, errors=288, skipped=22)
```

That is **408 failing/erroring tests** (~12.7% of the suite). The Linux CI matrix (`ubuntu-latest`, Python 3.11/3.12) is fully green, so none of this is visible to CI today. This issue tracks closing the gap so the suite can run on Windows, which would let #780's scoped `test-windows` job expand from the two guard-bearing modules to the full suite.

## Reproduce

On Windows (Python 3.11+, repo at `main`):

```
pip install -e .
PYTHONPATH=tests python -m unittest discover -s tests
```

### 1. Path-separator comparisons (largest visible class)
Code and tests compare expected POSIX paths against `pathlib`-normalized Windows paths.

Examples (both real):

- `tests/test_wrapper_notification_contract.py:112` — expected `['wrapper/contract.py']`, got `['wrapper\\contract.py']`
- `tests/test_config_adapter.py` — `external_dirs(text) == ["/tmp/omh/skills"]`, got `['\\tmp\\omh\\skills']` (absolute-looking `/tmp` paths are parsed as relative paths on Windows and normalized to backslashes)

### 2. POSIX shell / filesystem assumptions
34+ test modules reference `/bin/`, `/tmp/`, `os.symlink`, or `chmod`-style permission logic. Subprocess helpers that shell out to `bash`/`sed`/`awk`/`grep` will fail or misbehave on Windows. (`os.symlink` without the `SeCreateSymbolicLinkPrivilege` is already handled for two modules by #771/#778, but other call sites exist.)

### 3. Permission-bit semantics
POSIX mode-bit assertions (e.g. `stat().st_mode & 0o777`) are not meaningful on NTFS; guarded in `test_awareness_delivery.py` by #771, but related assertions exist elsewhere.

A per-class triage (run the suite, bucket each failure into one of these classes, then fix class-by-class) is the suggested first step. A representative non-failing module: `tests.test_chat_router` (140 tests) passes cleanly on Windows, so the gap is patchwork, not a broken baseline.

## Context

- #771 / #778 added and hardened the first two Windows guards (POSIX-only skips, symlink privilege probe). Both merged.
[...]

PR #851, opened 2026-08-07
feat: native Windows install path, and make it round-trip (closes #848)

### LINKED ISSUE #848: Enable Windows support
`oh-my-hermes` currently assumes a POSIX environment throughout. The installer (`install.sh`) is a Bourne-shell script, and the installation docs reference only `sh` / `curl | sh` flows. On native Windows there is no supported install path.

**What I tried:** On Windows 10, the documented `curl -fsSL ... | sh` install cannot run outside WSL. The substance of the installer is just `pip install` + `omh setup`, which I was able to replicate manually in PowerShell with a venv — so the package itself is close to working.

**Request:**
- A native Windows install path (a PowerShell installer, or documented `py -m pip install` + `omh setup` steps).
- Confirmation of which Hermes config home OMH targets on Windows (`~/.hermes` path handling differs between native Windows and WSL).
- A note in the docs on which skills, if any, are POSIX-only.

The package has zero third-party dependencies, so the main gap appears to be the installer shell script and path handling rather than the core library. Happy to test any Windows branch.

sionic-khope:
## Feature Report

Closes #848.

### What Changed

- **`install.ps1`** — a native Windows installer, next to `install.sh`. Same `OMH_*` environment contract, same package-source resolution, same ordered steps, same argument list handed to `omh setup`.
- **`omh update` / `omh remove` now recognize a Windows install.** The managed-command venv and command-link resolution was duplicated in `src/install/installer.py` and `src/commands/setup.py`, and both copies read `os.environ["HOME"]` and required `path.is_symlink()`. On native Windows `HOME` is unset and the exposed command is an `omh.cmd` shim, so both surfaces failed. They are now one set of helpers in `src/system/paths.py`.
- **The "rerun the installer" hint is platform-aware.** It printed `curl -fsSL … | sh` in all four languages regardless of host.
- **Docs** — a `## Windows` section in `docs/INSTALLATION.md` covering both install paths, the config-home answer, and a capability-boundary table; mirrored into `README.md` + the three localized READMEs, `INSTALL_FOR_AGENTS.md`, and both site pages.
- **CI** — five new `windows-latest` steps that parse `install.ps1` on both PowerShell 5.1 and 7, run it end to end, assert the install → setup → uninstall round trip, and exercise the default `%LOCALAPPDATA%` locations plus the user-PATH edit.

### Why This Exists

OMH has been Windows-green in the *library* since #781/#782 — the **full** suite, not a smoke subset, is an enforcing gate on `windows-latest`. What was missing was any way to install it. `install.sh` is a Bourne script, and the word "Windows" appeared **zero** times across `docs/INSTALLATION.md`, `INSTALL_FOR_AGENTS.md`, and the four READMEs.

The reporter's diagnosis was right — the package was close to working and the installer was the gap. But documenting an install would have been a half-fix: the install did not round-trip. `omh update` reported the command as unmanaged and `omh remove` left the venv and shim on disk, because both located the managed venv through `HOME` and identified the command by `is_symlink()`.

### User / Operator Impact

A Windows user can now run:

```powershell
irm https://raw.githubusercontent.com/rlaope/oh-my-hermes/main/install.ps1 | iex
omh setup
omh doctor
```

…and then `omh update` and `omh remove` behave the way they do on POSIX. The manual `py -m pip install` path is documented as a first-class alternative for anyone who will not pipe a remote script into `iex`.

They also get two answers the issue asked for directly:

| Environment | Hermes home | OMH home |
| --- | --- | --- |
| Native Windows | `C:\Users\<you>\.hermes` | `C:\Users\<you>\.omh` |
| WSL | `/home/<you>/.hermes` | `/home/<you>/.omh` |

Separate stores on separate filesystems; they never merge, and setting `HOME` in a PowerShell profile does not move either, because `ntpath.expanduser` reads `%USERPROFILE%` and ignores `HOME`.

And: **no skill is POSIX-only.** Every skill is guidance plus `omh` commands. What *is* POSIX-only is a set of storage/locking primitives that **fail closed** — the domain intelligence store, domain context attachment, `omh ops prompt-compatibility-audit`, `omh ops plugin-risk-audit`, the benchmark sandbox, and mode-bit enforcement. Generic record locking is deliberately *not* on that list: `local_store` uses `msvcrt.locking` and gets a real OS lock.

### How It Works

`install.ps1` is a port, not a second installer with its own opinions. Where it diverges, the platform forced it:

| Behavior | POSIX | Windows |
| --- | --- | --- |
| Default venv | `~/.local/share/omh/venv` | `%LOCALAPPDATA%\omh\venv` |
| Default command dir | `~/.local/bin` | `%LOCALAPPDATA%\omh\bin` |
| Venv layout | `bin/python` | `Scripts\python.exe` |
| How `omh` is exposed | symlink | `omh.cmd` shim |
| PATH | hint printed | appended to user PATH; `OMH_ADD_TO_PATH=0` opts out |

Two divergences are judgment calls, not mechanics, and both are flagged for review:

1. **`OMH_ADD_TO_PATH` defaults to `1`.** `~/.local/bin` is a convention most POSIX shells already carry, so `install.sh` only prints a hint. Windows has no equivalent, so a hint-only installer would hand every user a command they cannot run. The change is user-scope (not machine), additive, appended only when absent, announced in the installer output, and opt-out.
2. **Installer step labels are English on Windows**, even under `OMH_LANG`. Windows PowerShell 5.1 decodes a BOM-less script as the system ANSI code page, so a localized label would arrive as mojibake on exactly the hosts this file serves. Adding a BOM was rejected because it would ride into the `irm | iex` pipeline that is the documented path, and no local host can verify that is safe. `OMH_LANG` is still validated and still forwarded to `omh setup` as `--language`, so the surface carrying real localized content stays localized.

The shim-ownership test (`command_entry_belongs_to_venv`) parses the program the shim actually runs and asks whether that path resolves inside the venv — the same question the symlink branch already asked. Matching the shim's *text* against the venv path was the first attempt and was wrong in both directions; see Review Findings. The `.cmd` shape is refused entirely off Windows, so POSIX keeps exactly one delete path.

- `install.ps1` (new), `install.sh` (unchanged — it is the reference contract)
- `src/system/paths.py` — `user_home`, `managed_command_venv_dir`, `managed_command_bin_dir`, `managed_command_filenames`, `managed_command_venv_scripts_dir`, `command_entry_belongs_to_venv`
- `src/install/installer.py`, `src/commands/setup.py` — duplicated helpers removed; `POSIX_INSTALLER_COMMAND` / `WINDOWS_INSTALLER_COMMAND` / `installer_command()`
- `src/commands/language.py` — `update_command_next` takes a `{command}` placeholder in all four languages
- `src/commands/main.py`, `src/quality/parity.py` — `install.sh-managed` wording is now executor-neutral `installer-managed`
- `pyproject.toml` — `Operating System :: OS Independent`
- `.github/workflows/ci.yml` — five `test-windows` steps
- `tests/test_windows_install_path.py` (new, 30 tests), `tests/test_cli.py`, `docs/README.md`, `.gitattributes` (new)

## Validation

- [x] `uv run --group lint ruff check src tests` — All checks passed
- [x] `python3 -m unittest discover -s tests` — **Ran 4143 tests, OK** (30 new, 7 of them negative cases for the shim matcher)
- [x] `python3 -m compileall src` — clean
- [x] `python3 -m omh.cli docs workflows --check` — ok, tap-skills 115/115, 0 stale
- [x] `python3 -m omh.cli docs roles --check` / `docs capability-families --check` — ok
- [x] `python3 -m omh.cli harness validate` — ok
- [x] `python3 -m omh.cli release checklist --json` — ok
- [x] `python3 -m omh.cli release hermes-smoke` — ok
- [x] `git diff --check` — clean
- [x] **Round trip against a real venv on macOS**: `install.sh` → `omh setup` → `omh uninstall --dry-run --json` returned `command_package.status == "would_remove"` with both the link and the venv in `command_package_would_remove`. That is the exact assertion the Windows CI step makes, validated on the platform where it can be run locally.
- [x] **The two shim-matcher deletion bugs were reproduced against the old code** and confirmed refused by the new one (see Review Findings below).
- [x] **`install.ps1` — OBSERVED on `windows-latest`.** It parses on Windows PowerShell 5.1 *and* PowerShell 7, installs end to end, writes a working `omh.cmd` shim, and `omh uninstall --dry-run` recognizes that shim as installer-managed. The default `%LOCALAPPDATA%` path a real `irm | iex` user takes was run separately, with the user PATH asserted to still be `REG_EXPAND_SZ` afterwards. Full installer stdout is in a [PR comment](https://github.com/rlaope/oh-my-hermes/pull/851#issuecomment-5214969337).
  - Still unobserved: the `irm | iex` download path (CI runs the script from the checkout), and the OEM-code-page shim under a non-ASCII profile path (`runneradmin` is pure ASCII).
- [ ] Manual Hermes/TUI check — not run; this PR does not change Hermes-visible skill content or routing.

## Review Findings Already Fixed In This PR

The first push was reviewed and run on `windows-latest`; the second commit closes what that found. Recording it here because most of it was invisible from macOS, which is the argument for the CI steps existing at all.

**CI found two:**
- `install.ps1` arrived CRLF on the Windows runner (git autocrlf on checkout), failing the encoding gate → `.gitattributes` pins both installers to LF.
- `test_update_self_update_recognizes_venv_python_symlink_path` patched only `HOME`, so it broke once the resolver started reading `USERPROFILE`. That test had been green by *agreeing with the bug this PR fixes*.

**Review found, in `command_entry_belongs_to_venv` — the function this PR exists to add, whose `True` causes a `force=True` delete:**
- It substring-matched the venv's `Scripts` directory against the shim body. A user's own `omh.cmd` that merely put the venv on `PATH` before running something else matched, and so did anything under a `Scripts_backup` sibling. **Both reproduced.** It now parses the program the shim actually runs and asks whether that resolves inside the venv — the same question the symlink branch already asked, which also fixes a third case: a store behind a junction left the shim stranded because the reader resolves and the installer's spelling does not.

**Review found, in `install.ps1`:**
- `Split-OmhArgString` returned `@()`, which PowerShell unrolls to nothing, so callers got `$null`; `<array> + $null` appends a null element that `[string[]]` renders as `''` and PowerShell 7 passes to pip as a literal empty argument. **The default install would have failed on the host CI runs it on.**
- `Add-OmhBinDirToPath` used `[Environment]::SetEnvironmentVariable`, which reads the user PATH with `%VAR%` already expanded and writes it back as `REG_SZ` — permanently flattening every `%JAVA_HOME%\bin`-style entry on the machine, on every default install. Now goes through the registry with expansion suppressed and writes `ExpandString`.
- `Resolve-OmhPython` stopped at the first too-old candidate rather than trying the rest, which matters because `py` honors `PY_PYTHON` and can land on 3.10 while `python` one slot later is 3.12.
- Failure paths called `exit`, which under `irm | iex` terminates the user's shell along with the diagnostic printed one line earlier. They `throw` now, converted to a process exit only when the file was run as a script.
- Output moved from `[Console]::Out` to `Write-Host`, which the ISE and VS Code integrated console can actually display.
- The shim is written in the OEM code page, because `cmd.exe` decodes batch files with it and a Hangul or Kanji account name would otherwise be mojibake in the interpolated path. CI cannot see this under `RUNNER_TEMP`.

**CI coverage added in response:** the syntax gate now runs on Windows PowerShell 5.1 as well as 7 (the two hosts hide each other's bugs), and a new step runs the installer with no `OMH_VENV_DIR`, `OMH_BIN_DIR`, or `OMH_ADD_TO_PATH` — the code a real `irm | iex` user executes — asserting the user PATH is still `REG_EXPAND_SZ` afterwards.

## Risk

- **`install.ps1` has no local oracle.** It now runs green on `windows-latest`, but it cannot be executed on the authoring host, so every future change to it is unverified until CI says otherwise. That is not hypothetical: two of the defects listed above would have broken the default install and were caught only by review, and three more were caught only by the Windows job across three red runs. PowerShell 5.1 compatibility was handled by review plus a 5.1 parse gate: no PS6+ syntax (ternary, `??`, `$PSStyle`, 3-argument `Join-Path`), no reliance on ANSI VT (5.1 does not enable it on the legacy console, so color is gated to PS7+/Windows Terminal), no embedded double quotes in native command arguments, and the Python probe goes through a temp file rather than `python -c`.
- **`OMH_ADD_TO_PATH=1` mutates the user-scope PATH by default.** Reviewers should confirm they want that default. It is the one behavior in this PR that changes machine state beyond OMH's own directories. The write now preserves `REG_EXPAND_SZ` and CI asserts that, but the decision itself is still yours.
- **The PATH edit is not broadcast.** No `WM_SETTINGCHANGE` is sent, so the installing shell gets the new PATH immediately while other shells need a sign-out. Broadcasting needs runtime P/Invoke that no local host can verify; the installer says this in its output rather than pretending otherwise.
- The `irm | iex` download path is still unobserved — CI runs the script from the checkout, not from `raw.githubusercontent.com`.
- Localized READMEs are now at 253/257/244 lines against the asserted 260 budget. `README.ja.md` has 3 lines of headroom left.

## Compatibility / Rollout

- No behavior change on macOS or Linux. `command_entry_belongs_to_venv` is strictly no more permissive than the symlink check it replaces; the two existing uninstall tests pass unchanged.
- User-visible string change: `install.sh-managed` → `installer-managed` in uninstall/update output and CLI help, in all four languages. No test asserted those strings.
- `INSTALLER_COMMAND` was replaced by `installer_command()`; it had exactly one call site.
- No new dependencies. `install.ps1` needs Windows PowerShell 5.1, which ships with Windows 10 and 11.

## Release / Claims

- [x] Release-channel impact considered — `install.ps1` honors `OMH_CHANNEL` (`preview`/`stable`/`local`) with the same resolution and the same error messages as `install.sh`; a drift gate compares the two.
- [x] Runtime/native capability claims are backed by evidence or marked as not observed — every row of the POSIX-only table was verified against source (`msvcrt` fallback in `local_store`, `bwrap`/`sandbox-exec`/`unsupported` in the adapter sandbox, `ValueError` → `OmhError` in the domain commands, the silent `return None` in `domain_project_context`).
- [x] Known manual gaps listed — `install.ps1` execution and the `irm | iex` path, above.

## Follow-Up

Two real Windows defects were found while scoping this and are deliberately **out of scope** — they are a different user goal (correctness of concurrent writes), not "can I install it":

1. `src/workflows/memory.py`, `src/workflows/observation_journal.py`, and `src/plugin_bundle/omh/awareness_delivery.py` guard on `if fcntl is not None:` with **no `msvcrt` fallback and no degradation signal** — unlike `local_store.file_lock`, which reports `{"locked": False, "enforced": False, "reason": "no_os_file_lock"}`. Journal appends lose mutual exclusion on Windows silently.
2. `src/workflows/domain_project_context.py` swallows the fail-closed `ValueError` to `None`, so domain context attachment is invisibly off on Windows rather than reported as unavailable.

Neither is a regression from this PR. Happy to file both as issues if you want them tracked separately.


sionic-khope:
## `install.ps1` is now observed, not claimed

All checks pass, and the `windows-latest` job ran the installer end to end. Recording the actual output here, because until this run every statement about `install.ps1` in this PR was inference — there is no PowerShell on the authoring host.

**It parses on both hosts** — `Windows installer syntax (PowerShell 7)` and `Windows installer syntax (Windows PowerShell 5.1)` both green. 5.1 is the stated target and the reason for the ASCII-only and no-VT constraints.

**It installs** (`Windows installer end-to-end`):

```
OMH installer
      [note] Channel: local
      [note] Source ref: local
      [note] Mode: venv
[1/3] Create isolated Python environment at D:\a\_temp\omh-install-smoke\venv
      [ok] done
[2/3] Install OMH package
      [ok] done
[3/3] Expose the omh command
      [ok] D:\a\_temp\omh-install-smoke\bin\omh.cmd

oh-my-hermes is installed.
Next: run 'omh setup' to connect OMH to Hermes, then 'omh doctor' to verify.
```

**The round trip closes** (`Windows managed-command round trip`): `omh setup` ran from the installed `omh.cmd`, and `omh uninstall --dry-run --json` came back `command_package.status == "would_remove"` — `installer-managed command recognized on Windows`. That is the defect this PR exists to fix, proven on the platform it affects.

**The default path a real `irm | iex` user takes works** (`Windows installer default locations and PATH`), with no `OMH_VENV_DIR`, `OMH_BIN_DIR`, or `OMH_ADD_TO_PATH` set:

```
[1/3] Create isolated Python environment at C:\Users\runneradmin\AppData\Local\omh\venv
      [ok] done
[2/3] Install OMH package
      [ok] done
[3/3] Expose the omh command
      [ok] C:\Users\runneradmin\AppData\Local\omh\bin\omh.cmd
      [note] Added 'C:\Users\runneradmin\AppData\Local\omh\bin' to your user PATH.
      [note] This shell has it now; other shells pick it up after you sign out and back in.
      [note] Set OMH_ADD_TO_PATH=0 to skip this next time.
```

That step also asserts `HKCU\Environment\Path` is still `REG_EXPAND_SZ` afterwards, which pins the fix for the PATH-flattening bug review caught.

### What the three CI rounds cost, and why the gates are shaped this way

Worth stating plainly for the reviewer: the first two `windows-latest` runs **failed**, both in the unit-test step ahead of the installer gates, so the installer had never actually executed until this run.

1. `install.ps1` arrived CRLF (git autocrlf on Windows checkout) → `.gitattributes`.
2. A fixture patched `HOME` but not `USERPROFILE`.
3. The same fixture then cleared `XDG_DATA_HOME` but not `LOCALAPPDATA`.

(2) and (3) are the same class of bug: `managed_command_venv_dir` walks five inputs in a fixed precedence and had **no test of its own** — all coverage was incidental, through fixtures that happened to set the right subset. `ManagedCommandLocationTests` now pins that order on both platform flavours, including both cases that broke.

### Still not observed

- The `irm | iex` download path. CI runs the script from the checkout, not from `raw.githubusercontent.com`.
- The OEM-code-page shim under a non-ASCII profile path. `RUNNER_TEMP` and `runneradmin` are both pure ASCII, so a Hangul or Kanji account name is reasoned about, not tested.
- No `WM_SETTINGCHANGE` broadcast after the PATH edit, so other shells need a sign-out. The installer says so rather than pretending otherwise.

Issue #802, opened 2026-08-05
Honor organization safety rules in every Hermes workflow


sionic-khope:
> *This was generated by AI during triage.*

## Inspiration

The sources below are design inspiration only. The resulting capability must be native OMH behavior and must not require the source plugin, service, registry, or host feature at runtime.

- [OpenClaw local policy](https://docs.openclaw.ai/tools/skills-config)
- [OPA concepts](https://www.openpolicyagent.org/docs)

## User problem

Teams need local safety and governance rules applied consistently without teaching users a separate control plane or editing every workflow.

## Native Hermes experience

Users state the desired outcome normally. Hermes checks prepared work against organization rules and either continues or explains the exact conflict and safe next action.

## Current OMH base

OMH already has project-governance discovery, deterministic validation, fail-closed states, and chat-first status.

## Native capability to add

Add an opt-in organization rule check over bounded OMH metadata that stays invisible on success and explainable on failure.

## How it blends into OMH

Setup records the local rule source; governance contributes repository requirements; applicable preparation invokes the check; Hermes renders safe reason codes.

## Boundary-level design

- Send bounded metadata only, never code bodies, credentials, raw prompts, or hidden reasoning.
- Bound time and size and fail closed on timeout or malformed results.
- Record rule revision and deterministic reason code.

## Triage

- **Priority:** P0
- **Impact:** high
- **Area:** `area/quality`
- **Category:** enhancement
- **State:** needs-triage

## Acceptance criteria

- [ ] The same bounded input and revision produce the same decision or explicit failure.
- [ ] Every result identifies the active revision and safe reason.
- [ ] Unavailable or malformed checks cannot silently allow work.

## Out of scope

- Evaluating third-party package code.
- Calling a hidden remote policy service.

## User value

Organizations gain consistent safeguards while users continue working through ordinary Hermes conversation.

sionic-khope:
Backlog consolidation note (triage sweep of #789-#838): #802 and #804 are naturally ONE evaluator with two rule sources (built-in native profile + opt-in org rule source). Scope note: implement the preflight evaluator once (#804), with #802's org-rule source as a pluggable input honoring the same bounded-metadata, fail-closed, revision-recording contract.

PR #883, opened 2026-08-09
feat: recheck prepared work when safety rules change

### LINKED ISSUE #805: Recheck prepared work when safety rules change
> *This was generated by AI during triage.*

## Inspiration

The sources below are design inspiration only. The resulting capability must be native OMH behavior and must not require the source plugin, service, registry, or host feature at runtime.

- [OPA signed bundles](https://www.openpolicyagent.org/docs/management-bundles)
- [OpenClaw source pinning](https://docs.openclaw.ai/plugins)

## User problem

Work prepared under an older safety revision can become invalid before it starts, while users may assume earlier approval still applies.

## Native Hermes experience

Hermes detects stale preparation, explains which safety revision changed, and offers to recheck it before continuing.

## Current OMH base

OMH already has schema-versioned contracts, hashed governance sources, atomic storage, and prepared-not-observed enforcement.

## Native capability to add

Bind every safety-evaluated artifact to its rule revision and optionally verify a trusted local signature.

## How it blends into OMH

Preparation records the rule hash; dispatch compares it with current rules; status marks stale work; Hermes regenerates or re-evaluates without losing chat continuity.

## Boundary-level design

- Bind each artifact to the exact rule content hash.
- Keep local signature validity separate from the safety decision.
- Retain the last valid rule set when a new signed revision fails.

## Triage

- **Priority:** P2
- **Impact:** medium
- **Area:** `area/quality`
- **Category:** enhancement
- **State:** needs-triage

## Acceptance criteria

- [ ] Every evaluated artifact reports its rule revision.
- [ ] Changed rules block continuation until re-evaluated.
- [ ] Signature failure cannot activate the new revision or reuse its decisions.

## Out of scope

- Downloading remote policy bundles.
- Building a signing-key management service.

## User value

Users can trust Hermes to apply today’s safety rules rather than stale preparation.

sionic-khope:
## Feature Report

Closes #805.

Two of the three acceptance criteria on #805 were already satisfied before this
branch existed. They are now pinned by tests rather than reimplemented, and the
third — the one that was genuinely missing — is built on top of them.

### What Changed

- **AC1, already met, now pinned.** `quality/safety_preflight._verdict` has
  always stamped `safety_profile_revision` onto every verdict it emits, allow
  and deny alike, and `workflows/approval_receipts.py` and
  `workflows/blocked_work_records.py` already carry it onward.
  `AlreadyMetAcceptanceCriterionPins.test_ac1_every_evaluated_artifact_reports_its_rule_revision`
  locks that across four outcomes (allow, builtin denial, non-object request,
  org denial). No production code changed for AC1.
- **AC2, already met, now pinned.** `recheck_safety_preflight_revision` already
  reports `current` / `drifted` / `missing` and refuses to call a drifted
  revision dispatchable, and `coding/action_gate._revision_drift_denial` already
  turns that into a denial *before* arbitration (asserted in
  `tests/test_risky_action_confirmation.py`).
  `test_ac2_changed_rules_block_continuation_until_re_evaluated` locks the
  preflight half. No production code changed for AC2.
- **AC3, new.** `src/coding/safety_rule_attestation.py` adds an opt-in local
  attestation over the org safety rule source, plus the retention behaviour the
  issue's boundary-level design asks for: a failed verification cannot activate
  the new revision, and the last revision that *did* verify stays in force and
  is reported as retained with the failing revision named.

### Why This Exists

The org safety rule source (#802) is read from a local path and applied. Nothing
checked that the bytes at that path were the bytes the operator meant to
install, and nothing survived a bad revision: a source that changed simply took
effect, and a source that broke fell back to a denial rather than to the rules
that were last known good. #805 AC3 states the requirement directly — *signature
failure cannot activate the new revision or reuse its decisions* — with the
design constraint that local signature validity stays separate from the safety
decision.

### User / Operator Impact

An operator who sets `attestation_key_path` in the `org_rule_source_policy`
block of `setup-profile.json` gets three things they did not have:

- A changed rule source that has not been re-attested no longer takes effect.
- The last attested rule set keeps applying instead of the lane collapsing to a
  denial, and the report names which revision was rejected.
- Three distinct refusals with distinct corrections — no tag was written, the
  bytes no longer match the tag, or the key itself could not be read — rather
  than one opaque failure.

An operator who sets nothing sees no change at all. This is asserted, not
assumed: `UnconfiguredInstallsAreUntouched` checks that the payload handed to
the evaluator is `assertEqual` to what the pre-#805 reader returns for the same
file, that the resulting verdict is identical, that no trust file is created,
and that a stray tag file lying beside the source is ignored.

### How It Works

**The tag.** HMAC-SHA256 over the raw source bytes, keyed by a secret the
operator holds on this machine, computed with stdlib `hmac` and `hashlib`. The
tag lives in a sidecar beside the source it covers (`<source>.hmac-sha256`) —
derived rather than configured, because a tag path that can point elsewhere can
point at a stale file. Key and tag reads are bounded (4 KiB / 256 B), refuse
symlinks, and strip surrounding whitespace so an editor's trailing newline or a
Windows CRLF is the same secret on every host.

**Two axes, kept apart.** `signature_state` is one of `not_required`, `valid`,
`missing`, `invalid`, `unverifiable`. The org source `status` is untouched.
`TheTwoAxesStaySeparate.test_signature_state_and_safety_status_are_independent_fields`
drives five combinations and asserts the exact set: a valid tag over a malformed
document is still `unavailable` for `org_source_malformed`; an invalid tag over
a well-formed document is `unavailable` for `org_source_attestation_invalid`;
and an invalid tag with a retained set on record is `available` — the same
signature state against both safety answers, and the same safety answer against
both signature states.

**Activation and retention.** `activate_org_safety_rules()` returns
`activated`, `retained`, or `unavailable`, with `effective_revision` and
`rejected_revision`. `not_required` and `valid` activate; the other three never
do. Only a `valid` verification writes the trust state, which is what makes "the
last valid set" mean that rather than "the previous one" — asserted by
`test_the_retained_set_is_the_last_valid_one_not_merely_the_previous_one`, which
rejects two revisions in a row and checks that both fall back to the first. A
retained set is also checked to belong to the same source path, so repointing
the configured path cannot inherit trust another path earned.

**No partial application.** `test_the_retained_rule_set_carries_nothing_from_the_rejected_revision`
asserts the served rule set is exactly the retained one and that the rejected
revision's label and its denied-kind value appear nowhere in the payload.
`test_a_failed_verification_leaves_the_trust_file_byte_identical` compares the
trust file bytes before and after three consecutive failed activations.

**Trust state.** A flat, metadata-only record — two digests, a revision label,
the two bounded rule values, and a caller-supplied timestamp — at
`runtime/safety_rule_trust.json` inside the OMH home, written through
`atomic_write_json(private=True)`. Read defensively in the idiom
`capabilities/toggles.py` uses: an unexpected key set, a wrong schema version,
or any field failing its own shape check reads as *no* retained set, never as a
partially trusted one (ten tamper cases in `TrustStateIsDefensive`).

**Determinism.** No wall clock anywhere in a compared payload; `attested_at` is
a parameter. `test_the_same_inputs_give_the_same_answer_every_time` re-runs the
activation five times and compares serialized output.

| File | Change |
| --- | --- |
| `src/coding/safety_rule_attestation.py` | New. Tag verification, trust state, activation/retention. 438 lines including the docstring that states the limitation. |
| `src/coding/project_governance.py` | Three refusal codes join `ORG_SAFETY_RULE_SOURCE_REASON_CODES`. `_org_source_unavailable` became public `org_safety_rule_source_unavailable` (now validating its reason code), and `org_safety_rule_source_available` was extracted so the retained set is built from the same literal the reader uses. `read_org_safety_rule_source_with_bytes` exposes the bytes of the same read the rules came from. |
| `src/quality/safety_preflight.py` | Three `_CORRECTIONS` entries, one per new refusal. Nothing else; the existing org level already denies on any non-available reason code. |
| `src/capabilities/toggles.py` | `attestation_key_path` joins the `org_rule_source_policy` payload (blank = not required) plus its accessor. |
| `src/commands/capability_policy.py` | `omh capability-policy status` reports the attestation key path in JSON and prints a "Local rule attestation" line when the org source is on. |
| `src/system/paths.py` | `OmhPaths.safety_rule_trust_state_path`. |
| `docs/ARCHITECTURE.md` | New "Local Rule Attestation And Retention" subsection under the existing org-source section, including the honest statement of what HMAC does not prove. |
| `tests/test_safety_rule_attestation.py` | New. 43 tests. |
| `tests/test_safety_preflight.py` | Two AC pins; one moved assertion (below). |

**Assertions that moved, and why.** Exactly one:
`test_a_hand_edited_value_cannot_smuggle_extra_fields` asserted
`set(policy) == {"schema_version", "enabled", "source_path", "claim_boundary"}`.
`attestation_key_path` joins that closed key set, so the expected set grew by
one member. The bound itself was not weakened — a hand-edited `extra` key is
still dropped, and the test now additionally asserts that an absent attestation
key reads as the blank default. No other existing assertion changed.

**Contract note.** The three new reason codes enter `_RULES` through the
existing splat of `ORG_SAFETY_RULE_SOURCE_REASON_CODES` and enter `corrections`
through `_CORRECTIONS`, so `safety_profile_revision()` — the profile digest —
changes with this PR. That is correct behaviour for a rule change: prepared work
pinned under the old digest now reads as `drifted` and must be re-evaluated,
which is AC2 doing its job. No test pins a literal digest.

## Validation

All run on the rebased branch (`origin/main` at `f9251979`), full output
observed:

- [x] `uv run --group lint ruff check src tests` — `All checks passed!` (one
      pre-existing warning about an invalid `# noqa` directive at
      `src/commands/main.py:1`, untouched by this PR)
- [x] `PYTHONPATH=tests uv run python -m unittest discover -s tests` —
      `Ran 4965 tests in 185.336s / OK`
- [x] `uv run python -m compileall -q src tests` — exit 0
- [x] `uv run python -m omh.cli docs workflows --check` — `"ok":true`,
      `tap_skills` 115/115, no missing/stale/extra
- [x] `uv run python -m omh.cli docs roles --check` — `"ok":true`
- [x] `uv run python -m omh.cli docs capability-families --check` — `"ok":true`
- [x] `uv run python -m omh.cli harness validate` — `"ok":true`,
      `{"harnesses":72,"skills":104}`, no errors, no warnings
- [x] `uv run python -m omh.cli release checklist --json` — `"ok":true`,
      `required_item_count: 35`, mode `plan`
- [x] `uv run python -m omh.cli release hermes-smoke` — plan rendered, exit 0
- [x] `uv run omh --help` — exit 0 (console script inside the project venv; not
      a globally installed `omh` on PATH)
- [x] `git diff --check` — exit 0
      — not run. This PR touches no install, packaging, or skill-content
      surface; nothing it changes is reachable from the installed-command path.
- [ ] Workflow-learning review queue smoke — not applicable; no learning
      contract changed.
- [ ] Manual Hermes/TUI check — **not run.** Nothing in this PR is reachable
      from a Hermes chat turn. The org rule level is still not wired into the
      coding delegation lane (a boundary
      `tests/test_task_authority_envelope.py::test_the_org_level_is_not_wired_into_this_lane`
      asserts on purpose), so the attestation lane that extends it has no chat
      surface to check.

Targeted counts: `tests/test_safety_rule_attestation.py` — 43 tests, all
passing. `tests/test_safety_preflight.py` — 52 tests (was 50), all passing.

## Risk

**Read this section before approving. It is the reason this PR should be held
for an explicit decision.**

**HMAC is symmetric.** The key that verifies a tag is the same key that produces
one. This means the check proves the file was not altered by someone *without*
the key. It does not prove **who** authored the rules, and **anyone who can read
the key file can forge a tag that verifies**. That is not a bug in this
implementation; it is what a keyed digest is. It is asserted rather than
footnoted: `test_a_key_holder_can_mint_a_tag_that_verifies` writes an
attacker-authored rule source, tags it with the operator key, and asserts it
activates.

**A real provenance scheme needs a dependency.** Proving *who* signed a rule set
requires public-key signing — Ed25519 (`cryptography` or `pynacl`), minisign, or
GPG (a subprocess to an external binary). Each is a runtime dependency, and
`AGENTS.md` requires explicit owner approval for one. The trade in front of you:

| | Local HMAC (this PR) | Ed25519 / minisign / GPG |
| --- | --- | --- |
| Detects tampering by a non-key-holder | Yes | Yes |
| Proves who authored the rules | **No** | Yes |
| Survives a leaked/readable key | **No** — forgeable | Verification key is public; forging needs the private key |
| New dependency | None (stdlib `hmac`) | Yes — package or external binary |
| Packaging story | Unchanged | New wheel dependency or a binary probe with a host-capability fallback |

Because of this, no wording anywhere in the change calls the tag a "signature"
in the provenance sense. The module docstring, the claim boundary constant, the
reason codes, the CLI line, and the architecture doc all say **locally
attested**, and two tests (`test_the_claim_boundary_names_the_limitation`,
`test_the_module_says_plainly_what_the_tag_does_not_prove`) fail if that wording
is softened. If you would rather have real provenance, the right move is to
approve a dependency and replace `verify_safety_rule_attestation` — the two-axis
split, the refusal vocabulary, and the retention path are all independent of
which primitive verifies the bytes and would survive that swap unchanged.

**Second risk: retention is a deliberate softening of fail-closed.** Everywhere
else in `#802`'s reader, an unusable source means denial. Here, a source that
fails attestation falls back to the last set that passed, which may be *less*
strict than the revision that was just rejected. That is exactly what the issue's
boundary-level design asks for ("retain the last valid rule set when a new
signed revision fails"), and the fallback set is one an operator did attest —
but it is a real behavioural difference from the surrounding module, and it is
worth an explicit yes. The narrower alternative (refuse outright, never retain)
is a one-line change if you prefer it.

**Third risk: the key path is recorded but not yet consumed by any lane.** Same
status as `source_path` from #802: nothing in `src/` reads the policy into the
evaluator. The plumbing that would make this active in a chat turn is a separate
piece of work.

**Blast radius if wrong.** Nothing runs unless an operator sets a key path.
Anything that could regress for an existing user is confined to the changed
`org_rule_source_policy` key set and the profile-digest change described above.

## Compatibility / Rollout

- **Default off.** No key path means `not_required`, no file opened, no trust
  file created, and the org source result equal to the previous behaviour.
- **Setup profile.** `read_org_rule_source_policy` rebuilds the payload on
  read, so a profile written before this PR gains
  `attestation_key_path: ""` with no migration and no rewrite.
- **Profile digest changes.** Prepared work carrying the pre-#805
  `safety_profile_revision` now rechecks as `drifted` and must be re-evaluated
  before dispatch. This is the designed behaviour of AC2, not a regression, but
  it will be visible to anyone holding a prepared artifact across the upgrade.
- **Nothing was removed.** `read_org_safety_rule_source` keeps its exact
  signature and result shape for the unattested path.

## Release / Claims

- [x] Release-channel impact considered — `local`; no packaging, install, or
      skill-content surface touched.
      observed — every claim here maps to a named test; the manual Hermes check
      is marked not run with the reason above.
- [x] Known manual Hermes checks are listed, including the
      `omh release hermes-smoke --live` gap — not run, and not applicable:
      nothing in this PR is reachable from a Hermes chat turn.

## Follow-Up

- **Owner decision on the primitive.** If provenance rather than integrity is
  what #805 wanted, that needs a dependency approval; the swap is scoped to
  `verify_safety_rule_attestation`.
- Wiring the org level (and therefore this lane) into a real evaluation path.
  It is still reachable only from a direct evaluator call, exactly as #802 left
  it and as `docs/ARCHITECTURE.md` states.
- A CLI writer for `org_rule_source_policy`. Neither `source_path` nor
  `attestation_key_path` has one today; both are set by editing the setup
  profile.


sionic-khope:
**Holding this one for a maintainer decision before merge.** Everything below is green — full suite, all four byte gates, ruff, Windows CI — so this is not a quality hold. It is a design question that belongs to the repo owner.

**The question:** AC3 asks that "signature failure cannot activate the new revision". `AGENTS.md` forbids new dependencies without explicit approval, and the Python standard library has no public-key signature primitive. So this PR implements attestation as **HMAC-SHA256 over the raw source bytes with a key the operator holds locally**.

That is a real integrity control and it does satisfy the acceptance criterion as written, but it is symmetric. It proves the bytes were not altered by someone *without* the key. It does not prove who authored them, and anyone who can read the key can mint a tag that verifies. The PR does not paper over this — there is a test, `test_a_key_holder_can_mint_a_tag_that_verifies`, that writes an attacker-authored source, tags it with the operator key, and asserts it activates. Nothing in the code, reason codes, CLI output, or docs uses the word "signed"; two tests fail if that wording is softened to imply provenance.

**The choice:**

| Option | Cost | What you get |
| --- | --- | --- |
| Merge as-is | none | Local tamper-evidence against a non-key-holder. Honest about its limits. |
| Swap for Ed25519 / minisign / GPG | one new runtime dependency plus a packaging story, which `AGENTS.md` says needs your explicit approval | Real third-party provenance. |
| Drop attestation, keep the rest | lose AC3 | AC1 and AC2 were already met before this PR; the two-axis split and retention path are the genuinely new parts and stand on their own. |

The two-axis split (signature validity is a separate field from the safety decision) and the retention path (a failed attestation keeps the last *valid* rule set rather than the last one) are independent of the primitive. Either survives a later swap without redesign.

One more thing worth your eye, separate from the primitive: **retention deliberately softens fail-closed.** Everywhere else in this reader an unusable source denies. Here a failed attestation falls back to the previously attested rule set, which may be *less* strict than the revision that was just rejected. The issue's boundary design asks for exactly that, but it differs from the surrounding module, so it is a yes/no you may want to make explicitly rather than inherit.

PR #902, opened 2026-08-10
fix: stop one ambiguous word and an unlisted noun deciding the workflow

sionic-khope:
### What Changed

Two independent routing defects, both reproducible on clean `main`, both sending ordinary messages to the wrong workflow at **high** confidence.

**1. A trigger matched in both directions.** `_phrase_match` was `query in value or value in query`. The first arm means a *fragment* of a trigger counts as having said it. Triggers now fire one-directionally via a new `_trigger_phrase_match`.

**2. A closed noun list decided whether a build request routed at all.** Greenfield creation had no guard; the only guard shaped like one, `_direct_coding_task_guard_applies`, enumerates surfaces you *edit*. A new `GREENFIELD_BUILD_GUARD` matches the creation shape and routes to `deep-interview`.

### Why This Exists

Measured on `main`, read-only:

| message | routed to | why |
| --- | --- | --- |
| `test` | command-operator, score **73**, high | substring of `npm test`, `cargo test`, `pytest`, `python -m unittest` — +6 each |
| `design` | design-quality-gate, **56**, high | same shape |
| `review` | security-safety-review, **58**, high | same shape |
| `audit` | workspace-audit, **56**, high | same shape |
| `omh` | workflow-learning, **28**, high | substring of `did not use OMH`, `OMH was not used` |

That last row is the clearest statement of the bug: typing the product's own name opened *"record a missed-route improvement candidate"* — a workflow whose matching triggers all say **you did not use OMH**.

The second defect, with the verb and sentence shape held constant and only the product noun varying:

| message | routed to |
| --- | --- |
| `build a login component` | ultraprocess (44) |
| `build a settings page` | ultraprocess (44) |
| `build a navbar` | ultraprocess (44) |
| `build a todo list` | **oh-my-hermes (4)** |
| `build a react app` | **oh-my-hermes (4)** |
| `build a dashboard` | **oh-my-hermes (4)** |
| `let's build a react todo list` | **oh-my-hermes (4)** |

`navbar`, `toggle`, and `readme` are in the literal set; `app`, `todo`, and `dashboard` are not. The arbitrariness is the defect.

### User / Operator Impact

- One ambiguous word no longer dispatches to a specialist that merely *contains* it. `omh` reaches the OMH picker; `audit` asks instead of guessing.
- A greenfield build request reaches the clarification lane whatever the product noun is. `deep-interview` — one question at a time, clarified brief with non-goals before planning — is what OMH already had for "materially ambiguous product request"; nothing routed to it.
- Requests that were already routed correctly are untouched: `npm test` still reaches `command-operator`, `make me a landing page` still reaches `frontend`.

### How It Works

`_trigger_phrase_match` is one-directional. The bidirectional `_phrase_match` is **left alone** and still backs description/use_when scoring, where a short query legitimately appears inside a long prose field.

`GREENFIELD_BUILD_GUARD` is a **floor, not an override**, in two ways that both had to be found by measurement:

- **Registered last.** The guarded fast path takes the *first* rule in the list, not the highest boost. Registered mid-list it selected `deep-interview` for "…create a research brief" over `research-brief`, whose own guard sat two places behind it with a boost of 38.
- **Boost 8, not a round number.** The candidate is injected at score 0, so the boost *is* the score of an unclaimed greenfield request. It has to sit above the nothing-matched band (3–7) and below a genuine multi-signal match (10). At 30 it overrode `frontend` (15) and `websearch-setup` (18); at 10 it tied `agent-board` (10, off four separate signals) and turned a correct dispatch into a clarify.

| File | Change |
| --- | --- |
| `src/routing/recommend.py` | `_trigger_phrase_match`; trigger loop switched to it; guard id added to `_GUARDRAIL_CANDIDATE_INJECTION_IDS` |
| `src/routing/policy.py` | `GREENFIELD_BUILD_GUARD` + `_greenfield_build_guard_applies` + three phrase/token sets |
| `src/quality/routing_precision.py` | 6 new intervention cases (2 positive, 4 overroute guards) |
| `tests/test_routing_scoring.py` | **new** — 7 scorer-level contracts |
| 5 test files | exact-count fixtures 151→157, 208→214; two frozen assertions corrected |

### Review

**Two assertions were updated, with the reason rather than the value.** Both froze the pre-fix routing of bare `omh` to `workflow-learning`. That routing existed only because the bare word was a substring of that skill's negative-usage triggers. `oh-my-hermes` carries a genuine bare `omh` trigger. `test_the_display_prefix_alone_never_invents_a_route` is *better* served by the new behaviour — the picker is precisely "not inventing a route"; a missed-route learning card is one.

**Two proposals were measured and rejected**, both from the analysis that produced this change:

- *Derive trigger tokens only from single-token triggers.* Measured: **42 workflow changes**, including `아침마다 시장 리서치 요약해줘` research-department(43) → nothing(0). Reverted.
- *Make `_phrase_match` one-directional everywhere.* Deletes description/use_when scoring for every skill.

The two bare-word cases (`test`, `audit`) could not become intervention cases: that corpus also asserts the raw message never appears in the machine payload, and a single common word always does. They live in `tests/test_routing_scoring.py` instead.

## Validation

- [x] `PYTHONPATH=tests uv run python -m unittest discover -s tests` — **5717 tests OK**
- [x] Main-vs-branch routing differential, 239 messages (208 fixture + 31 probes) run in **separate processes** on each checkout — `_recommend_skills_cached` (maxsize=2048) makes in-process monkeypatching silently report zero change
- [x] **0 of 208 fixture cases regressed.** One changed and improved: `ok` visual-qa(24, high) → context-budget-review(6, medium)
- [x] `docs workflows --check`, `docs roles --check`, `docs capability-families --check` — exit 0
- [x] `ruff check src tests`, `compileall -q src tests`, `git diff --check` — exit 0
- [x] `harness validate`, `release checklist --json` — exit 0
- [x] Routing precision gate: 57/57 negative controls, 157/157 interventions, overroutes 0, missed interventions 0
- [ ] Live Hermes chat rendering — **not observed**

## Risk

- **Broad blast radius by nature.** A scoring change reaches every routable skill. The differential is the evidence, and it covers the fixture corpus plus 31 hand-written probes — not real user traffic.
- **The greenfield guard's boost is a threshold between two measured bands**, so a future skill scoring 8–9 on a creation-shaped message could tie it. The `agent-board` case is in the suite as the canary.
- **The creation-phrase list is shape-based, not exhaustive.** `write a …` is deliberately absent (documents, not software). Korean `만들어줘` is broad; it is held in check by the floor boost rather than by vocabulary, which is weaker than an explicit rule.
- Bare `test` still reaches `build-failure-triage` at 10 — much lower, but that skill declares a bare `test` trigger, which is a catalog-data question this PR does not open.

## Compatibility / Rollout

- No schema, no persisted state, no generated artifact. Reverting the commit restores the previous scoring exactly.
- Release channel: `local`. No version file or changelog touched.

## Release / Claims

- [x] Release-channel impact considered — none
- [x] Runtime capability claims backed by evidence or marked not observed — no new capability claimed
- [x] Known manual Hermes checks listed — live chat rendering not observed

## Follow-Up

- `build-failure-triage` declaring the bare token `test`, and `code-review` declaring bare `review`/`audit`, are the remaining single-token-trigger ambiguities. CLAUDE.md's own rule is "token triggers only when a single token is unambiguous"; these are not.


sionic-khope:
Closing without merging, on the repo owner's call.

The two defects here are real and reproduce on clean main — bare `test` reaching
`command-operator` at 73, and `build a todo list` collapsing to the picker at 4
while `build a navbar` reached 44. CI was green on all four checks including
Windows.

But the fix is four more keyword lists (`_GREENFIELD_CREATION_PHRASES`,
`_GREENFIELD_QUESTION_OPENERS`, `_GREENFIELD_EXISTING_REFERENT_PHRASES`,
`_GREENFIELD_OTHER_LANE_TOKENS` — ~80 literals) plus a 53rd guard and a
score_boost tuned against measured bands. The symptoms improve; the structure
that produces them gets worse.

That structure is the actual issue. Routing is ~16,900 lines across `policy.py`
(7,804, 52 guards), `chat.py` (6,722), and `recommend.py` (2,373) for 104
skills, and it carries the fossils of every previous round of this same fight:
`_SIBLING_POINTER_METADATA_TOKENS`, `_GENERIC_TRIGGER_TOKENS`, per-skill token
subtractions inside `_score_definition`, and an `ULTRAPROCESS` score_boost whose
own comment records that **renaming a skill** forced a retune.

`src/routing/input_language.py:112` already states the principle: "a
deterministic trigger score is not evidence of intent and the decision belongs
to model selection over supplied candidates." It is applied only to
non-Latin/non-Hangul scripts. Generalising that is the work, and it should not
land under a bugfix that pushes the other way.

Superseded by routing-architecture work. The measurements, the 239-message
main-vs-branch differential harness, and the overroute guard cases are worth
keeping and will be reused there.

sionic-khope:
Reopening. I closed this on an argument that did not survive measurement.

The replacement I proposed — stop letting vocabulary overlap read as certainty, and let the existing `clarify` path ask instead — was built and measured on a branch. It fails this repo's own routing accuracy ratchet:

| | baseline | with that change |
| --- | --- | --- |
| ko resolved | 6/6 | **5** |
| overall resolved | 13/36 | **12** |

`tests/test_routing_accuracy.py` says floors are "never to be lowered to make a change pass", and it already scopes the model-selection idea deliberately: the floor is 0 for languages with no trigger table because "the fix for these is model selection, not more tokens". I pushed that boundary into English and Korean, where the tables do carry real entries, and resolution dropped.

This PR, by contrast, passed the full suite including that ratchet — 5717 tests, all four CI checks green including Windows. The bidirectional trigger match is a plain bug: `test` is not evidence the user said `npm test`.

The maintenance concern that motivated closing it is still real — 18 per-skill exceptions inside `_score_definition`, 52 guards, and an `ULTRAPROCESS` boost whose comment records that renaming a skill forced a retune. But that is a separate problem from this fix, and the answer to it cannot be paid for in resolution rate.

PR #923, opened 2026-08-11
fix(plugin): add kind: standalone so omh tools/hooks register in live Hermes

### LINKED ISSUE #922: [Bug]: omh plugin declares 10 tools and 4 hooks but Hermes never registers them (auto-coerced to kind=exclusive)
## Summary

The plugin bundle ships a `plugin.yaml` with no `kind` field. Since April 2026 Hermes auto-coerces any user plugin whose `__init__.py` mentions `register_memory_provider` to `kind: exclusive` (hermes-agent #14005). The general PluginManager skips exclusive plugins entirely, so omh's `register()` only ever runs through the memory provider loader, which passes a collector whose only real method is `register_memory_provider`. The function registers the provider and returns before reaching the tool and hook wiring.

Net result: the 10 `omh_*` tools and 4 hooks declared in `plugin.yaml` never register in a live Hermes session. The memory provider itself works. `omh doctor` stays green because its import and register smoke tests call the module directly with a fake context, and its runtime observation is satisfied by the memory provider's own `session_start_recovery` record. None of those checks prove Hermes loaded the tool and hook surface.

Verified on Hermes 0.20.0 by running the real PluginManager against the installed bundle:

```
kind: exclusive | enabled: False
tools declared: 10, registered: 0
hooks declared: 4, registered: 0
```

Adding `kind: standalone` to `plugin.yaml` fixes it. All 10 tools and 4 hooks register, and the memory provider still activates through the separate `plugins/memory` text scan.

## Reproduction

1. Install omh 1.0.5 with the plugin bundle (`omh setup --with-plugin`).
2. Set `memory.provider: omh` and `plugins.enabled: [omh]` in Hermes config.
3. Start a Hermes session.
4. Inspect the tool catalog. No `omh_capabilities`, `omh_hud`, `omh_status`, or any other omh tool is present.
5. Check `~/.omh/runtime/awareness_delivery.json`. The file does not exist. `pre_llm_call` writes it on every turn, so its absence means that hook has never run.

## Expected behavior

The plugin's declared tools and hooks register in a live Hermes session, or the bundle stops declaring tools and hooks it cannot register.

## Environment

- OS: macOS
- Hermes: 0.20.0 (2026.8.3), git install
- Python: 3.11
- omh: 1.0.5

## Logs or output

Session log excerpt:

```
22:46:18 agent.memory_manager: Memory provider 'omh' registered (0 tools)
22:46:18 run_agent: Memory provider 'omh' activated
```

Loader probe against the installed bundle:

```
kind: exclusive
enabled: False
registered omh tools: []
```

Loader probe with `kind: standalone` added to a copy of the bundle:

```
kind: standalone
enabled: True
registered omh tools: ['omh_capabilities', 'omh_context', 'omh_gather_evidence', 'omh_hud', 'omh_interact', 'omh_memory', 'omh_probe', 'omh_recommend', 'omh_role', 'omh_status']
registered hooks: ['on_session_end', 'pre_llm_call', 'pre_tool_call', 'pre_verify']
```

rgxbrel:
## Summary

The omh plugin bundle ships a `plugin.yaml` with no `kind` field. Since April 2026 Hermes auto-coerces any user plugin whose `__init__.py` mentions `register_memory_provider` to **kind: exclusive** (hermes-agent #14005). The general PluginManager skips exclusive plugins entirely, so omh's `register()` only ever runs through the memory provider loader, which passes a collector whose only real method is `register_memory_provider` — the provider registers and returns before reaching the tool/hook wiring.

**Net result:** the 10 `omh_*` tools and 4 hooks declared in `plugin.yaml` never register in a live Hermes session. The memory provider works; `omh doctor` stays green because its smoke tests call the module directly.

## Fix

Adding `kind: standalone` to `plugin.yaml` resolves it — all 10 tools and 4 hooks register in a live session, and the memory provider still slots through the separate `plugins/memory` text scan.

```yaml
name: omh
kind: standalone
version: "1.0.5"
```

## Verification

- Reproduced on Hermes 0.20.0 with the real PluginManager against the installed bundle: `kind: exclusive | enabled: False`, tools registered: 0, hooks registered: 0.
- After the change: `kind: standalone | enabled: True`, all 10 `omh_*` tools and 4 hooks registered.
- `tests/test_plugin_distribution.py`: **29 passed** (12 subtests).
- Distribution tests confirm the bundle still packages the yaml and advertises the full tool/hook set.

## Scope / Risk

One-line YAML manifest fix, no runtime code path touched. The memory provider's activation is unchanged (still via `plugins/memory` scan).

Fixes #922

### DIFF
--- a/src/plugin_bundle/omh/plugin.yaml
+++ b/src/plugin_bundle/omh/plugin.yaml
@@ -1,4 +1,5 @@
 name: omh
+kind: standalone
 version: "1.0.5"
 description: "Oh My Hermes native bridge for metadata-only runtime status, HUD, role context, bounded evidence probes, evidence-boundary context, and a deterministic file-backed memory provider."
 author: oh-my-hermes maintainers


sionic-khope:
Thanks for this — the evidence is the good kind. The `awareness_delivery.json` observation in #922 is what makes it convincing rather than plausible: `pre_llm_call` writes that file every turn, so its absence is direct evidence the hook never ran, independent of any loader probe.

Worth stating plainly from our side: **we cannot verify this locally.** OMH's own checks are exactly the ones you identified as blind — `omh doctor`'s register smoke test calls the module directly with a fake context, and `tests/test_plugin_distribution.py` asserts the bundle *packages* the yaml and *advertises* the tool set, never that anything registered. So our green was a derived signal, not an observed one. That gap is being tracked separately.

One blocker before this can land: **the DCO check is failing.** Every commit needs a `Signed-off-by:` trailer matching the commit author. From your branch:

```sh
git commit --amend -s --no-edit
git push --force-with-lease
```

If you'd rather not, say so and we'll land the same one-line change from a branch in this repo with your finding credited here — but the signoff is the shorter path and keeps the commit yours.

PR #935, opened 2026-08-12
fix(plugin): verify real Hermes registration

### LINKED ISSUE #919: [Docs]: dreaming's automatic triggers and selection criteria exist only in code comments
## Summary

Dreaming's automatic trigger points and its selection criteria live only in code comments. The one place that describes dreaming to a reader — the `memory-sync` skill — says:

> Dreaming remains reminder-only: it prepares reminders for `stale_review_required` and `expired_volatile_records`, never consolidation, retirement, restore, or prune.

That sentence is true about what OMH *writes* and misleading about when it *runs*. Read on its own it says dreaming is a passive notice a human goes looking for. It is not: the scheduler fires by itself at five distinct points, and a reader who wants to know which ones has to open `memory_provider.py`.

## What is actually missing from every reader-facing surface

**The five automatic trigger points** (`src/plugin_bundle/omh/memory_provider.py`):

| hook | trigger | why it exists |
| --- | --- | --- |
| `on_turn_start` | `turn` | the interval came due (default 5, Letta's sleeptime cadence) |
| `on_pre_compress` | `compaction` | compression is about to discard messages |
| `on_session_end` | `session_end` | the session produced something worth keeping |
| `shutdown` | `shutdown` | last chance before the process goes |
| `initialize` | `session_start_recovery` | the previous session died without consolidating |

**The selection criteria**, which is what anyone evaluating the design actually asks about:

- the brief nominates duplicate clusters, records at or near their deadline, and headroom below the floor
- `unexplained_entries_are_not_candidates` — an entry OMH cannot account for is never proposed for eviction, and the executor instruction says so directly: *"Leave anything you cannot source; an unsourced entry is not evidence it is wrong."*
- suppression: a standing condition is not restated until its value changes (the comment records 19 consecutive identical briefs as the reason)

**The ranking privileges and their limits**, all documented in comments and nowhere a reader would look:

- pins are guaranteed-inclusion anchors but explicitly not eligibility overrides — a pinned record still fails closed on expiry, scope, perspective, and review
- attention tiers say how much working context a record may occupy, never whether it is true
- `approved_manual` carries 100% veracity weight vs 90% for `approved_auto_safe`, and an unknown mode fails closed to the lower — "a trust signal must never default to maximum trust"
- age tiers degrade old records only within an equal relevance rank
- usage ranks on saturating buckets so delivery counts "cannot compound into a permanent head start"

## Why this is worth fixing rather than leaving in comments

These are good answers to the question people actually ask about agent memory — *what stops a bad source from persisting, and what decides what survives?* Right now answering it requires reading four modules, and the one summary a reader is offered points the wrong way on the first question.

Concretely observed: working from the `memory-sync` wording plus a skim of `memory_dreaming.py`, the automatic triggers were missed entirely and dreaming was described as reminder-driven. The docstrings are excellent; they are just not where anyone reads.

## Expected

- `skills/omh-memory-sync` (via `src/skills/catalog.py`, since the SKILL.md is generated) states that the scheduler fires automatically, names the five triggers, and keeps the existing and correct boundary that OMH never performs the consolidation.
- A reference page — `docs/MEMORY_CONTEXT.md` or a dreaming section beside it — carries the selection criteria and the ranking privileges with their stated limits.
- No behaviour change; this is a docs and generated-guidance issue.

## Environment

macOS 25.5.0, main at `9c134895`.

### LINKED ISSUE #924: [Bug]: nothing verifies the plugin bundle's tools and hooks actually register, so zero registered reads the same as all ten
## Summary

Nothing in this repo checks that the plugin bundle's tools and hooks actually register in a live Hermes session. Every green signal we have is derived from something upstream of registration, so a bundle that registers **zero** tools reports exactly the same as one that registers all ten.

#922 is the concrete instance: `plugin.yaml` carries no `kind`, Hermes coerces the bundle to `kind: exclusive`, the general PluginManager skips exclusive plugins, and the 10 `omh_*` tools and 4 hooks never register. That went unnoticed through a full release. This issue is about the blind spot, not the one-line manifest fix (#923).

## What each green signal actually proves

| signal | what it asserts | what it does not |
| --- | --- | --- |
| `omh doctor` → `plugin_enabled` | the plugin is listed in Hermes config | that Hermes loaded it |
| `omh doctor` register smoke test | `register()` runs against a **fake context** we construct | that the real PluginManager ever calls it |
| `omh doctor` runtime observation | a runtime record exists | that record is written by the **memory provider's** own `session_start_recovery`, which loads by a different path and would exist even with every tool dead |
| `tests/test_plugin_distribution.py` | the bundle packages `plugin.yaml` and advertises 10 tools | that any of them register |
| `omh harness validate` | the harness adapter contract | nothing about the Hermes plugin surface |

Read together they say "the bundle is well-formed and well-advertised." They were read as "the bundle works."

This is the same failure shape as [pipe-masked exit codes]: a derived success signal standing in for a terminal state nobody checked.

## Why `omh doctor` in particular is worth fixing

`doctor` is the command a user runs to answer "is my install healthy?" It answered yes for a plugin whose entire tool surface was inert. A check whose fake context cannot reproduce the real loader's behaviour should either reproduce it or stop claiming it.

Note that the memory provider genuinely does work through the `plugins/memory` text scan, so the failure is partial — which is why it looked green rather than obviously broken.

## What would have caught it

Ideas in rough order of cost, not a prescription:

1. **A manifest conformance check.** `plugin.yaml` must declare every field Hermes needs to load a standalone plugin, `kind` included. Cheap, offline, no Hermes required, and it catches this exact class before release. Needs the field list to come from somewhere authoritative rather than being hand-copied.
2. **A registration assertion against the real loader.** Import Hermes' `PluginManager` when it is importable, run it against the built bundle, and assert the declared tool and hook names come back registered. Skipped with a named reason when Hermes is absent, so CI does not silently pass on the skip.
3. **A doctor check that reports the truth it has.** Today the register smoke test says "registered" when it means "callable with our own fake context". At minimum it should say which of the two it observed, so a reader is not told more than was checked.

(3) is worth doing regardless — it costs almost nothing and removes a claim we cannot support.

## Boundary

Everything above stays inside OMH's rules: no network, no LLM call, no patching Hermes. (2) reads a locally installed Hermes if one is present and reports `not_observed` otherwise; it never installs one.

## Related

- #922 — the bug this blind spot hid
- #923 — the one-line manifest fix

sionic-khope:
### What Changed

- Validate the installed OMH `plugin.yaml` against the standalone Hermes loader contract and the exact 11-tool/4-hook metadata surface.
- Run the installed plugin through a real Hermes `PluginManager` subprocess inside an isolated temporary `HERMES_HOME`.
- Distinguish fake-context smoke, real-loader observation, and host runtime activity in doctor/probe output across English, Korean, Japanese, and Chinese.
- Document dreaming automatic triggers, selection/suppression rules, and ranking limits in the generated memory-sync skill and memory reference.

### Why This Exists

- A bundle that registered zero tools previously looked identical to a healthy bundle because every check stopped before the real Hermes loader.
- Dreaming automatic scheduling and retention-ranking boundaries existed only in implementation comments, while the reader-facing skill sounded passive.

Closes #924
Closes #919

### User / Operator Impact

- `omh doctor` now blocks when an observed real loader registers the wrong tool/hook surface and names `not_observed` when Hermes is unavailable.
- Missing `kind: standalone` is caught offline and by the real loader instead of surviving a release.
- Memory reviewers can see when dreaming runs and what its ranking signals can and cannot do.

### How It Works

- Manifest conformance compares `kind`, `provides_tools`, and `provides_hooks` with OMH runtime metadata without adding a YAML dependency.
- The loader probe copies only the managed plugin bundle into a temporary Hermes home, enables only `omh`, launches the local Hermes Python, and reads registered tools/hooks from the actual `PluginManager` result.
- Probe failure never reads as healthy: unavailable Hermes is a named advisory; observed registration mismatch is blocking.

- `src/install/plugin_loader_observation.py`, `src/install/plugin_pack.py`
- Doctor/probe output and localized observation boundaries
- Generated `omh-memory-sync` guidance and `docs/MEMORY.md`
- Plugin distribution, CLI localization, and process-spawn governance tests

### Affected Surfaces

- [x] Hermes chat or wrapper
- [ ] Codex
- [ ] Claude Code
- [x] Generic executor
- [x] Not executor-specific

## Validation

- [x] `git diff --check`
- [x] `omh --help`
- [x] Relevant dry-run or smoke command: isolated real Hermes loader, missing-kind reproduction, Hermes-absent observation, doctor/probe happy and bad-input paths
- [x] Manual Hermes/TUI check, or explicit reason it was not run: no TUI behavior changed; doctor/probe CLI and real loader are the matching surfaces

### Observed Evidence

- Full suite: 6,019 tests OK on the final commit tree.
- Focused plugin/memory-sync tests: 49 OK.
- Real local Hermes loader registered 11 tools and 4 hooks from the current bundle.
- Removing `kind:` reproduced #922: Hermes classified the bundle as exclusive, registered zero tools/hooks, and doctor exited 1 with two blocking checks.
- `HERMES_PYTHON` pointing to a missing interpreter produced a named `hermes_not_installed` warning and no false registration claim.
- Ruff, compileall, three generated-doc gates, wheel-content inspection, diff check, and independent review/QA passed.

- Hermes versions whose private PluginManager result fields differ were not available. Such drift returns named `loader_probe_failed`/not-observed evidence rather than green.
- Network, LLM, vendor messenger, and native memory mutation paths are not used by these changes.

## Risk

- Moderate: doctor now launches one bounded local subprocess when the plugin is installed. The subprocess runs only from the explicit operator doctor command, uses a temporary isolated Hermes home, and is governed by the process-spawn allowlist.

## Compatibility / Rollout

- No runtime dependency added and no persisted data migration.
- Hermes absent or incompatible remains advisory/not-observed; an actual observed registration mismatch is blocking.
- Existing fake-context smoke remains available but is labeled as such.

## Release / Claims

## Follow-Up

- None for #924 or #919.


sionic-khope:
Superseded by #936. The replacement commit is DCO-compliant and has the exact same verified tree hash.

PR #970, opened 2026-08-15
feat(routing): add shared route decision contract

ewardGPT:
## What changed

- Add compact privacy-safe `route_decision/v1` propagation across chat, route hints, runtime records, wrappers, and coding handoffs.
- Add deterministic routing-quality golden coverage for known lanes, ambiguity, and fallback.
- Fix project-root temp anchoring, nested glob cache invalidation, symlink-loop degradation classification, and pytest collection configuration.

## Why

Consumers now share one bounded routing decision contract while legacy payloads remain compatible. The quality gate makes catalog/scoring regressions visible before rollout.

## Validation

- Full suite: 3,231 passed, 9 skipped, 5,414 subtests
- Targeted route/E2E checks: passed
- Ruff, compileall, and `git diff --check`: passed

Closes ubuntu-a768


src/routing/chat.py
```diff
@@ -1391,6 +1392,7 @@ def _enriched_route(
     candidate_handoff = build_candidate_handoff(route, message)
     if candidate_handoff:
         route["candidate_handoff"] = candidate_handoff
+    route["route_decision"] = build_route_decision_contract(route)
```
sionic-khope, line 1395: **Important — build the shared contract after governance has finalized the route.** This line snapshots the pre-policy route and `_apply_skill_governance()` then mutates the outer action/skill/harness. Reproduction with the existing project policy `{project: {skills: ["code-review"]}}` and message `plan a product change` returns outer `dispatch / code-review / critic` but embedded `route_decision` `dispatch / plan / planning`; fail-closed policy can similarly embed `dispatch` inside an outer `clarify`. Every new wrapper/runtime consumer therefore receives a stale contract. Apply governance first, then build `route_decision` from the final route, and add both override and fail-closed regression cases.

src/runtime/records.py
```diff
@@ -839,6 +855,7 @@ def build_routing_record(routing: dict[str, Any]) -> dict[str, Any]:
     record = {
         "schema_version": SCHEMA_VERSION,
         "updated_at": utc_now(),
+        "route_decision": routing.get("route_decision", {}),
```
sionic-khope, line 858: **Important — do not copy an untrusted nested contract verbatim into the runtime record.** A route decision containing `action=123`, `selected_skill={...}`, `margin=true`, and `candidates=[{"prompt": "SECRET" + 500000 chars}]` is accepted by `validate_routing_record()` with zero errors and can be persisted, because the new validator checks only the outer key set plus two field types. That breaks the advertised compact/privacy-safe `route_decision/v1` boundary. Rebuild through the central contract builder or add one strict validator for required fields, enums, scalar types, candidate keys/skill strings, and size bounds, then use it at every persistence boundary.

sionic-khope (commented):
6 important, 1 suggestion. Round 1 at `2d6c132428c96b89b14a58fbf22b84e9811e6ab5`; the whole PR was reviewed.

This PR adds a compact `route_decision/v1` projection across chat, route hints, runtime records, wrappers, and workflow-learning metrics, plus several filesystem/test compatibility fixes. The inline comments identify five reproduced contract/privacy defects.

One additional Important finding is in commit metadata: both `b32610aaad0fffda4382cb71e317886c9112a54e` and `2d6c132428c96b89b14a58fbf22b84e9811e6ab5` have valid DCO signoffs but omit every required Lore trailer from `AGENTS.md` (`Constraint`, `Rejected`, `Confidence`, `Scope-risk`, `Directive`, `Tested`, and `Not-tested`). Please rewrite the commits with the full trailer block and keep `Signed-off-by` last.

Suggestion: replace the current short PR body with the repository PR template. The change affects public routing, wrapper, runtime, and learning contracts, but the body currently omits user/operator impact, boundary-level mechanics, affected surfaces, compatibility/rollout, risk, claim-boundary confirmations, and not-tested gaps.

Observed gates:
- affected/claimed-fix suite: `Ran 355 tests in 3.817s` / `OK`
- Ruff: passed (with the pre-existing invalid-`noqa` warning)
- compileall: passed
- workflows/roles/capability-families byte gates: passed
- `git diff --check origin/main...HEAD`: passed
- full suite: `Ran 6550 tests in 266.260s`, with 16 setup/plugin/doctor failures matching the known clean-main baseline observed in this session; they are not attributed to this PR
- GitHub currently reports DCO only; no CI workflow checks are attached to this head

src/workflows/workflow_learning.py
```diff
@@ -441,6 +448,80 @@ def write_learning_trace(paths: OmhPaths, trace: dict[str, Any]) -> dict[str, An
     return trace
 
 
+def build_routing_quality_metrics(paths: OmhPaths, *, limit: int | None = None) -> dict[str, Any]:
+    """Aggregate bounded routing outcomes from local metadata-only traces."""
+    traces = _read_valid_learning_records(paths.learning_traces_dir, validate_workflow_learning_trace)
+    if limit is not None:
+        traces = traces[-max(limit, 0) :]
```
sionic-khope, line 455: **Important — `--limit` does not select the newest N traces.** `_read_valid_learning_records()` returns filename order, then this slice treats the last filename as newest. With `a-new.json` dated December (`dispatch`) and `z-old.json` dated January (`fallback`), `limit=1` reports the old fallback. Also, `limit=0` and negative limits return *all* traces because `[-0:] == [0:]`. Sort by `(created_at, trace_id)` before slicing and make non-positive limits empty or reject them consistently; add newest, zero, and negative controls.

src/workflows/workflow_learning.py
```diff
@@ -441,6 +448,80 @@ def write_learning_trace(paths: OmhPaths, trace: dict[str, Any]) -> dict[str, An
...
+        "schema_version": ROUTING_QUALITY_METRICS_SCHEMA_VERSION,
+        "generated_at": utc_now(),
+        "privacy": {
+            "mode": PRIVACY_MODE,
+            "raw_prompt_stored": False,
+            "raw_platform_event_stored": False,
+            "dimensions": ["action", "confidence", "route margin", "operator correction metadata"],
+        },
+        "scope": {
+            "learning_dir": str(paths.learning_dir),
```
sionic-khope, line 492: **Important — this contradicts the payload's own `do not identify users` claim.** `str(paths.learning_dir)` is an absolute local path and commonly contains the account name, organization, or project path. A temp path containing `alice-private-home-*` is emitted verbatim immediately above the claim at lines 518-520. Omit the path, use a stable non-identifying scope label/fingerprint, or narrow the claim and privacy contract with explicit review.

src/workflows/workflow_learning.py
```diff
@@ -947,6 +1028,14 @@ def record_missed_route(
         outcome="failed",
         feedback_summary="Operator reported that Hermes did not use the intended OMH workflow.",
     )
+    trace["feedback"] = {
+        "kind": "missed_route",
+        "expected_workflow": str(expected_workflow or ""),
+        "expected_harness": str(expected_harness or ""),
+        "expected_next_action": str(expected_next_action or ""),
```
sionic-khope, line 1035: **Important — arbitrary operator text is now persisted in a trace that declares `metadata_only` and `raw_prompt_stored: false`.** The CLI accepts these fields as unconstrained strings. Passing `--expected-next-action RAW_PRIVATE_PROMPT_DO_NOT_STORE` writes that sentinel verbatim to the learning-trace JSON while the trace privacy block lists only hashes/lengths/route summaries as stored fields. Parse workflow/harness/next-action as bounded identifiers from closed vocabularies (or store only hash/length), and add a hostile raw-text regression proving it never reaches disk.

PR #981, opened 2026-08-16
fix(update): stream pip output in human mode and bound the self-update

sionic-khope:
### What Changed

`omh update`'s command-package self-update no longer runs pip silently or unboundedly. Human mode inherits stdio so pip's own download and build lines stream to the terminal; JSON mode keeps capturing (with `-q`) so machine output stays parseable. Both self-update paths — the pip archive install and the package-manager variant — are bounded by a 900-second timeout that raises an actionable `OmhError` naming the package URL/command and the retry, instead of waiting forever.

### Why This Exists

Observed live: a user ran `omh update`, saw `[1/2] Updating omh command package...` with the archive URL, and nothing else — for minutes. The step downloads a repository archive and builds a large sdist through `pip -q` with stdout/stderr captured, so a working multi-minute install is indistinguishable from a hang, and a genuinely stalled network hangs forever. Users kill it and report `omh update` as frozen; this is the delivery pipeline every other OMH improvement depends on.

### User / Operator Impact

- `omh update` now shows pip's `Collecting… / Downloading… / Building wheel…` progress while the archive installs; a slow link looks slow instead of dead.
- A stalled download fails after 15 minutes with: `command package update timed out after 900s downloading/building <url>; check the network and rerun \`omh update\``.
- `omh update --json` output is byte-for-byte unaffected on success.

### How It Works

The pip argv is built once; `-q` is inserted and stdio captured only when `--json` is set, otherwise stdio is inherited (`stdout=None`). `subprocess.run(..., timeout=SELF_UPDATE_TIMEOUT_SECONDS)` wraps both self-update call sites; `TimeoutExpired` is converted to `OmhError`. The reentry exec that refreshes workflows stays unbounded on purpose — it is the real update work, not a network wait. The load-bearing `--no-cache-dir --force-reinstall` cache-busting flags are unchanged.

- `src/commands/setup.py` — `_run_command_package_self_update`, `_run_package_manager_self_update`, new `SELF_UPDATE_TIMEOUT_SECONDS`
- `tests/test_cli.py` — two regression tests (mode-dependent stdio/`-q`/timeout wiring; timeout→actionable-error conversion)

### Affected Surfaces

- [ ] Hermes chat or wrapper
- [ ] Codex
- [ ] Claude Code
- [ ] Generic executor
- [x] Not executor-specific

### Verification (observed)

- `PYTHONPATH=tests uv run python -m unittest test_cli.CliTests -k self_update` — 7/7 (5 existing + 2 new).
- `ruff` clean, `compileall` clean, `release drift` 16 checks, `git diff --check` clean.

### Risks

- Installs that legitimately need >15 minutes (extremely slow links) now fail with the retry message instead of eventually succeeding; the message names the exact rerun.
- Human-mode pip output formatting is pip's own and may vary across pip versions; nothing parses it.

### DIFF
--- a/src/commands/setup.py
+++ b/src/commands/setup.py
@@ -470,6 +470,12 @@ def _command_package_self_update_plan(args: argparse.Namespace) -> dict[str, obj
     }
 
 
+# Bounds the package download + sdist build. A repository archive of this repo
+# builds in well under two minutes on ordinary hardware; fifteen gives slow
+# links headroom while still turning a stalled network into an error.
+SELF_UPDATE_TIMEOUT_SECONDS = 900
+
+
 def _run_command_package_self_update(args: argparse.Namespace, plan: dict[str, object]) -> int:
     if plan.get("method") == "package_manager":
         return _run_package_manager_self_update(args, plan)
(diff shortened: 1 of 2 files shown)


sionic-khope:
Superseded by #987. The timeout half was rejected by the owner — a bound turns a slow but working download into a partial failure without making it faster. The streaming half was the genuinely useful part and ships in #987, without the timeout, alongside the revert of the PyPI publish step now that PyPI is not being set up.

PR #985, opened 2026-08-16
fix(update): slim the install/update artifact and publish to PyPI

sionic-khope:
## Capability

`omh update` and fresh installs stop pulling the whole repository. The stable
channel names the published ~2.7 MB release wheel, fresh installs default to
the newest release and resolve it in the installer shell, and the release
workflow now publishes that same wheel to PyPI so a future version-less update
can let pip resolve "latest" without omh making a network call.

## Motivation

Measured before this change:

| artifact | size | wall clock |
| --- | --- | --- |
| `archive/refs/heads/main.zip` (preview default) | 44 MB, 1485 files | **5m 05s** |
| `archive/refs/tags/v1.0.6.zip` | 45,921,555 B | — |
| `releases/download/v1.0.6/oh_my_hermes-1.0.6-py3-none-any.whl` | 2,714,885 B | **0.57s** |
| `releases/latest` redirect (tag resolution) | header only | **0.26s** |

Two independent causes, not one:

1. **Size.** `assets/` (31M), `tests/` (20M) and `site/` (8.3M) ride along in the
   archive and none of it is needed to run `omh`.
2. **Serving path.** GitHub *generates* `archive/refs/heads/<branch>.zip` per
   request; it is not a cached static object. Release assets are static CDN
   objects. That is why 44 MB took five minutes instead of the ~10s a 44 MB CDN
   download costs — and it means shrinking the archive would not have been
   enough. The fix has to leave the on-demand endpoint.

## Implementation, at the boundary level

- **Stable channel → release wheel.** `package_url_for` resolves the published
  wheel asset for any `vX.Y.Z` tag, and falls back to the tag archive for any
  version outside the shape the release workflow accepts, rather than pointing
  pip at a guessed URL. A missing asset names the archive that does exist.
- **Installers default to the newest release.** `install.sh` / `install.ps1`
  default `OMH_CHANNEL` to `stable` and resolve the newest tag from the
  `releases/latest` 302, in `curl`/`wget`/`Invoke-WebRequest`. This lives in the
  installer scripts on purpose: they are shell, not core `omh`, so the
  no-network boundary is untouched. An unresolvable redirect stops with a
  message naming both escapes (`OMH_VERSION`, `OMH_CHANNEL=preview`).
- **Release workflow publishes to PyPI.** The same immutable wheel that becomes
  the GitHub asset is published to PyPI, publish-or-verify like the existing npm
  step, through the `id-token: write` permission the job already holds.
- **Preview is unchanged and still honest.** It tracks `main`, where no per-tag
  asset can exist, and it says what the download costs.

## Why PyPI rather than resolving in omh

An earlier revision of this branch resolved `releases/latest` with `urllib`
inside `omh` itself. It worked and measured 0.3s, but `AGENTS.md` forbids
network calls in core features absent an owner-approved scoped integration, the
no-network claim is stated as fact in `CLAUDE.md`, `CONTEXT.md` and
`docs/DIRECTION.md`, and `plugin_risk_audit.py` flags exactly that pattern.
Shipping it would have made four documents false. The owner chose PyPI instead;
that revision is fully reverted and `src/` contains no `urlopen`.

Letting pip resolve against an index was verified as the only workable
alternative: `pip install --no-index --find-links
https://github.com/rlaope/oh-my-hermes/releases oh-my-hermes` fails with
"from versions: none" because the page's asset hrefs are lazy-loaded.

## Observed verification

- Full suite: **6710 tests, 16 failures**. Baseline on `main` in the same
  environment: **6690 tests, 17 failures**. Zero failures added. The failures
  are pre-existing and environment-dependent (`omh doctor` exits 1 on this
  machine); CI is green.
- `uv run python -m compileall -q src tests` — clean.
- `docs workflows|roles|capability-families|ulw-inventory|ulw-site --check` — all pass.
- `uv run --group lint ruff check src tests` — All checks passed.
- `git diff --check` — clean.
- Wheel and redirect timings above measured with `curl -w '%{size_download} %{time_total}'`.

## Risks

- **The PyPI step has never run.** It requires Trusted Publishing configured for
  this repo, `release.yml`, and the `npm` environment — owner-only setup. Until
  then the step fails and the release stops there. No observed publish evidence
  exists and none is claimed.
- **Installers now default to `stable`.** A fresh install without `OMH_VERSION`
  depends on the `releases/latest` redirect and on the tag carrying a wheel
  asset. A release that publishes no wheel breaks the default install path for
  everyone, not only for people who pinned that version.
- **`omh update`'s own default is still `preview`** and still pays the 44 MB
  on-demand archive. Flipping it to a name-based `pip install --upgrade
  oh-my-hermes` is deliberately left for a follow-up, because it must not land
  before a release has actually reached PyPI.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -200,6 +200,33 @@ jobs:
             npm publish "$OMH_TARBALL" --access public --provenance
           fi
 
+      - name: Publish or verify PyPI package
+        shell: bash
+        run: |
+          set -euo pipefail
+          # The same immutable wheel that became the GitHub release asset also
+          # goes to PyPI, so `pip install --upgrade oh-my-hermes` resolves the
+          # newest version itself. That is what lets `omh update` leave the
+          # ~44 MB on-demand branch archive behind WITHOUT omh's own process
+          # ever opening a socket to ask which version is newest -- the
+          # no-network boundary in AGENTS.md stays intact because the index
+          # does the resolving.
+          # Idempotent across re-runs, and it re-verifies rather than trusting
+          # a previous run: a version already on PyPI must carry this exact
+          # wheel, byte for byte.
+          published_sha256=""
+          if response="$(curl -fsS "https://pypi.org/pypi/oh-my-hermes/$OMH_VERSION/json" 2>/dev/null)"; then
+            published_sha256="$(
+              printf '%s' "$response" \
+                | python3 -c 'import json,sys; print(next((u["digests"]["sha256"] for u in json.load(sys.stdin)["urls"] if u["packagetype"] == "bdist_wheel"), ""))'
+            )"
+          fi
+          if [ -n "$published_sha256" ]; then
+            test "$published_sha256" = "$OMH_WHEEL_SHA256"
+          else
+            uv publish --trusted-publishing always "$OMH_WHEEL"
+          fi
+
       - name: Render verified Homebrew formula
         shell: bash
         run: |
--- a/docs/INSTALLATION.md
+++ b/docs/INSTALLATION.md
@@ -351,9 +351,12 @@ equivalent — this is what the installer automates:
 ```powershell
 py -m venv $env:LOCALAPPDATA\omh\venv
 & $env:LOCALAPPDATA\omh\venv\Scripts\python.exe -m pip install --upgrade `
-    https://github.com/rlaope/oh-my-hermes/archive/refs/heads/main.zip
+    https://github.com/rlaope/oh-my-hermes/releases/download/v<version>/oh_my_hermes-<version>-py3-none-any.whl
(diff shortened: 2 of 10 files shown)


sionic-khope:
CI observed green on `c2f76bcf`, all six checks:

| Check | Result |
| --- | --- |
| `test (3.11)` | pass, 9m10s |
| `test (3.12)` | pass, 9m52s |
| `test-windows` | pass, 12m39s |
| `distribution (ubuntu-latest)` | pass, 26s |
| `distribution (macos-latest)` | pass, 1m24s |
| DCO | pass |

This settles the one open question in the validation section. The 16 local suite
failures reported above are host-environment only — both full-suite jobs pass on
CI, which has neither a real Hermes Agent install nor the missing `bun` toolchain
that caused them.

`test-windows` passing is the closest available evidence for the `install.ps1`
stable-channel change, which cannot be executed off Windows.

PR #1000, opened 2026-08-18
feat(hud): show Hermes-native delegate_task subagents live in the OMH HUD

sionic-khope:
## Capability

The `[OMH]` dock now shows Hermes-native `delegate_task` subagents in real time — one line per child under the composer: id, goal, `model:effort`, mixture category, tokens, tok/s, cache%, cost, elapsed, state. Finished children linger 15 minutes as `✓ … done` rows and the header reads "N done" instead of "0 agents".

## Motivation

A ralplan/ulw-work wave spawns Hermes-native children (`sa-*`/`deleg_*`), but the HUD only read OMH's own runtime store, so the dock said `0 agents • $0.000 • ctx --` while three subagents were working, and the assistant narrated "추론 강도: unknown" even though Hermes records the child's reasoning effort. The owner reported both live from a running session.

- New `plugin_bundle/omh/hermes_delegation.py`: read-only join of the surfaces Hermes already writes — `state.db` child sessions (`model`, `reasoning_config.effort`, `_delegate_from`), `session_model_usage` (tokens/cache/cost/first-last seen), `async_delegations` lifecycle, and `cache/delegation/live/*/manifest.json` + task-log mtimes (goal text, liveness). Every read degrades to an idle segment; SQLite is opened `mode=ro`.
- Category label is an explicit projection from the shipped mixture chains (dict-parity test against `SHIPPED_MODEL_RECOMMENDATIONS`); a child on the parent session's own model is labeled `inherit` so an unrouted wave is visible as unrouted.
- `read_omh_hud` merges the native rows into the existing `subagents` projection; the widget reuses the existing `ActivityRow` renderer, plus a `✓` marker for done rows and an honest "N done" header label.

OMH still never patches Hermes, makes no network calls, and writes nothing.

## Observed verification

- 12 new tests (`tests/test_plugin_hermes_delegation.py`) + widget-frame contract additions; adjacent suites (`test_tui_widget_pack`, `test_hud`, `test_runtime_reader_filesystem_faults`, `test_plugin_capabilities`) green in a clean worktree.
- Live render observed against a real `~/.hermes` mid-delegation: reader returned the two running review lanes + one lingering done row with model `gpt-5.6-sol`, effort `medium`, category `inherit`, tokens/tok/s/cache% populated — matching the session the owner screenshotted.
- Full suite running in a clean worktree; result will be posted before merge.

## Risks

- Liveness windows (150s recent, 15m linger) are heuristics over log mtime and usage recency; a child idling longer than 150s between events briefly reads "done", then returns to "running" on its next event.
- The category label is attribution, not an observed routing record — documented in-module, and `inherit` deliberately outranks chain matches.

### DIFF
--- a/src/omh/tui_widgets/omh-status.mjs
+++ b/src/omh/tui_widgets/omh-status.mjs
@@ -68,9 +68,14 @@ export default function register(sdk) {
     if (!active) return 'ready'
     const running = Number(agents.running) || 0
     const blocked = Number(agents.blocked) || 0
+    const done = Number(agents.completed) || 0
+    // Lingering just-finished subagents keep the block alive without live
+    // work; "2 done" is the honest label there, not "0 agents".
+    if (!running && !blocked && done) return `${done} done`
     const parts = [plural(Number(agents.active) || 0, 'agent')]
     if (running) parts.push(`${running} running`)
     if (blocked) parts.push(`${blocked} blocked`)
+    if (done) parts.push(`${done} done`)
     return parts.join(' · ')
   }
   const readHud = () => new Promise(resolve => {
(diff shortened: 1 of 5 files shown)


sionic-khope:
Full-suite verification (clean worktree, commit 3c2ae317): **6,832 tests, 1 failure** — the pre-existing `test_global_tarball_launchers` (`bun` not installed on this machine), unrelated to this change. All five `docs --check` byte gates, ruff, and compileall green. Live-verified on the owner's running TUI: subagent rows render under the composer (spinner + ✓ done rows observed in session screenshots).

PR #1004, opened 2026-08-18
fix(hud): stop repainting the docks when nothing changed so drag-copy works

sionic-khope:
## Capability

The `[OMH]` docks under/over the composer are drag-copyable like the rest of the TUI: an unchanged snapshot no longer repaints them, so terminal selection over those lines survives.

## Motivation

The owner reported the bottom `[OMH]` area could not be drag-copied while every other region could. Cause: the widget called `updateWidget` on every 2-second poll regardless of change — each repaint of those exact lines clears an in-progress selection — and the shimmer subscription kept repainting at animation rate while rows lingered. (A first diagnosis blamed DEC mouse tracking and toggled `display.mouse_tracking`; that was wrong — other regions copied fine, so capture was never the difference — and it was reverted on the owner's machine.)

- `applySnapshot` skips byte-identical snapshots entirely (no `updateWidget`, no repaint).
- The shimmer-driven activity component mounts only while a RUNNING row needs the spinner; done/blocked-only states render static rows with no clock subscription.
- The delegation reader freezes a finished child's `elapsed_seconds` at its last activity so a lingering done row serializes identically poll after poll (a done task also stops visually aging).

Live rows still animate — quiescence applies exactly when the payload is unchanged, which running rows never are.

## Observed verification

- New reader contract: a finished row is byte-stable across polls with different clocks; widget-frame contract pins the snapshot skip and the animated/static split.
- Affected suites green (single failure = the known stale-worktree environmental case, green in a clean worktree); node module-load of the widget; ruff; compileall. Full suite running in a clean worktree — result posted before merge.

## Risks

- Any future now()-derived field on a finished row would reintroduce the 2s repaint (recorded as a Directive in the commit).
- Live drag-copy on the owner's terminal is verified after the TUI restart that also activates the routing tool.

### DIFF
--- a/src/omh/tui_widgets/omh-status.mjs
+++ b/src/omh/tui_widgets/omh-status.mjs
@@ -219,14 +219,7 @@ export default function register(sdk) {
     )
   }
 
-  function LiveActivity({ columns, mainRows, rows, t, tick }) {
-    // Mounted only while live rows exist. The SDK shimmer clock is bounded —
-    // it stops advancing animateMs after MOUNT — so the old top-level
-    // useShimmerPhase(30_000) froze thirty seconds into a session and the
-    // spinner then jumped once per poll. A fresh mount per activity burst
-    // restarts the window, and thirty minutes bounds the render cost of one
-    // very long wave; the poll tick still nudges frames past that.
-    const frame = useShimmerPhase(1_800_000) + tick
+  function ActivityRows({ columns, frame, mainRows, rows, t }) {
     return h(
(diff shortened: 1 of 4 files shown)


sionic-khope:
Full-suite verification (clean worktree, branch head): **6,850 tests, 1 failure** — the pre-existing `test_global_tarball_launchers` (`bun` not installed on this machine), unrelated to this change. Widget module-load, ruff, compileall green.

PR #1098, opened 2026-08-24
revert(release): remove the automated stable cadence

sionic-khope:
### What Changed

- Removed the automated stable release cadence entirely: `.github/workflows/auto-release.yml`, `tools/package_manager/bump_version.py`, `tests/test_auto_release.py`, the `docs/DISTRIBUTION.md` cadence section, the `docs/RELEASE.md` pointer, and the auto-release entry in the pinned-actions gate.
- `docs/DISTRIBUTION.md` now states plainly that releases are cut by hand and that `main` sitting ahead of the published packages is expected, with a one-line note that the auto-tagging attempt was removed.

### Why This Exists

- The cadence moved the public version without a person deciding a release was warranted. It cut v1.0.8 and v1.0.9 within a day, taking the published packages from 1.0.6 to 1.0.9 while the repo had been sitting at 1.0.7. Owner decision: revert it, remove the bump tool too.
- The workflow was disabled in the Actions UI first, so no further runs were possible before this landed.

### User / Operator Impact

- No user-visible change. `tools/` is not shipped in the wheel and workflows are not part of any published artifact, so no new release is needed to carry this.
- Maintainers are back to bumping versions and pushing `vX.Y.Z` tags by hand.

### How It Works

- Pure removal plus a documentation rewrite. No change to `release.yml`'s ordering, guards, immutability checks, or the beta contract.

### What Is Deliberately Kept

- **Version 1.0.9 on every surface.** npm 1.0.9 is immutable, the Homebrew tap refuses a downgrade, and the GitHub release exists — the repo has to match what is published. Reverting the version to 1.0.7 would break parity with all three and block the next release. (`1.0.7.1` is not an option either: it fails npm's semver validation, this repo's `X.Y.Z` metadata pattern, the tag regex in `release.yml`, and the tap's monotonic version guard.)
- **The bash 3.2 empty-array fix in `release.yml` and its regression test.** That is an independent bug in the manual release path — reverting it would re-break every stable release, automated or not.

- Deleted: `.github/workflows/auto-release.yml`, `tools/package_manager/bump_version.py`, `tests/test_auto_release.py`.
- Edited: `docs/DISTRIBUTION.md`, `docs/RELEASE.md`, `tests/test_homebrew_distribution.py`.

### Affected Surfaces

- [x] Not executor-specific

## Validation

- [x] `PYTHONPATH=tests uv run python -m unittest tests.test_homebrew_distribution tests.test_distribution_packages tests.test_distribution_surfaces tests.test_release_smoke` (80 tests OK)
- [x] `uv run --group lint ruff check src tests tools packaging`
- [x] `git diff --check`

### Observed Evidence

- `grep -rn "auto-release|auto_release|bump_version|Automated stable cadence"` across `src tests docs .github` returns nothing.
- `Auto Release` workflow state confirmed `disabled_manually` before this change.
- Published state left intact and verified: GitHub release `v1.0.9` with its wheel asset, npm `latest` 1.0.9, tap formula `version "1.0.9"`.

- No release is cut by this PR; nothing published changes.

## Risk

- Removal only. The one behavioral consequence is the intended one: no workflow will tag `main` on its own again.

## Compatibility / Rollout

- The `v1.0.8` tag remains as a tag that never published anything (the run failed before its first write). It is harmless; `releases/latest` and the tap both resolve 1.0.9.

## Release / Claims

https://claude.ai/code/session_01GRSYnsBfXd5hALXNA6333f


sionic-khope:
Closing: the concern was that this ran on every merge. It does not -- the trigger is a single daily cron (07:30 KST) that only cuts when main has moved past the last tag, so at most one release per day regardless of merge volume. Keeping version 1.0.9 and the feature in place.

PR #1213, opened 2026-08-30
feat(quality): classify prompt/schema overlap on the omh tool surface

sionic-khope:
### What Changed

- New `src/quality/schema_overlap.py`: a deterministic probe that finds text in an omh tool description already carried by its own JSON schema, and classifies each hit as a reviewed `keep` (with a named reason) or a `prune candidate`.
- New `builtin_tool_schemas()` / `BUILTIN_TOOL_NAMES` in `src/plugin_bundle/omh/tools/__init__.py`, so a check can walk the tool surface without a Hermes registration context.
- `docs/ADDING-A-SKILL.md` section 7: the bucketing table (what is a prune candidate, what is always a keep) and the three rules the measurement cannot apply.
- `tests/test_schema_overlap.py`: the gate plus both directions of all five rules.

### Why This Exists

- Every tool description is paid for in every context window that loads the toolset, and part of it is often reconstructible from `(name, schema, blank outline)` — parameter types, enum members, clamp ranges, defaults, `required`.
- The valuable half of that observation is the word "candidate". Much of what looks redundant is incident scar tissue: somebody wrote that sentence because a model got it wrong once, and the schema has not started saying it since. So the deliverable is a recorded verdict, not an edit.
- The gate is therefore not "no overlap" but "every overlap is classified". A new description that restates its schema gets a human decision instead of sliding in unread.

### User / Operator Impact

- No runtime behavior changes; no tool description was edited in this PR.
- Authors adding or rewording an omh tool now get a test failure naming the finding id and the two verdict forms to choose from.

### How It Works

- Five rules over each property spec: bare type restatement, enum listing, clamp restatement, default restatement, and unconditional `required`.
- A hit becomes a `keep` when the description carries what the schema cannot — per-member semantics, what a bound means, which way a default points, or a requirement conditional on another field. Otherwise it stays a `prune candidate`.
- The enum floor is measured rather than guessed: the two descriptions that teach what each member does carry 62 and 287 characters of text beyond the member names; the one that merely lists them carries 35. `ENUM_SEMANTIC_RESIDUE_MIN_CHARS = 40` sits between.
- `stale_overlap_decisions()` also fails a decision left behind after its description was rewritten, so the registry cannot accumulate claims about text that is gone.

- `src/quality/schema_overlap.py` (new) — `omh_schema_overlap/v1`.
- `src/plugin_bundle/omh/tools/__init__.py` — collector and name list.
- `docs/ADDING-A-SKILL.md` — new section 7.
- `tests/test_schema_overlap.py` (new).

Measured over all 14 registered tools: **4 findings, 3 reviewed keeps, 1 prune candidate**. That ratio is the expected shape — self-documenting flag-style tools prune heavily, DSL and capability tools barely, and omh's tools are the second kind. The one candidate (`omh_role.action`) is left in place with its reason recorded.

### Affected Surfaces

- [ ] Hermes chat or wrapper
- [ ] Codex
- [ ] Claude Code
- [ ] Generic executor
- [x] Not executor-specific

## Validation

- [x] `git diff --check`
- [ ] `omh --help`
- [ ] `omh --omh-home /tmp/omh-smoke ... release hermes-smoke ...`
- [ ] Relevant workflow-learning review queue smoke: no learning contract changed.
- [x] Relevant dry-run or smoke command: `uv run python -m omh.cli release drift` (No drift, 18 checks passed); `docs ulw-inventory --check`; `docs ulw-site --check`
- [ ] Manual Hermes/TUI check, or explicit reason it was not run: not run — no tool description, schema, or handler changed, so the plugin surface Hermes loads is byte-identical.

### Observed Evidence

- `PYTHONPATH=tests uv run python -m unittest tests.test_schema_overlap` -> Ran 14 tests, OK.
- Probe output over the live tool surface:
  ```
  omh tool prompt/schema overlap: 4 finding(s), 1 prune candidate(s), 3 reviewed keep(s).
    [OVERLAP_ENUM_RESTATED] omh_delegate_route.action -> keep (member_semantics)
    [OVERLAP_ENUM_RESTATED] omh_role.action -> prune_candidate
    [OVERLAP_ENUM_RESTATED] omh_todo.action -> keep (member_semantics)
    [OVERLAP_BOUND_RESTATED] omh_todo.items[].depth -> keep (bound_meaning)
  ```
- `PYTHONPATH=tests uv run python -m unittest discover -s tests` -> Ran 8359 tests, failures=21 errors=7. The identical 28 were measured on unmodified `origin/main` in this worktree before any edit; they are the known cross-harness `.claude`-path failures caused by running inside a linked worktree under `.claude/worktrees/`, unrelated to this change.
- `uv run python -m compileall -q src tests` -> clean. `ruff check src tests` -> All checks passed. `git diff --check` -> clean.
- Five `docs ... --check` byte gates and `release drift` -> all ok.

- `harness validate`, `release checklist`, `release hermes-smoke`, install smoke: no installer, harness manifest, or release-channel surface changed.
- The model-based recovery probe the upstream audit describes (does a model reconstruct this text from name + schema alone?): OMH makes no model calls, so cross-sample and cross-model recovery is out of scope by boundary, not by omission. The upstream caveat that a public repo may be training data — making recovery recitation rather than inference — is the reason that half is not worth faking locally.

## Risk

- Low. Nothing in the shipped plugin bundle changed except an additive collector function whose imports stay inside the function body, matching how `register()` already defers them.
- The gate can fail a future PR that adds a tool. That is the intent, and the failure message names the finding id and the two verdict forms.

## Compatibility / Rollout

- Additive. No tool description, schema, or handler changed.

## Release / Claims

- [x] Release-channel impact considered (`local`)

## Follow-Up

- `omh_role.action` is recorded as a prune candidate, not cut. Cutting it needs a `git blame` pass on the line first: it predates the role catalog split and may be scar tissue from a model calling `read` without a `role`.

### DIFF
--- a/docs/ADDING-A-SKILL.md
+++ b/docs/ADDING-A-SKILL.md
@@ -130,6 +130,40 @@ Two rules the gate cannot check for you:
   text the remaining words are the payload; an undeclared loss is a silent
   regression, and a single-digit win is not worth re-reading every rule for.
 
+## 7. Tool-facing text: a prune candidate is never a delete
+
+The same question applies to the `src/plugin_bundle/omh/tools/` descriptions,
+with a sharper test available: a JSON schema sits beside each one, so text a
+reader could reconstruct from `(name, schema, blank outline)` is measurable
+rather than argued. `tests/test_schema_overlap.py` runs that measurement over
+every registered tool; the rules and the reviewed verdicts live in
+`src/quality/schema_overlap.py`.
+
+What the probe finds is a **prune candidate**, never an automatic delete:
+
+| Bucket | Examples |
+| --- | --- |
+| Prune candidate | Parameter names and types. `required`. Value examples that only restate an enum. Clamp ranges already declared as `minimum` / `maximum`. |
+| Keep, always | Defaults **and their direction** — `gitignore: true` does not say "respects gitignore". Routing and escalation rules. Exact output shape. Worked anti-patterns. Constraints the type system cannot express, such as a field required only for one action. |
+
+Three rules the measurement cannot apply for you:
+
+- **`git blame` the line before cutting it.** Much of what restates a schema is
+  incident scar tissue: somebody added that sentence because a model got it
+  wrong once, and the schema has not started saying it since.
+- **One sample is noise.** A single overlap hit is a question for review, not a
+  finding. The gate is that every hit carries a verdict, not that the count is
+  zero.
+- **Decide what ships before you compress it.** Scope first, density second; a
+  tighter description of a parameter that should not exist is still a worse
+  tool.
+
+The gate fails when a finding has no reviewed verdict, and it fails with the
+finding id and paste-ready instructions. Add the id to
+`REVIEWED_OVERLAP_DECISIONS` with `keep — <which bucket>` or
+`prune candidate — <what git blame showed>`. Self-documenting flag-style tools
+prune heavily; DSL and capability tools barely, so most verdicts are `keep`.
+
 ## Acknowledgements
 
 Domain taxonomy adapted from revfactory/harness (https://github.com/revfactory/harness), Apache License 2.0, Copyright 2025 robin.
--- a/src/plugin_bundle/omh/tools/__init__.py
+++ b/src/plugin_bundle/omh/tools/__init__.py
@@ -1,1 +1,68 @@
 from __future__ import annotations
+
+from typing import Any
+
+# Every tool name registered by `register()` in the parent package, in
+# registration order. Kept as data so a check can walk the tool surface
+# without importing the Hermes registration context; parity with the
+# `register()` body is a test, not a convention.
+BUILTIN_TOOL_NAMES: tuple[str, ...] = (
+    "omh_capabilities",
+    "omh_context",
+    "omh_delegate_route",
+    "omh_gather_evidence",
+    "omh_hud",
+    "omh_interact",
+    "omh_memory",
+    "omh_probe",
+    "omh_recommend",
+    "omh_role",
+    "omh_run_summary",
+    "omh_source_trust",
+    "omh_status",
+    "omh_todo",
+)
+
+
+def builtin_tool_schemas() -> tuple[dict[str, Any], ...]:
+    """Return every registered omh tool schema, name-sorted.
+
+    Imports live inside the function for the same reason `register()` does it:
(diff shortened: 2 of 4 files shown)


sionic-khope:
Fixed the `test-windows` failure in 6eddb57.

**Cause.** Two reads in the new coverage test called `Path.read_text()` with no encoding, which uses `locale.getencoding()` — cp1252 on Windows, where 0x81 is undefined. The byte is real content, not corruption: `run_summary_tool.py` documents its localized end-of-run summary in Korean, and 0x81 is a UTF-8 continuation byte inside it at LF offset 177. A Windows checkout converts the 5 preceding newlines to CRLF, shifting it to exactly **182** — the position CI reported.

**Reproduced locally** (no Windows runner needed) by decoding that file's CRLF form as cp1252:

```
REPRODUCED: UnicodeDecodeError: character maps to <undefined> byte 0x81 at position 182
utf-8 decode of the same bytes: ok
```

**Fix.** `encoding="utf-8"` on both sites, plus a class docstring recording why: every read in that class walks committed repository sources rather than a self-written temp fixture, which is the one case where the default encoding matters. The ~20 other unencoded `read_text()` calls in `tests/` read ASCII-only fixtures the tests wrote into temp dirs, so they cannot hit this; sweeping them would put unrelated files in a CI-fix commit.

**Verification.**

- `PYTHONPATH=tests uv run python -X warn_default_encoding -W error::EncodingWarning -m unittest tests.test_schema_overlap` -> Ran 14 tests, OK. The same command against the pre-fix file errors at exactly the two sites, so the check has teeth.
- `PYTHONPATH=tests uv run python -m unittest tests.test_schema_overlap tests.test_plugin_capabilities` -> Ran 19 tests, OK.
- `PYTHONPATH=tests uv run python -m unittest discover -s tests` -> Ran 8359 tests, failures=21 errors=7. One earlier run showed errors=8; I chased it rather than assuming, captured the sorted FAIL/ERROR set on this branch and on current `origin/main` (8f96d3d1), and diffed them — **byte-identical, 28 lines**. The transient extra error was a flake in the same subprocess-sensitive cross-harness family, not a regression from this branch.
- `uv run python -m compileall -q src tests`, `uv run --group lint ruff check src tests`, `git diff --check` -> clean.

Rebased/merged nothing: the branch still merges cleanly with `origin/main` after #1212 and #1214 landed (`git merge-tree` clean, PR reports MERGEABLE), and the touched files are disjoint from both.

PR #1221, opened 2026-08-31
feat(tui): selectable TUI themes with an arrow-key picker

sionic-khope:
### What Changed

OMH's terminal identity gains **selectable colour themes** and a user-facing
switch. Four palettes now ship — `sky` (the current turquoise look, still the
default), `amber`, `crimson`, and `mono` — installed together at setup/update
time and selected through a new `omh theme` command group.

- `omh theme` (bare) opens an **arrow-key picker with a live colour preview**:
  up/down (or `j`/`k`) move the cursor, each theme paints a sample of its own
  palette as you pass over it, Enter applies, `q`/Escape/Ctrl-C cancels
  without writing anything.
- `omh theme list` prints the same information as a plain listing, always.
- `omh theme use <name>` selects one directly, with `--dry-run` and `--json`.
- `omh theme status` reports the active skin, whether OMH owns it, the Hermes
  config path, and the per-file managed-install state.

### Why This Exists

The OH-MY-HERMES look was a single sky-blue skin with no supported way to
change it. Changing it meant knowing that `display.skin` exists, knowing the
skin name `omh`, and hand-authoring a YAML in `$HERMES_HOME/skins/` — which
also silently opted that file out of every future refresh. A theme is also the
one setting whose name tells you nothing: "crimson" is not a decision anyone
can make from a list, which is why the bare command previews instead of
listing.

Three defects surfaced while building it and are fixed here:

1. **The identity check compared against one name.** `display.skin != "omh"`
   meant an operator on any OMH-owned theme read as having chosen a *foreign*
   skin. The TUI verdict reported the branding as declined while it was
   working, and the setup/update branded-TUI prompt would have asked them to
   switch back — with a "yes" silently erasing their theme.
2. **The skin manifest was single-file.** One shared record meant a
   hand-edited theme file would have made the whole directory look
   user-authored and frozen the default skin too.
3. **`ui_label` equalled `banner_title`** in the default skin. `ui_label`
   paints the user's own prompt rows in the transcript, so those rows blended
   into other label-coloured chrome.

### User / Operator Impact

After this change an operator can:

- Run `omh theme`, arrow through four palettes seeing each one painted, and
  press Enter — no YAML, no skin names, no docs lookup.
- Switch instantly and offline; every palette is already on disk.
- Keep that choice. `omh setup` and `omh update` leave any OMH-owned theme
  alone forever, exactly as they already left a foreign skin alone.
- Hand-edit one theme file and keep it, without that decision freezing the
  other three.

Nothing changes for someone who does not run `omh theme`: the default is still
`omh`, and `display.skin: omh` is still what a fresh config gets.

### How It Works

**Theme registry** (`src/omh/skin_pack.py`). `SKIN_THEMES` is the ordered
source of truth: short name (`crimson`) ↔ Hermes skin name (`omh-crimson`) ↔
filename ↔ one-line summary. `theme_for_name()` resolves short names, the
`default`/`sky` aliases, and full skin names; `theme_for_skin_name()` matches
config values by skin name **only**, so a foreign skin literally named `amber`
is never read as ours. `is_omh_skin_name()` is the ownership predicate every
"is the OMH identity active?" question now goes through.

**Manifest v2.** `omh_skin_manifest/v1` described one file; the new
`omh_skin_manifest/v2` carries a per-filename sha256 map. The reader migrates
a v1 document to a single-entry map rather than discarding it — discarding it
would classify an already-installed `omh.yaml` as user-authored and freeze it
at the version it shipped with, on every machine installed before themes
existed. Install and uninstall decide each file independently and report
per-file statuses alongside an aggregate for the setup summary. Symlink
rejection, atomic writes, and "never overwrite what OMH did not write" are
unchanged.

**Selection** is Hermes-side and unchanged in mechanism: one scalar,
`display.skin`. `omh theme use` takes the forcing `activate_omh_skin` path
because running it *is* the explicit consent the managed-artifact rule asks
for; `ensure_omh_skin` stays the narrow unset-only default writer. The picker
and `theme use` share one `_apply_theme()` function so the picker cannot drift
into a second, laxer way of writing the key.

**The picker** (`src/commands/theme_picker.py`) is pure stdlib. Every escape
sequence lives in one block of module constants. Three seams — `read_key`,
`write`, and the frame width — are arguments, so tests drive navigation by
feeding key tokens and read the frames back out of a list; the default
`read_key` is the only part that touches a tty. It writes nothing but frames:
applying the selection is the caller's job, so a cancelled picker cannot have
touched the Hermes config by construction. Preview colours are read from each
theme's own YAML (`skin_colors()`) and emitted as 24-bit `ESC[38;2;r;g;bm`, so
a palette edit cannot drift from what the picker shows.

**Windows takes the documented fallback.** There is no `termios` there, and a
`msvcrt` raw-key path is more terminal-behaviour risk than a cosmetic picker
earns. `picker_available()` returns False and bare `omh theme` prints the same
plain listing `omh theme list` prints. `omh theme list`, `use`, and `status`
are fully supported on every platform. The same fallback covers `--json`, a
pipe, `TERM=dumb`, and `OMH_NO_TUI=1` (the two env escapes match the keyboard
menu `omh setup` already ships).

**Palettes.** Every theme keeps the identity `branding`, spinner block, and
logo geometry, and omits the `background` token for the reason `omh.yaml`
already documents. Two colour decisions are noted in the YAML comments:
crimson pulls its accent warm (ember `#FF6A3D`) and pushes `ui_error` the
other way across the red point (rose `#FF3D6E`) so a failure never reads as
ordinary accent text; mono keeps ok/warn/error chromatic, because a greyscale
palette that also greys out failure removes the one distinction a terminal
user needs.

| File | Change |
| --- | --- |
| `src/omh/skin_pack.py` | Theme registry, `available_skins()`/`skin_payload(name)`/`skin_colors()`/`hex_to_rgb()`/`is_omh_skin_name()`, manifest v2 + v1 read migration, per-file install/uninstall, `installed_skin_report()` |
| `src/omh/skins/omh-{amber,crimson,mono}.yaml` | New shipped palettes (packaged automatically by the existing `"omh.skins" = ["*.yaml"]` glob) |
| `src/omh/skins/omh.yaml` | `ui_label` brightened off `banner_title` |
| `src/commands/theme.py` | New `omh theme {list,use,status}` group |
| `src/commands/theme_picker.py` | New arrow-key picker and truecolor preview |
| `src/commands/main.py` | Registers `_add_theme_commands` |
| `src/commands/setup.py` | Identity-active check and the accept-branded-TUI path use the ownership predicate, so a chosen theme is never clobbered |
| `src/install/config_adapter.py` | `ensure_omh_skin` names an OMH theme distinctly from a foreign skin; both writers default to `SKIN_NAME` |
| `src/maintenance/hermes_tui.py` | `tui_identity_verdict` reports an OMH theme as identity-active, not as a foreign choice |
| `src/commands/language.py` | Four-locale TUI prompt/note no longer claim sky-blue is the only look |
| `docs/INSTALLATION.md`, `README.md` | New "TUI Themes" subsection; README highlight row |

Schema versions added: `omh_skin_manifest/v2`, `omh_theme_list/v1`,
`omh_theme_change/v1`, `omh_theme_status/v1`.

### Affected Surfaces

- [x] Hermes chat or wrapper
- [ ] Codex
- [ ] Claude Code
- [ ] Generic executor
- [x] Not executor-specific

## Validation

- [x] `git diff --check`
- [x] `omh --help`
- [ ] `omh ... release hermes-smoke --install-path setup --include-command-smoke` — not run; this change installs no command package and touches no install path
- [ ] Workflow-learning review queue smoke — not applicable, no learning contract changed
- [x] Relevant dry-run or smoke command: `omh theme use crimson --dry-run --json`, plus a real-pty drive of the picker
- [ ] Manual Hermes/TUI check — see Not Tested

### Observed Evidence

- **Full suite**: `PYTHONPATH=tests uv run python -m unittest discover -s tests`
  → 8514 tests, 21 failures + 7 errors. **All 28 are in the
  `test_cross_harness_adapter*` family**, failing with `unsafe_read_root` —
  the known local artifact of running under a `.claude/worktrees/` path.
  Neither file references skins or themes
  (`grep -li "skin\|theme"` matches nothing in either). Baseline before this
  branch was the same 28. **CI is authoritative** for these.
  Non-`cross_harness_adapter` failures: **0**.
- **Byte gates**, all `ok: true`: `docs workflows --check` (155/155 tap
  skills), `docs roles --check`, `docs capability-families --check`,
  `docs ulw-inventory --check`, `docs ulw-site --check`.
- `uv run --group lint ruff check src tests` → `All checks passed!`
- `uv run python -m compileall -q src tests` → exit 0
- `git diff --check` → exit 0
- `PYTHONPATH=tests uv run python -m unittest tests/test_broad_exception_policy.py`
  → 6 tests OK (no new broad `except`; the new code uses
  `except (OSError, json.JSONDecodeError)` and `except ImportError`).
- **New/changed tests**: `tests/test_skin_pack.py` 20 → 32,
  `tests/test_theme_command.py` +11, `tests/test_theme_picker.py` +22,
  `tests/test_hermes_tui_preflight.py` 20 → 22, `tests/test_cli.py` +2.
  Coverage: registry ↔ packaged-files parity, per-theme payload, structural
  key-set parity against `omh.yaml`, logo width looped over all four themes,
  no `background` token in any theme, `ui_label` ≠ `banner_title` and brighter
  than every dim token, manifest v2 write, v1 read migration, per-file
  `kept_unmanaged` isolation, uninstall symmetry, `ensure_omh_skin` leaving
  `omh-amber` alone while still seeding an unset config and still ignoring
  `ares`, verdict on an OMH-owned non-default skin, `theme list/use/status`
  happy paths, invalid name, `--dry-run`, `--json`, foreign-skin cases, picker
  navigation/cancel/wrap, non-TTY degradation, `NO_COLOR` emitting zero escape
  bytes, per-theme preview hex values, and the absent-`termios` fallback.
  Negative cases accompany every positive one.
- **Real-pty drive of the picker** (`pty.fork`, sending `ESC[B ESC[B CR`):
  opening frame drawn with the cursor hidden; two in-place repaints observed
  (`ESC[9A` rewind + `ESC[2K` per line, no scroll); the preview switched from
  sky's `38;2;0;206;209` to amber's `38;2;255;179;0` to crimson's
  `38;2;255;106;61`; Enter wrote `display.skin: omh-crimson` and installed
  `omh-crimson.yaml`; the cursor was restored. This also caught a real bug —
  reading keys through `sys.stdin.read(1)` buffers the tail of an arrow
  sequence in userspace where `select` cannot see it, so every arrow cancelled
  the picker. Reads now go through `os.read` on the descriptor, with a
  regression test over a pipe.
- `omh --help` lists `theme`.

- **How the four palettes actually look on a real terminal is
  `prepared_not_observed`.** The hex values, contrast decisions, and the
  crimson accent/error separation were chosen by reasoning about hue and
  brightness, not by looking at a terminal. Every colour is verified to be a
  parseable hex, structurally complete, distinct where the tests require it,
  and correctly plumbed into the preview — none of that is a claim that they
  are attractive or legible on any particular display. A human should look at
  all four before this is called done.
- **The Windows fallback is not exercised on Windows.** It is covered by a
  unit test that patches `termios` to `None`, which proves the branch, not the
  platform.
- **No manual Hermes/TUI session was run**, so "the banner recolours" is
  inferred from Hermes' skin engine contract (a YAML in `$HERMES_HOME/skins/`
  themes the CLI, TUI, and GUI), not observed. The HUD widget resolves colours
  only through the host theme object and was deliberately not touched.
- `release hermes-smoke --include-command-smoke` against a live install: not
  run; no install path or command package changed.

## Risk

- **Low, and the default is unchanged.** `display.skin: omh` remains what a
  fresh config gets, `omh.yaml` is byte-identical except for the one
  `ui_label` value, and the existing assertions on `"  skin: omh\n"` in
  `tests/test_cli.py` and `tests/test_plugin_distribution.py` were verified to
  still pass rather than weakened.
- **`install_skin`/`uninstall_skin` return shape changed** from a flat dict to
  an aggregate plus a `skins` list. The aggregate `status` key is preserved
  with the same vocabulary, and no caller reads the other keys, so the setup
  summary and uninstall payload are unaffected — but an out-of-tree consumer
  reading `result["path"]` now gets the skins directory, not one file.
- **Manifest v2 is not readable by an older OMH build.** A downgrade would see
  an unrecognised schema, treat all four files as user-authored, and stop
  refreshing them (it would not delete anything). Forward migration is
  covered; backward is not, and matches how the widget/plugin manifests
  already behave.
- **The picker's terminal handling is the genuinely new risk surface.** Raw
  mode is entered and restored through a context manager that always runs its
  `finally`, the cursor is always shown again, and the key loop is bounded so
  a pathological reader cannot spin inside a terminal in raw mode. Any
  environment it cannot drive falls back to the plain list rather than
  degrading.
- **`ui_label` moved in the default theme.** This is a visible change to
  existing users' prompt rows. It is deliberate and small (#7FDBFF →
  #9FE8FF), and noted in the YAML.

## Compatibility / Rollout

- Reaches a machine through `omh update` plus a Hermes restart, like every
  other managed artifact. The new YAMLs ship automatically under the existing
  `"omh.skins" = ["*.yaml"]` package-data glob; no packaging change.
- A theme change applies on the next Hermes start — `omh` execs `hermes` as a
  child and Hermes reads its skin at startup. The command says so, reusing the
  existing `tui_verdict_ready` copy rather than inventing new wording.
- Existing installs keep their v1 manifest until the next `omh update`, which
  migrates it in place.
- No new dependencies. Core `omh` still makes no LLM, API, or network calls;
  the picker reads a terminal and the command writes one config key.

## Release / Claims

- [x] Release-channel impact considered — normal `stable` feature, no channel
      or version gating
      observed — the palettes' visual quality is explicitly
      `prepared_not_observed`
- [x] Prepared handoffs or plans are labeled `prepared_not_observed`, never as
      execution, review, CI, or merge evidence
- [x] Evidence contains no credentials, raw prompts, raw platform events,
      transcripts, or unrelated private data
- [x] Known manual checks are listed — a human eye on all four palettes, and a
      Windows run of the fallback

## Follow-Up

- Have someone open each theme in a real terminal and adjust the hex values;
  that is the one thing this PR cannot self-verify.
- If Windows users ask for the picker, the `msvcrt` reader slots behind the
  existing `read_key` seam without touching `run_picker`.
- Adding a fifth theme is a registry row plus a YAML; the parity, logo-width,
  `background`-token, and label-contrast guards fail with a clear message if
  any piece is missing.


sionic-khope:
## Follow-up: skin manifest self-heal (commit `1c7e0614`)

**Live-machine finding** while smoke-testing this branch against a real managed install: the owner's `~/.hermes/skins/omh.yaml` was byte-identical to the shipped template, but the on-disk **v1** manifest still recorded an older sha (`1420e408…` vs the actual `e99be0e8…`) — a past update refreshed the file without refreshing its record.

### Why that mattered

Ownership proved *only* by the manifest record read that machine's own OMH-written skin as user-authored:

- `installed_skin_report` → `unmanaged`, so `omh theme status` misreported it
- `install_skin` → `kept_unmanaged` **forever**
- no future template change would ever reach such a machine — **including the `ui_label` brightening in this very branch**

### Fix

`_is_managed_file` now accepts a second, equally sufficient proof: **the file's bytes match the currently shipped template for that filename.** That is safe precisely because the bytes are ours — adopting the file cannot destroy anything a user wrote, since overwriting it with the template is a no-op. The install path then rewrites the record from the current payload, healing the machine to `omh_skin_manifest/v2`.

`uninstall_skin` uses the same predicate, so a file install would adopt is a file uninstall takes back rather than orphaning.

**A file matching neither the record nor the template is genuinely user-edited and stays `kept_unmanaged`, exactly as before.**

### Observed evidence

Reproduced the owner's exact state locally (template-identical file + stale v1 record):

```
report BEFORE: ['managed', 'missing', 'missing', 'missing']
per-file     : {'omh.yaml': 'unchanged', 'omh-amber.yaml': 'installed', ...}
manifest     : omh_skin_manifest/v2 | omh sha -> 122577bf   (heals to the real hash)
report AFTER : ['managed', 'managed', 'managed', 'managed']
next release : {'omh.yaml': 'installed', ...}   # a changed template now lands
file updated : True
```

New tests in `tests/test_skin_pack.py` (20 → **36**), positive and negative together:

- `test_a_stale_manifest_over_our_own_bytes_self_heals` — reported `managed` before any write; manifest healed to v2 with the correct sha
- `test_a_healed_machine_receives_the_next_template_change` — a patched next-release template actually lands on the file (the point of healing is the *next* release, not the record)
- `test_a_genuinely_edited_file_is_still_never_adopted` — the negative half: neither proof matches → `unmanaged` / `kept_unmanaged`, content preserved, no manifest record written
- `test_uninstall_takes_back_a_file_it_would_have_adopted` — install/uninstall symmetry
- the pre-existing v1→v2 migration test still covers a *valid* v1 record

### Gates re-run

- Full suite: **8518 tests**, 21 failures + 7 errors — all 28 in `test_cross_harness_adapter*` (`unsafe_read_root`, the `.claude/worktrees` path artifact). Non-`cross_harness_adapter` failures: **0**. CI authoritative.
- `ruff check src tests` → All checks passed · `compileall` → 0 · `git diff --check` → 0
- All five `docs … --check` byte gates → 0
- `test_broad_exception_policy.py` → 6 OK

### Not tested

Not re-run against the owner's live machine after the fix — the repro is a faithful local reconstruction of the reported state, not the machine itself. Worth one `omh update` + `omh theme status` on that machine to confirm it flips to `managed`.

PR #1275, opened 2026-09-02
feat: add user-configurable model price overrides

### LINKED ISSUE #1274: cost: token prices are hardcoded in source with no user override, against the repo's own config-vs-source rule
## What breaks

Token prices are a hardcoded dict literal in shipped source (`APPROX_PRICE_PER_MTOK` in `src/plugin_bundle/omh/hermes_delegation.py`), with no way for a user to override them. A user cannot correct a price without editing our source.

This contradicts the repo's own stated rule. `docs/INSTALLATION.md` says it outright for the neighbouring document: **"Chain customization is a config edit, not a source edit."** Every other per-user-variable routing fact already lives under `~/.omh/routing/` — `model-chains.json` (`mixture_chain_overrides/v1`), `providers.json` (`provider_entitlements/v1`), `model-providers.json` (`model_provider_routes/v1`). Prices do not.

Prices are exactly the kind of thing that differs per user and drifts over time:

- a gateway (OpenGateway, OpenRouter, any relay) applies its own markup
- enterprise and committed-use rates are not list prices
- free tiers, promos, and subscription plans bill nothing per call
- vendors change list prices, and our table has already gone stale once — the source comment says so: *"Earlier entries here were stale."*

## Supporting problems in the same table

- **Coverage hole.** `grok-code-fast` ships in the recommendation catalog and has a provider-family entry, but no price row — so a run on it silently produces no cost at all, forever.
- **Uneven provenance.** 8 of 19 entries carry no date and no source. Two blocks are dated and cited (`docs.claude.com pricing, 2026-09`, `docs.z.ai pricing, 2026-08`); the rest are bare numbers a reader cannot check or age out.
- **Nothing verifies currency.** The existing tests assert the *arithmetic* is self-consistent given the literals; none of them can notice a number that no longer matches any vendor's sheet.

## Target behaviour

A validated price-override document under `~/.omh/routing/`, modelled on the two siblings that already exist for exactly this shape (alias → structured value, strict atomic parse, `(value, status)` return, `absent`/`applied`/`invalid: <reason>`):

- `mixture_chain_overrides/v1` — `MIXTURE_CHAIN_OVERRIDES_SCHEMA_VERSION`, `mixture_chain_overrides_path`, `load_mixture_chain_overrides`, `parse_mixture_chain_overrides`, writer in `src/commands/model_chains.py`
- `model_provider_routes/v1` — whose loader docstring already says "matching `load_mixture_chain_overrides`", i.e. the repo treats this as the reusable template

The shipped table stays as the fallback ballpark it already declares itself to be; the user's document wins where it has an entry. A price sourced from the user's own document should be distinguishable from the shipped ballpark in the same way `cost_approximate` already marks a derived figure.

Also in scope: add the missing `grok-code-fast` row, and give every shipped entry a date and source comment so a reader can tell how old a number is.

## Scope

One PR: schema + loader + parser + the consume site, the `grok-code-fast` row, provenance comments, and tests mirroring the sibling documents'. Note `src/commands/model_chains.py` imports `APPROX_PRICE_PER_MTOK`'s key set as a model-name vocabulary — changing keys moves that surface too.

Related: #1272, #1273 (both about what a zero means, not what a price is).

baibhavkumarsingh:
## Summary

Fixes #1274.

Adds a validated, user-configurable model price override document at `~/.omh/routing/model-prices.json` using the `model_price_overrides/v1` schema. The shipped `APPROX_PRICE_PER_MTOK` table remains the fallback, while explicit user prices take precedence for approximate cost calculations.

Also adds the missing `grok-code-fast` shipped price and provenance comments for the shipped pricing table.

## Changes

- Add `model_price_overrides/v1` parsing and loading under `~/.omh/routing/model-prices.json`
- Follow the existing strict sibling routing-document contract with duplicate-key rejection and atomic validation
- Validate model tokens and non-negative finite numeric prices, while allowing `0.0` for free tiers/subscriptions
- Resolve user overrides before shipped approximate prices without changing existing cache-read ratio behavior
- Mark override-derived approximate costs with `cost_approximate: true` and `cost_override: true`
- Add `grok-code-fast` pricing from xAI's documented list price and add provenance comments to shipped prices
- Update installation/model onboarding documentation and model optimization notes

## Validation

- `tests/test_plugin_hermes_delegation.py`: 61 tests passed
- `tests/test_model_chains_command.py`: 9 tests passed
- `tests/test_model_recommendations.py`: 26 tests passed
- `tests/test_model_routing.py`: 80 tests passed
- `tests/test_provider_entitlements.py`: 26 tests passed
- `uv run python -m compileall -q src tests`: passed
- `uv run --group lint ruff check src tests`: passed
- `git diff --check`: passed
- Documentation projection checks (`workflows`, `roles`, `capability-families`, `ulw-inventory`, `ulw-site`): passed

The full repository unittest discovery was also exercised locally; unrelated Windows-specific failures were investigated separately. The directly affected and dependent suites are green.

## Design / Compatibility

- When the override document is absent, existing shipped-price behavior is unchanged.
- Invalid override documents are rejected atomically rather than partially applied.
- `0.0` remains a valid price and is not treated as missing.
- Existing observed host cost remains distinct from token-derived approximate cost.

## Files changed

- `src/plugin_bundle/omh/hermes_delegation.py`
- `tests/test_plugin_hermes_delegation.py`
- `docs/INSTALLATION.md`
- `docs/MODEL-ONBOARDING.md`
- `MODEL_OPTI.md`

### DIFF
--- a/MODEL_OPTI.md
+++ b/MODEL_OPTI.md
@@ -573,12 +573,14 @@ gate requires a completed paired run on the intended execution surface.
 - **Cost approximation** — `APPROX_PRICE_PER_MTOK` in
   `src/plugin_bundle/omh/hermes_delegation.py` supplies `~$` estimates only
   when the host recorded no cost; models absent from the table show no
-  approximation (never a fabricated number). Cache reads are priced at a
-  tenth of input unless `APPROX_CACHE_READ_RATIO` names the model: Claude
-  Fable 5.1 lists $10 / $50 per MTok with cache reads at $0.25 (0.025x) and
-  cache writes at $12.50 (5-minute TTL) / $20 (1-hour TTL); Opus 5 reads at
-  the tenth. Mythos 5.1 carries the Fable figure because its cache-read rate
-  was open at launch — approximate, like every number in the table.
+  approximation (never a fabricated number). Overrides live in
+  `~/.omh/routing/model-prices.json` (`model_price_overrides/v1`). Cache reads
+  are priced at a tenth of input unless `APPROX_CACHE_READ_RATIO` names the
+  model: Claude Fable 5.1 lists $10 / $50 per MTok with cache reads at $0.25
+  (0.025x) and cache writes at $12.50 (5-minute TTL) / $20 (1-hour TTL); Opus 5
+  reads at the tenth. Mythos 5.1 carries the Fable figure because its
+  cache-read rate was open at launch — approximate, like every number in the
+  table.
 - **`max_tokens` is a failure signal, not a stop** — a unit whose final turn
   ended on `stop_reason: max_tokens` is a failed attempt: the output was cut
   mid-thought and nothing after the cut was verified. It is never a done
--- a/docs/INSTALLATION.md
+++ b/docs/INSTALLATION.md
@@ -389,6 +389,36 @@ configured mapping for labeling and mark the provider source as
 `model_provider_routes`; that configured metadata is not provider execution
 or credential evidence.
 
+### Customizing token prices for cost approximation
+
+When hosts record no per-call cost (such as subscription-billed plans), OMH
+derives an approximate cost from observed tokens using a shipped fallback
+ballpark table. Users and operators who use custom gateways with markups,
+enterprise committed-use discounts, promotional rates, or free tiers ($0.00)
+can override these prices without modifying source code.
+
+Supply overrides in `~/.omh/routing/model-prices.json`
+(`model_price_overrides/v1`), a sibling of the chain and provider documents:
+
+```json
+{
+  "schema_version": "model_price_overrides/v1",
+  "prices": {
+    "glm-5.2": {"input": 0.6, "output": 2.2},
+    "free-tier-model": {"input": 0.0, "output": 0.0}
+  }
+}
+```
+
+Prices are expressed in USD per million tokens (`input` and `output`). Values
+must be non-negative finite numbers (0.0 is explicitly accepted for free tiers
+and subscriptions). A model listed here supersedes the shipped ballpark table;
+unlisted models continue to use the shipped fallback. When cost is derived from
+a user override, the HUD activity row marks the derivation with both
+`cost_approximate: true` and `cost_override: true`. Validation matches the
+sibling documents: strict, atomic, and token-only, with an invalid file ignored
+whole rather than partially applied.
+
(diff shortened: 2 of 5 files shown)


sionic-khope:
Thank you for this — and an apology, because the outcome here is my fault.

You opened this on Sep 2 at 17:24Z. Roughly nine hours later I implemented the same issue myself and merged it as #1284, without first checking whether #1274 already had a PR against it. That is exactly the check I should have run before writing a line, and this PR is now conflicting because of that omission, not because of anything wrong with it.

Reviewing it properly: your version was better than mine in three specific ways, all of which I had flagged in the audit that produced #1274 and then failed to close myself.

- **Provenance on every rate.** Eight entries in `APPROX_PRICE_PER_MTOK` carried no source and no date. You gave each one its vendor page and month. Mine left them bare.
- **`grok-code-fast`.** It ships in the recommendation catalog with a provider family and no rate, so every run on it silently reported no cost at all. You added the xAI list price; I explicitly deferred it.
- **`cost_override`.** Marking a figure that came from the operator's own rate, separately from `cost_approximate`. I wrote that down as desirable in the issue and then did not build it.

All three are now in `main` via #1289, with you as a co-author on the commit. What I did not take is your schema shape (`prices` / `input` / `output`) — not because it is worse, but because the shipped one is already public and carries an optional `cache_read_ratio`, and churning a user-facing document to settle authorship would make users pay for my mistake.

I am leaving this PR open rather than closing it unilaterally, since it is yours. If you would like it closed now that its substance has landed, say so and I will; if you would rather rebase what remains onto the shipped schema, that is welcome too.

Sorry again for the duplicated effort. The process fix on my side is the obvious one: check open PRs on an issue before implementing it.

sionic-khope:
Closing this — its substance has landed in `main` via #1289, with you credited as a co-author on the commit, so leaving it open would only imply a merge that is no longer possible against the shipped schema.

To be clear about why it is closed and not merged: that is a consequence of my duplicate work, not a judgement on yours. The three things your version had and mine did not — source citations on every rate, the `grok-code-fast` row, and the `cost_override` marker — are the parts that shipped, and a test now fails any future rate added without a citation, which is your idea made permanent.

Thanks for the contribution, and sorry for the wasted round-trip. Future PRs are welcome.

PR #1291, opened 2026-09-03
fix(loop): preserve executor dispatch identity across retries

### LINKED ISSUE #1277: [Bug]: Prevent loop queue re-dispatch from overwriting recorded executor identity
## Summary

`dispatch_loop_queue_item` accepts repeated dispatch recording for the same `prepared_not_observed` queue item and replaces the existing `executor_session` object. A second call can silently erase the first session reference, thread reference, dispatch evidence, and summary while `validate_loop_cycle` still reports the cycle as valid.

This makes the persisted loop state unable to answer which executor dispatch actually owns the queue item. It is especially unsafe when a wrapper retries after losing an acknowledgement: the first dispatch may already exist, but a second recording with different metadata is accepted as the canonical state.

## Complete user goal

Run or resume one OMH loop queue item through an executor while retaining a stable, auditable dispatch identity. Exact retries should be idempotent, conflicting retries should fail safely, and recovery from an unknown delivery outcome should never destroy the first recorded dispatch evidence.

## Reproduction

Affected implementation: `src/workflows/goal_loop.py`, function `dispatch_loop_queue_item`.

1. Create a loop cycle with `create_loop_cycle`.
2. Prepare one queue item with `tick_loop_runtime`.
3. Call `dispatch_loop_queue_item` for that queue ID with `session_ref="session-first"`, `thread_ref="thread-first"`, and `evidence_refs=["dispatch:first"]`.
4. Call `dispatch_loop_queue_item` again for the same queue ID with `session_ref="session-second"`, `thread_ref="thread-second"`, and `evidence_refs=["dispatch:second"]`.
5. Read the queue item and run `validate_loop_cycle`.

The queue item remains `prepared_not_observed`, so the second call passes the current status guard. The whole `executor_session` mapping is replaced.

## Expected behavior

A queue item should have one stable logical dispatch identity unless an explicit recovery transition creates a new attempt while preserving the previous attempt.

- Replaying the exact same dispatch metadata should be a no-op or return the existing record.
- Different session, thread, executor, or evidence metadata for an already dispatched item should be rejected with an actionable conflict unless the caller uses an explicit recovery API.
- A recovery path should retain append-only attempt history and distinguish a known failed delivery from an unacknowledged or otherwise uncertain outcome.
- Observation should bind to a specific recorded dispatch attempt rather than whichever metadata was written most recently.

## Affected surfaces

- Hermes runtime or handoff
- Not executor-specific

## Observed evidence

A deterministic local run against commit `87916cf8254f902c701ad9ee084a16618c82ebc0` produced these state transitions in a temporary OMH home:

- First call: `dispatch_status="dispatched"`, `session_ref="session-first"`, `thread_ref="thread-first"`, `dispatch_evidence_refs=["dispatch:first"]`.
- Second call: `dispatch_status="dispatched"`, `session_ref="session-second"`, `thread_ref="thread-second"`, `dispatch_evidence_refs=["dispatch:second"]`.
- The first session, thread, evidence reference, and summary were no longer present after the second call.
- `validate_loop_cycle(second)` returned `{"ok": true, "errors": []}`.

The current tests exercise one dispatch followed by observation, but no test asserts replay idempotency, conflicting re-dispatch rejection, or retention of earlier dispatch attempts.

## Acceptance criteria

- [ ] An exact repeated dispatch-recording request for the same queue item is idempotent and does not duplicate or replace evidence.
- [ ] A conflicting repeated request cannot overwrite the original executor, session reference, thread reference, evidence references, or summary.
- [ ] If re-dispatch is supported, it requires an explicit recovery transition and preserves an ordered attempt history with stable attempt identifiers.
- [ ] Unknown or unacknowledged outcomes are represented separately from confirmed delivery and confirmed failure.
- [ ] Observation records identify the dispatch attempt they observe.
- [ ] `validate_loop_cycle` rejects impossible or ambiguous dispatch state.
- [ ] Regression tests cover exact replay, conflicting replay, explicit recovery, uncertain outcome, and observation-to-attempt binding.
- [ ] Existing single-dispatch loop records remain readable or receive a documented schema migration path.

## Environment

- Repository: `rlaope/oh-my-hermes`
- Checkout commit: `87916cf8254f902c701ad9ee084a16618c82ebc0`
- OMH package version: `1.0.2`
- macOS `26.5.1`
- Python `3.12.13` for the deterministic reproduction
- Hermes Agent `0.21.0` installed from git
- The affected file was unmodified in the local checkout during reproduction.

## Logs or output

```shell
first.executor_session.session_ref = "session-first"
first.executor_session.thread_ref = "thread-first"
first.executor_session.dispatch_evidence_refs = ["dispatch:first"]

second.executor_session.session_ref = "session-second"
second.executor_session.thread_ref = "thread-second"
second.executor_session.dispatch_evidence_refs = ["dispatch:second"]

validate_loop_cycle(second) = {"ok": true, "errors": []}
```

vumrra:
## Capability

Loop queue dispatch now keeps one stable, auditable identity across retries.

An exact retry is a no-op. Conflicting executor, session, thread, evidence, or summary metadata is rejected instead of replacing the first dispatch. When delivery is uncertain or known to have failed, the explicit `recover-dispatch` path appends a new attempt while retaining the original evidence.

Closes #1277.

## Motivation

`dispatch_loop_queue_item` accepted repeated writes while a queue item remained `prepared_not_observed`. Each write replaced the entire `executor_session`, so a wrapper retry after a lost acknowledgement could erase the first executor session and make the persisted record unable to identify which dispatch owned the work. `validate_loop_cycle` still accepted that state.

- Persist ordered `loop_dispatch_attempt/v1` entries with stable attempt IDs and separate `delivery_unknown`, `delivery_failed`, and `delivery_confirmed` outcomes.
- Make identical dispatch writes idempotent and reject divergent writes with an actionable recovery error.
- Add `omh loop queue recover-dispatch` as the only path for a subsequent attempt; it records the prior outcome and evidence before appending.
- Bind observations to a recorded attempt and reject duplicate, mismatched, or ambiguous attempt state during loop validation.
- Continue accepting legacy single-dispatch records without a migration.

## Rejected

- **Reject every re-dispatch** — safe for overwrites, but leaves wrappers with no recovery path when acknowledgement is lost.
- **Keep replacing `executor_session`** — preserves the old API shape but destroys the evidence needed to distinguish a retry from a second execution.

## Verification (observed)

- 8 focused dispatch, recovery, observation-binding, validation, compatibility, and CLI regressions pass.
- Full suite: **8,726 tests OK**.
- `uv run --group lint ruff check src tests`, `uv run python -m compileall -q src tests`, and `git diff --check` pass.

## Risks

- Loop queue records now carry additive attempt-history fields; legacy records remain readable.
- No live executor delivery was run from this branch. Remote CI is pending.

### DIFF
--- a/src/workflows/goal_loop.py
+++ b/src/workflows/goal_loop.py
@@ -49,11 +49,13 @@ LOOP_SMALL_LOOP_GUIDANCE_SCHEMA = "loop_small_loop_guidance/v1"
 LOOP_RUN_ONCE_RESULT_SCHEMA = "loop_run_once_result/v1"
 EXECUTOR_LOOP_CAPABILITY_SCHEMA = "executor_loop_capability/v1"
 LOOP_CYCLE_NARRATION_SCHEMA = "loop_cycle_narration/v1"
+LOOP_DISPATCH_ATTEMPT_SCHEMA = "loop_dispatch_attempt/v1"
 LOOP_INVOCATION_SCHEMA = "loop_invocation/v1"
 LOOP_CONSTRAINT_ASSESSMENT_SCHEMA = "loop_constraint_assessment/v1"
 LOOP_STICKY_RULE_SCHEMA = "loop_sticky_rule/v1"
 LOOP_STICKY_RULE_ATTACHMENT_SCHEMA = "loop_sticky_rule_attachment/v1"
 LOOP_STOP_LADDER_SCHEMA = "loop_stop_ladder/v1"
+LOOP_DISPATCH_RECOVERY_OUTCOMES = ("delivery_failed", "delivery_unknown")
 
 # The ordered stop ladder evaluated before a tick advances. The tuple order IS
 # the ladder: assess_loop_stop_ladder walks it top to bottom, the first rung
@@ -1911,21 +1913,31 @@ def dispatch_loop_queue_item(
 ) -> dict[str, Any]:
     refs = _safe_list(evidence_refs or [], limit=320)
     capability = loop_executor_capability(executor)
+    dispatch = _dispatch_request(capability, session_ref, thread_ref, refs, summary)
 
-    def mutate(cycle: dict[str, Any]) -> dict[str, Any]:
+    def mutate(cycle: dict[str, Any]) -> dict[str, Any] | None:
         runtime, item = _queue_item_ref(cycle, queue_id)
         if item.get("status") != "prepared_not_observed":
             raise ValueError("only prepared_not_observed loop queue items can record executor dispatch")
+        existing = _dict_value(item, "executor_session")
+        if str(existing.get("dispatch_status", "")) in {"prepared", "dispatched"}:
+            if _executor_session_dispatch(existing) == dispatch:
+                return None
+            if existing.get("dispatch_status") == "dispatched":
+                raise ValueError(
+                    "loop queue item already has a different executor dispatch; use explicit recovery"
+                )
+        attempts = [entry for entry in existing.get("dispatch_attempts", []) if isinstance(entry, dict)]
+        attempt = (
+            _new_dispatch_attempt(queue_id, len(attempts) + 1, dispatch)
+            if dispatch["dispatch_status"] == "dispatched"
+            else None
+        )
         item["executor_session"] = {
             **_empty_executor_session(str(capability["executor"])),
-            "executor": capability["executor"],
-            "loop_mode": capability["loop_mode"],
-            "dispatch_owner": capability["dispatch_owner"],
-            "dispatch_status": "dispatched" if refs or session_ref.strip() or thread_ref.strip() else "prepared",
-            "session_ref": _safe_summary(session_ref, limit=220) if session_ref.strip() else "",
-            "thread_ref": _safe_summary(thread_ref, limit=220) if thread_ref.strip() else "",
-            "dispatch_evidence_refs": refs,
-            "summary": _safe_summary(summary, limit=320) if summary.strip() else "Executor dispatch metadata recorded for this loop queue item.",
+            **dispatch,
+            "active_attempt_id": str(attempt["attempt_id"]) if attempt else "",
+            "dispatch_attempts": [*attempts, *([attempt] if attempt else [])],
             "capability": capability,
         }
         runtime["last_queue_id"] = str(item["queue_id"])
@@ -1948,6 +1960,97 @@ def dispatch_loop_queue_item(
     )
 
 
+def recover_loop_queue_item_dispatch(
+    paths: OmhPaths,
+    loop_id: str,
+    queue_id: str,
+    *,
+    prior_attempt_id: str,
+    prior_outcome: str,
+    outcome_evidence_refs: Iterable[str],
+    executor: str,
+    session_ref: str = "",
+    thread_ref: str = "",
+    evidence_refs: Iterable[str] | None = None,
+    outcome_summary: str = "",
+    summary: str = "",
+    expected_revision: int | None = None,
+    mutation_id: str | None = None,
+) -> dict[str, Any]:
+    attempt_id = _safe_summary(prior_attempt_id, limit=120)
+    if not attempt_id:
+        raise ValueError("prior dispatch attempt id is required")
+    if prior_outcome not in LOOP_DISPATCH_RECOVERY_OUTCOMES:
+        raise ValueError(
+            f"prior dispatch outcome must be one of: {', '.join(LOOP_DISPATCH_RECOVERY_OUTCOMES)}"
+        )
+    outcome_refs = _safe_list(outcome_evidence_refs, limit=320)
+    if not outcome_refs:
+        raise ValueError("dispatch recovery requires prior outcome evidence")
+    capability = loop_executor_capability(executor)
+    dispatch_refs = _safe_list(evidence_refs or [], limit=320)
+    dispatch = _dispatch_request(capability, session_ref, thread_ref, dispatch_refs, summary)
+    if dispatch["dispatch_status"] != "dispatched":
+        raise ValueError("dispatch recovery requires new dispatch identity or evidence")
+
+    def mutate(cycle: dict[str, Any]) -> dict[str, Any]:
+        runtime, item = _queue_item_ref(cycle, queue_id)
+        if item.get("status") != "prepared_not_observed":
+            raise ValueError("only prepared_not_observed loop queue items can recover executor dispatch")
+        existing = _dict_value(item, "executor_session")
+        attempts = [dict(entry) for entry in existing.get("dispatch_attempts", []) if isinstance(entry, dict)]
+        if not attempts:
+            raise ValueError("legacy executor dispatch has no attempt identity; record its outcome before recovery")
+        if existing.get("active_attempt_id") != attempt_id or attempts[-1].get("attempt_id") != attempt_id:
+            raise ValueError("prior dispatch attempt must be the active attempt")
+        attempts[-1] = {
+            **attempts[-1],
+            "delivery_outcome": prior_outcome,
+            "outcome_evidence_refs": outcome_refs,
+            "outcome_summary": (
+                _safe_summary(outcome_summary, limit=320)
+                if outcome_summary.strip()
+                else "Prior dispatch outcome recorded before explicit recovery."
+            ),
+        }
+        next_attempt = _new_dispatch_attempt(queue_id, len(attempts) + 1, dispatch)
+        item["executor_session"] = {
+            **existing,
+            **dispatch,
+            "active_attempt_id": next_attempt["attempt_id"],
+            "dispatch_attempts": [*attempts, next_attempt],
+            "capability": capability,
+        }
+        runtime["last_queue_id"] = str(item["queue_id"])
+        runtime["last_queue_status"] = str(item["status"])
+        cycle["runtime"] = runtime
+        cycle["next_action"] = "observe_runtime_queue"
+        cycle["updated_at"] = utc_now()
+        return cycle
+
+    return _guarded_cycle_update(
+        paths,
+        loop_id,
+        mutate,
+        operation="recover_loop_queue_item_dispatch",
+        expected_revision=expected_revision,
+        mutation_id=mutation_id,
+        mutation_digest=_mutation_digest(
+            "recover_loop_queue_item_dispatch",
+            queue_id,
+            attempt_id,
+            prior_outcome,
+            outcome_refs,
+            outcome_summary,
+            executor,
+            session_ref,
+            thread_ref,
+            dispatch_refs,
+            summary,
+        ),
+    )
+
+
 def observe_codex_loop_queue_item(
     paths: OmhPaths,
     loop_id: str,
@@ -1957,6 +2060,7 @@ def observe_codex_loop_queue_item(
     evidence_refs: Iterable[str] | None = None,
     codex_log_ref: str = "",
     summary: str = "",
+    dispatch_attempt_id: str = "",
     expected_revision: int | None = None,
     mutation_id: str | None = None,
 ) -> dict[str, Any]:
@@ -1976,6 +2080,11 @@ def observe_codex_loop_queue_item(
         if item.get("status") != "prepared_not_observed":
             raise ValueError("only prepared_not_observed loop queue items can observe Codex progress")
         existing = _dict_value(item, "executor_session")
+        observed_attempt_id = _resolve_observed_dispatch_attempt(
+            existing,
+            dispatch_attempt_id,
+            expected_executor="codex",
+        )
         executor_session = {
             **_empty_executor_session("codex"),
             **existing,
@@ -1986,9 +2095,11 @@ def observe_codex_loop_queue_item(
             "progress_summary": progress,
             "summary": _safe_summary(summary, limit=320) if summary.strip() else str(progress.get("chat_summary", "")),
             "progress_evidence_refs": all_refs,
+            "dispatch_attempts": _confirm_dispatch_attempt(existing, observed_attempt_id, all_refs),
             "capability": loop_executor_capability("codex"),
         }
         item["executor_session"] = executor_session
+        item["observed_dispatch_attempt_id"] = observed_attempt_id
         item["status"] = "observed"
         item["observed"] = True
         observed_at = utc_now()
@@ -2040,7 +2151,9 @@ def observe_codex_loop_queue_item(
         operation="observe_codex_loop_queue_item",
         expected_revision=expected_revision,
         mutation_id=mutation_id,
-        mutation_digest=_mutation_digest("observe_codex_loop_queue_item", queue_id, all_refs, summary),
+        mutation_digest=_mutation_digest(
+            "observe_codex_loop_queue_item", queue_id, all_refs, summary, dispatch_attempt_id
+        ),
     )
 
 
@@ -2087,6 +2200,7 @@ def observe_loop_queue_item(
     subagent_evidence_refs: Iterable[str] | None = None,
     connector_evidence_refs: Iterable[str] | None = None,
     summary: str = "",
+    dispatch_attempt_id: str = "",
     expected_revision: int | None = None,
     mutation_id: str | None = None,
 ) -> dict[str, Any]:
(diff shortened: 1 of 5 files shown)


src/workflows/goal_loop.py
```diff
@@ -3186,6 +3319,114 @@ def _empty_executor_session(executor: str = "choose") -> dict[str, Any]:
...
+    attempts = [dict(entry) for entry in executor_session.get("dispatch_attempts", []) if isinstance(entry, dict)]
+    if not attempt_id:
+        return attempts
+    refs = _safe_list(evidence_refs, limit=320)
+    for index, attempt in enumerate(attempts):
+        if attempt.get("attempt_id") != attempt_id:
+            continue
+        attempts[index] = {
+            **attempt,
+            "delivery_outcome": "delivery_confirmed",
```
sionic-khope, line 3419: **Important.** This overwrites `delivery_outcome` unconditionally on an id match, so an attempt that `recover-dispatch` recorded as `delivery_failed` — with the operator's evidence — flips to `delivery_confirmed` when a late observation binds to it. Worse, the concatenation on the next line keeps the failure ref in `outcome_evidence_refs` under the confirmation `outcome_summary`, so the record ends up citing `wrapper:delivery-failed:...` as evidence that delivery was confirmed. `validate_loop_cycle` accepts it (reproduced in the review body).

A conclusive outcome an operator asserted should not be silently reversible. Refusing the observation is the conservative direction, and it matches how `dispatch` already refuses a divergent retry:

```suggestion
    for index, attempt in enumerate(attempts):
        if attempt.get("attempt_id") != attempt_id:
            continue
        if attempt.get("delivery_outcome") == "delivery_failed":
            raise ValueError(
                "dispatch attempt was recorded as delivery_failed; it cannot be confirmed by a later observation"
            )
        attempts[index] = {
            **attempt,
            "delivery_outcome": "delivery_confirmed",
```

If you would rather keep the observation and preserve the history, the other honest shape is to leave `outcome_evidence_refs` alone and record the confirmation refs under a separate key, so the two claims stay attributable. Either way the current merge should not ship.

src/workflows/goal_loop.py
```diff
@@ -1948,6 +1960,97 @@ def mutate(cycle: dict[str, Any]) -> dict[str, Any]:
...
+        raise ValueError("dispatch recovery requires new dispatch identity or evidence")
+
+    def mutate(cycle: dict[str, Any]) -> dict[str, Any]:
+        runtime, item = _queue_item_ref(cycle, queue_id)
+        if item.get("status") != "prepared_not_observed":
+            raise ValueError("only prepared_not_observed loop queue items can recover executor dispatch")
+        existing = _dict_value(item, "executor_session")
+        attempts = [dict(entry) for entry in existing.get("dispatch_attempts", []) if isinstance(entry, dict)]
+        if not attempts:
+            raise ValueError("legacy executor dispatch has no attempt identity; record its outcome before recovery")
```
sionic-khope, line 2003: **Important.** This makes any loop cycle written before this PR permanently un-re-dispatchable. `dispatch_loop_queue_item` sends a divergent retry here with *"use explicit recovery"* (goal_loop.py:1926-1928), and this line then refuses because the legacy record carries no attempt identity — telling the caller to *"record its outcome"*, which no command or API in the repo does. The item can only be observed or blocked from that point; before this change it could be re-dispatched. Reproduced in the review body against a record downgraded to the legacy shape.

That is precisely the population #1277 is about: in-flight items already on disk. The PR body's "legacy records remain readable" is true, but readable is not recoverable.

The narrow fix is to let recovery adopt a legacy session as attempt 1 rather than refusing it — the legacy `executor_session` already holds everything `_new_dispatch_attempt` needs, and `_executor_session_dispatch(existing)` is right there to build it from. `prior_attempt_id` would then be optional in exactly this case. If you would rather not widen recovery, the error at least needs to name a real next step instead of one that does not exist.

src/workflows/goal_loop.py
```diff
@@ -3186,6 +3319,114 @@ def _empty_executor_session(executor: str = "choose") -> dict[str, Any]:
...
+) -> str:
+    attempts = [entry for entry in executor_session.get("dispatch_attempts", []) if isinstance(entry, dict)]
+    requested = _safe_summary(requested_attempt_id, limit=120) if requested_attempt_id.strip() else ""
+    if not attempts:
+        if requested:
+            raise ValueError("dispatch attempt id does not exist on this loop queue item")
+        return ""
+    if not requested:
+        if len(attempts) != 1:
+            raise ValueError("dispatch attempt id is required after explicit recovery")
```
sionic-khope, line 3395: **Important.** This is the right guard, but nothing on the operator-facing side hands the wrapper the id it now has to supply. After a recovery the item is still `prepared_not_observed`, so `build_loop_queue_handoff` renders and its `actions` list still reads `['observe_loop_queue', 'block_loop_queue', 'show_loop_status']` (goal_loop.py:1542) — and neither the handoff payload nor `_queue_item_summary` (goal_loop.py:4121) carries `active_attempt_id` or `dispatch_attempts`. A wrapper that follows the handoff, which is the documented contract for what to do next with a queue item, hits this ValueError. Reproduced in the review body.

Adding `active_attempt_id` to the handoff payload (and `recover_loop_queue_dispatch` to `actions` once attempts exist) closes it, and gives `_queue_item_summary` somewhere to surface the attempt count too.

src/workflows/goal_loop.py
```diff
@@ -1911,21 +1913,31 @@ def dispatch_loop_queue_item(
...
             raise ValueError("only prepared_not_observed loop queue items can record executor dispatch")
+        existing = _dict_value(item, "executor_session")
+        if str(existing.get("dispatch_status", "")) in {"prepared", "dispatched"}:
+            if _executor_session_dispatch(existing) == dispatch:
+                return None
+            if existing.get("dispatch_status") == "dispatched":
+                raise ValueError(
+                    "loop queue item already has a different executor dispatch; use explicit recovery"
+                )
+        attempts = [entry for entry in existing.get("dispatch_attempts", []) if isinstance(entry, dict)]
```
sionic-khope, line 1930: Nit. `attempts` is provably always empty here. Reaching this line means `existing`'s `dispatch_status` is either not in `{prepared, dispatched}`, or is `prepared` with divergent metadata — and an entry is only ever appended to `dispatch_attempts` when the status becomes `dispatched` (here and in `recover_loop_queue_item_dispatch`), while `progress_observed` implies the item status is `observed`, which the precondition above already rejects. So the `[*attempts, ...]` spread on line 1940 always starts from `[]`.

Harmless, but it reads as if a prepared session could carry attempt history, which is the one thing the invariant says it cannot. Either drop the carry-forward or leave a line saying it is defensive against a hand-edited record.

sionic-khope (commented):
**3 important, 2 nits**

Round one — no review exists at any SHA on this PR, so the `REVIEW.md` re-review ratchet does not apply. Reviewed at head `6ddb72a56ad7722034a5499e5e2fc3932163e73a` in a clean worktree, not from the diff text.

**Shape of the work.** `executor_session` grows an append-only `loop_dispatch_attempt/v1` history; `dispatch_loop_queue_item` becomes replay-idempotent and refuses divergent metadata; a new `omh loop queue recover-dispatch` is the only way to append an attempt; observations bind to an attempt id; `_validate_executor_dispatch_session` pins the invariants and skips records that predate the field. The design is right and the test coverage of the happy paths is genuinely good. The three findings below are all about states the new code can reach and then cannot leave, or can leave with a record that misstates its own evidence. Each one is reproduced, script and output included.

### Gates, run here at this head

```
Ran 8726 tests in 458.773s

OK
```

```
$ uv run --group lint ruff check src tests
All checks passed!
$ uv run python -m compileall -q src tests
$ uv run python -m omh.cli docs workflows --check
{"checked":".../docs/WORKFLOWS.md","ok":true,"tap_skills":{"checked":175,"expected":175,"extra":[],"missing":[],"ok":true,"stale":[]}}
$ uv run python -m omh.cli docs roles --check
{"checked":".../docs/ROLES.md","ok":true}
$ uv run python -m omh.cli docs capability-families --check
{"checked":".../src/plugin_bundle/omh/tools/capability_families.json","ok":true}
$ uv run python -m omh.cli docs ulw-inventory --check
{"checked":".../README.md","ok":true}
$ uv run python -m omh.cli docs ulw-site --check
{"checked":".../site/index.html","ok":true}
$ git diff --check
```

All green. The PR body's `8,726 tests OK` is accurate — that is the number I got. Commit trailers are complete and `Signed-off-by:` is last and matches the author. No generated artifact was hand-edited. No network, LLM, or new dependency. No routing change, so the positive/negative-case rule does not bite here.

### Important

**1. Observing an attempt already recorded `delivery_failed` rewrites it to `delivery_confirmed` and keeps the failure evidence under the confirmation summary.** Inline at `src/workflows/goal_loop.py`. `_confirm_dispatch_attempt` matches on `attempt_id` alone and overwrites `delivery_outcome` unconditionally, then concatenates the observation refs onto the existing `outcome_evidence_refs`. So an attempt the operator asserted had failed, with evidence, ends up asserting the opposite while still citing the failure ref. `validate_loop_cycle` accepts the result. For a PR whose whole subject is keeping dispatch evidence auditable, this is the one state that must not be silently reachable.

**2. A loop cycle written before this change can no longer be re-dispatched at all, and the error names an action that does not exist.** Inline at `src/workflows/goal_loop.py`. `dispatch_loop_queue_item` refuses a divergent retry with *"use explicit recovery"*, and `recover_loop_queue_item_dispatch` then refuses because the legacy record has no attempt identity, telling the caller to *"record its outcome before recovery"* — there is no command or API that does that. Legacy records are readable, as the body says, but the recovery capability the PR adds is unavailable to exactly the in-flight records that exist today, which are the ones #1277 is about. Before this change such an item could be re-dispatched; now it can only be observed or blocked.

**3. After a recovery, `build_loop_queue_handoff` still advertises `observe_loop_queue` without the attempt id that `observe` now requires.** Inline at `src/workflows/goal_loop.py`. The item stays `prepared_not_observed` after recovery, so the handoff renders, and its `actions` list is the contract a wrapper follows. Neither the handoff nor `_queue_item_summary` (`src/workflows/goal_loop.py:4121`) carries `active_attempt_id` or `dispatch_attempts`, so the only place the id is obtainable is the raw `dispatch` / `recover-dispatch` JSON. Following the handoff literally is a hard error.

### Reproductions

All three, on this head, in one script:

```python
import sys
from pathlib import Path
from tempfile import TemporaryDirectory
sys.path.insert(0, "src")
from omh.paths import resolve_paths
from omh.goal_loop import (
    create_loop_cycle, tick_loop_runtime, dispatch_loop_queue_item,
    recover_loop_queue_item_dispatch, observe_codex_loop_queue_item,
    observe_loop_queue_item, build_loop_queue_handoff, validate_loop_cycle,
)

def prepared(tmp, summary):
    paths = resolve_paths(Path(tmp) / ".omh", Path(tmp) / ".hermes")
    cycle = create_loop_cycle(
        paths, goal_summary=summary, goal_reframe=summary,
        success_criteria=[summary], permission_profile="execute_with_gates",
        allowed_executors=["codex"],
    )
    queue_id = tick_loop_runtime(paths, cycle["loop_id"])["runtime"]["queue"][0]["queue_id"]
    first = dispatch_loop_queue_item(
        paths, cycle["loop_id"], queue_id, executor="codex",
        session_ref="codex-session-1", evidence_refs=["wrapper:dispatch:1"],
    )
    return paths, cycle["loop_id"], queue_id, first["runtime"]["queue"][0]["executor_session"]["active_attempt_id"]

# 1 - a recorded delivery failure is rewritten as a confirmation
with TemporaryDirectory() as tmp:
    paths, loop_id, queue_id, a1 = prepared(tmp, "failed delivery then observation")
    recover_loop_queue_item_dispatch(
        paths, loop_id, queue_id, prior_attempt_id=a1, prior_outcome="delivery_failed",
        outcome_evidence_refs=["wrapper:delivery-failed:executor-rejected-the-dispatch"],
        outcome_summary="The executor rejected the dispatch.",
        executor="codex", session_ref="codex-session-2", evidence_refs=["wrapper:dispatch:2"],
    )
    observed = observe_codex_loop_queue_item(
        paths, loop_id, queue_id,
        codex_log_text='{"role":"assistant","content":"Work completed."}',
        evidence_refs=["codex-jsonl:1"], dispatch_attempt_id=a1,
    )
    print(observed["runtime"]["queue"][0]["executor_session"]["dispatch_attempts"][0])
    print(validate_loop_cycle(observed))

# 3 - the handoff names an action the wrapper cannot run
with TemporaryDirectory() as tmp:
    paths, loop_id, queue_id, a1 = prepared(tmp, "handoff after recovery")
    recover_loop_queue_item_dispatch(
        paths, loop_id, queue_id, prior_attempt_id=a1, prior_outcome="delivery_unknown",
        outcome_evidence_refs=["wrapper:ack-timeout:1"],
        executor="codex", session_ref="codex-session-2", evidence_refs=["wrapper:dispatch:2"],
    )
    print(build_loop_queue_handoff(paths, loop_id, queue_id)["actions"])
    observe_loop_queue_item(paths, loop_id, queue_id, evidence_refs=["wrapper:observed:1"])
```

Finding 1 output:

```
{... 'delivery_outcome': 'delivery_confirmed',
 'outcome_evidence_refs': ['codex-jsonl:1',
                           'wrapper:delivery-failed:executor-rejected-the-dispatch'],
 'outcome_summary': 'Executor progress or result was observed for this dispatch attempt.'}
{'ok': True, 'errors': []}
```

Finding 3 output:

```
['observe_loop_queue', 'block_loop_queue', 'show_loop_status']
ValueError: dispatch attempt id is required after explicit recovery
```

Finding 2 needs the legacy shape, which I produced by popping the two new keys off the on-disk record — the same downgrade `test_loop_cycle_validation_keeps_legacy_single_dispatch_records_readable` performs, but written back to `loop_cycle_path(...)` before the retry:

```
legacy record validates: {'ok': True, 'errors': []}
re-dispatch  -> ValueError: loop queue item already has a different executor dispatch; use explicit recovery
recover      -> ValueError: legacy executor dispatch has no attempt identity; record its outcome before recovery
```

### Nits

1. `src/workflows/goal_loop.py:1930` — the `attempts` carry-forward in `dispatch_loop_queue_item` is unreachable. Inline.
2. `dispatch_loop_queue_item` rebuilds `executor_session` from `_empty_executor_session(...)` while `recover_loop_queue_item_dispatch` spreads `**existing`. The asymmetry is probably deliberate (recovery keeps progress fields, first dispatch resets them), but nothing says so, and the two functions read as if they should be the same shape. One sentence of comment on the recovery side would settle it.

sionic-khope:
Follow-up on the review above: rather than leave the three findings as a round trip, I carried your commit unchanged onto `claude/loop-dispatch-identity-followup` and fixed them on top — **#1299**.

Your commit is preserved verbatim, with your authorship and sign-off intact; my changes are a second commit on top of it. The design is yours and it is the right one — attempt identity, the idempotent replay, the explicit recovery path, the validator. What #1299 adds is only the three states the review reproduced:

- **A recorded `delivery_failed` is now terminal.** `_resolve_observed_dispatch_attempt` refuses to bind an observation to a failed attempt instead of flipping it to `delivery_confirmed`, and confirming an *uncertain* attempt moves the superseded claim — outcome, refs, summary — into a new `outcome_history` entry rather than concatenating the two ref lists. An acknowledgement timeout no longer ends up cited as evidence that the dispatch arrived.
- **A legacy record recovers instead of wedging.** `recover_loop_queue_item_dispatch` adopts the legacy `executor_session` as attempt 1, so `--prior-attempt-id` is optional in exactly that case. The adopted attempt carries an empty `recorded_at`, since the legacy shape never stored a dispatch timestamp and inventing one would be a claim the record cannot support.
- **The handoff hands over what `observe` now needs.** `build_loop_queue_handoff` and `_queue_item_summary` carry `active_dispatch_attempt_id` and `dispatch_attempt_count`, and the handoff lists `recover_loop_queue_dispatch` once attempts exist.

Also folded in: the nit about the unreachable attempt carry-forward in `dispatch_loop_queue_item`, four regressions in `tests/test_loop_cycle.py` pinning each fix, and the `docs/ARCHITECTURE.md` paragraph extended to state the outcome ratchet, the legacy adoption, and what the handoff carries.

Full suite green at 8,731 on the rebased tree, plus ruff, compileall, and the byte gates — the numbers and the before/after reproductions are in #1299's body.

Closing this one in favour of #1299 so the two do not diverge. The work and the credit are yours; thank you for the PR.

Issue #1056, opened 2026-08-21
Run the paid-live benchmark for the four new family calibrations (mistral, llama, codestral, solar)


sionic-khope:
**Contributions welcome — this is a well-scoped task with an existing harness.**

## Context

#1055 added subagent+composer calibration pairs for `mistral`, `llama`, `codestral`, and `solar` in `src/coding/unit_prompt_protocol.py` (`HIGH_EFFORT_CALIBRATIONS` / `MAIN_AGENT_COMPOSITION_CALIBRATIONS`). They were authored from provider-published characteristics, and each of their MODEL_OPTI.md sections is explicitly marked **live benchmark validation pending** — the only calibrations in the table without a measured effect. The eight older families went through this same pipeline (#948).

## Current state (exact)

- Harness: `benchmarks/live-model-tools/v1` (`omh_live_model_tool_benchmark/v1`). `fake` is the default offline adapter; the `omh` path requires `--allow-paid-live` and dispatches through the explicit `omh coding hermes-child dispatch --confirm-dispatch` boundary; `hermes_current_session` uses `hermes --oneshot` with the caller's profile (labeled as such in every artifact).
- The pairing contract is pinned: `tests/test_omh_live_model_benchmark.py::test_prompt_pair_changes_only_omh_calibration` asserts the calibration block is the ONLY difference between baseline and optimized prompts.
- Models resolve via `omh coding model-route --executor hermes --model <id> --effort high --role implementation --json` (see `lib/omh_live.py::route_argv`).

## Plan

1. `python benchmarks/live-model-tools/v1/bench.py doctor` then `corpus --verify` then an offline `smoke` to confirm the harness runs on your machine.
2. Pick one concrete model id per family you can actually serve (e.g. a current Mistral Large id, a hosted Llama id, a Codestral id, `solar-pro2`). The family label comes from `model_family()`, so any id with the right prefix selects the right calibration.
3. Run the paid-live comparison per family with `--allow-paid-live` (or `hermes_current_session` if the model is already authenticated in your Hermes profile). Raw prompts/credentials are never persisted by design.
4. Record the signed `routing_observation/v1` evidence per the harness contract.

## Acceptance

- One baseline-vs-calibrated run per family with signed observations.
- MODEL_OPTI.md sections updated: drop "pending", state the observed result honestly — a calibration that measures worse gets revised or removed, not kept.
- `tests/test_omh_live_model_benchmark.py` stays green (no pairing contamination).
[...]

PR #1305, opened 2026-09-03
feat(benchmarks): manifest-aware analysis for targeted family runs

GentechLabs:
### What Changed

- `benchmarks/live-model-tools/v1/analyze.py` now accepts `--manifest <path>`, defaulting to the canonical `manifest.json`, and forwards the selected manifest object to `statistics.analyze(...)`.

### Why This Exists

- Issue #1056 defines a targeted-manifest path for family-calibration contributors (`bench.py run --manifest <targeted.json>`), but the analysis CLI reloaded the canonical manifest directly, so the analysis coverage check (`live manifest matrix`) could never pass for a targeted family run. The harness could run but not produce an auditable analysis report.

### User / Operator Impact

- Operators validating one family (the guide's path for mistral/llama/codestral/solar contributors) can now analyze their targeted run artifacts with the same manifest they ran, end to end, without touching the canonical manifest or its coverage contract.

### How It Works

- The CLI previously hard-coded `load_object(BASE / "manifest.json")`; it now resolves `--manifest` (default unchanged) and passes the parsed object to `statistics.analyze`, which was already manifest-relative. Coverage, digest-pairing, split, and instance checks are unchanged; a targeted manifest simply defines its own live matrix.

- `benchmarks/live-model-tools/v1/analyze.py` (CLI `--manifest` flag)
- `benchmarks/live-model-tools/v1/tests/test_framework.py` (new `OmhTargetedManifestAnalysisTests`)

### Affected Surfaces

- [x] Not executor-specific

## Validation

- [x] `PYTHONPATH=tests uv run python -m unittest discover -s tests -v` (8,867 tests, OK, 22 skips; benchmark suite 15/15)
- [x] `uv run python -m omh.cli docs workflows --check` (177/177 skills, no stale)

Sabotage check: the new end-to-end test fails against the pre-change CLI (`unrecognized arguments: --manifest`) and passes with the fix.

**Not Tested / Not Applicable:** paid live model runs were intentionally not executed for this PR — it is harness-only (issue thread scope from the Aug 30 guide, step 0). Family calibration runs with observed evidence land in follow-up PRs per the one-family-per-PR guidance.

Refs #1056

### DIFF
--- a/benchmarks/live-model-tools/v1/analyze.py
+++ b/benchmarks/live-model-tools/v1/analyze.py
@@ -8,7 +8,7 @@ from common import load_object, write_json
 from statistics import analyze
 
 def main(argv=None):
- p=argparse.ArgumentParser(); p.add_argument("--baseline",type=Path,required=True); p.add_argument("--optimized",type=Path,required=True); p.add_argument("--bootstrap-repetitions",type=int,default=10000); p.add_argument("--seed",type=int,default=20260813); p.add_argument("--output",type=Path,required=True); a=p.parse_args(argv)
+ p=argparse.ArgumentParser(); p.add_argument("--baseline",type=Path,required=True); p.add_argument("--optimized",type=Path,required=True); p.add_argument("--bootstrap-repetitions",type=int,default=10000); p.add_argument("--seed",type=int,default=20260813); p.add_argument("--manifest",type=Path,default=BASE/"manifest.json"); p.add_argument("--output",type=Path,required=True); a=p.parse_args(argv)
  if a.bootstrap_repetitions<100: p.error("bootstrap repetitions must be at least 100")
- report=analyze(a.baseline,a.optimized,a.bootstrap_repetitions,a.seed,load_object(BASE/"manifest.json")); write_json(a.output,report); print(json.dumps(report,sort_keys=True,indent=2)); return 0
+ report=analyze(a.baseline,a.optimized,a.bootstrap_repetitions,a.seed,load_object(a.manifest)); write_json(a.output,report); print(json.dumps(report,sort_keys=True,indent=2)); return 0
 if __name__=="__main__": raise SystemExit(main())
--- a/benchmarks/live-model-tools/v1/tests/test_framework.py
+++ b/benchmarks/live-model-tools/v1/tests/test_framework.py
@@ -245,5 +245,150 @@ class OmhBenchmarkFrameworkTests(unittest.TestCase):
                 analyze(baseline, optimized, 10, 7, manifest)
 
 
+class OmhTargetedManifestAnalysisTests(unittest.TestCase):
+    """Issue #1056 step-0: targeted family manifests must flow end-to-end.
+
+    bench.py run accepts --manifest, but the analyze.py CLI used to reload the
+    canonical manifest directly, so a targeted manifest (for example, one live
+    solar family entry) could run but never pass the analysis coverage check.
+    These tests pin the CLI contract: --manifest selects the manifest used for
+    analysis coverage, defaults stay canonical, and mismatched manifests still
+    fail loudly.
+    """
+
+    def _run_cli(self, args):
+        return subprocess.run(
+            [sys.executable, str(BASE / "analyze.py"), *args],
+            capture_output=True,
+            text=True,
+        )
+
+    def _live_manifest(self):
+        manifest = json.loads((BASE / "manifest.json").read_text())
+        manifest["models"] = [m for m in manifest["models"] if m.get("live")][:1]
+        return manifest
+
+    def _evaluation_rows(self, manifest, condition):
+        import corpus
+
+        model = next(m for m in manifest["models"] if m.get("live"))
+        rows = []
+        for template, _task_class in corpus.TEMPLATES:
+            for seed in corpus.EVALUATION_SEEDS:
+                rows.append(
+                    {
+                        "schema_version": "omh_live_model_tool_run/v1",
+                        "created_at": "2026-08-13T00:00:00+00:00",
+                        "instance_id": f"E-{template}-{seed}",
+                        "split": "evaluation",
+                        "harness": "omh",
+                        "model": {"provider": model["provider"], "id": model["id"]},
+                        "task_digest": "a" * 64,
+                        "route": {},
+                        "observation": {
+                            "status": "completed",
+                            "tools": 1,
+                            "tokens": 1,
+                            "cost_usd": 0.0,
+                        },
+                        "grade": {"pass": True},
+                        "final_tree_digest": "b" * 64,
+                        "condition": condition,
+                    }
+                )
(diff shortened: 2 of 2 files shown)


benchmarks/live-model-tools/v1/tests/test_framework.py
```diff
@@ -245,5 +245,150 @@ def test_analysis_rejects_optimized_development_split(self) -> None:
...
+            self.assertEqual(
+                report["models"][label]["n"],
+                len(self._evaluation_rows(manifest, "baseline")),
+            )
+
+    def test_targeted_manifest_cli_rejects_mismatched_manifest(self) -> None:
+        manifest = self._live_manifest()
+        excluded = manifest["models"][0]
+        other = self._live_manifest()
+        other["models"] = [
```
sionic-khope, line 339: `other` starts from `_live_manifest()`, which is already `[first live model]`, and this comprehension removes exactly that model — `other["models"]` ends up empty. So the assertion below rejects an empty matrix rather than a mismatched one. Filter the full `manifest.json` down to a *different* live model instead.

benchmarks/live-model-tools/v1/tests/test_framework.py
```diff
@@ -245,5 +245,150 @@ def test_analysis_rejects_optimized_development_split(self) -> None:
...
+            result = self._run_cli(
+                [
+                    "--baseline", str(baseline),
+                    "--optimized", str(baseline),
+                    "--manifest", str(other_path),
+                    "--bootstrap-repetitions", "100",
+                    "--output", str(root_path / "report.json"),
+                ]
+            )
+            self.assertNotEqual(result.returncode, 0)
```
sionic-khope, line 365: Any non-zero exit satisfies this, including argparse rejecting `--manifest` on the pre-change CLI. Assert on the coverage-failure text in `result.stderr` so the test pins the analysis contract, not just "it failed".

benchmarks/live-model-tools/v1/tests/test_framework.py
```diff
@@ -245,5 +245,150 @@ def test_analysis_rejects_optimized_development_split(self) -> None:
...
+                ]
+            )
+            # One targeted live model against the canonical manifest violates
+            # the canonical matrix check: the old failure mode, still intact.
+            self.assertNotEqual(result.returncode, 0)
+
+
 if __name__ == "__main__":
     unittest.main()
+
```
sionic-khope, line 394: Blank line at EOF — `git diff --check origin/main...HEAD` flags this (`new blank line at EOF`) and the CI whitespace step will fail on it.

sionic-khope (commented):
Round 1, against b32356c8.

**2 important, 3 nits.**

Shape: a small, correct CLI change (`--manifest` on `analyze.py`, default stays canonical, forwarded to the already manifest-relative `statistics.analyze`) plus three subprocess-level tests. The core change is right, and the end-to-end test genuinely fails on the pre-change CLI (`unrecognized arguments: --manifest`). Observed locally on this head: benchmark suite `python tests/test_framework.py` 15/15 OK.

### Important

1. **CI whitespace gate fails on this head.** `git diff --check origin/main...HEAD` reports `benchmarks/live-model-tools/v1/tests/test_framework.py:394: new blank line at EOF.` That is the `Whitespace and conflict-marker check` step in `.github/workflows/ci.yml`. `analyze.py` also lost its trailing newline (`\ No newline at end of file` in the diff). CI has not run on this PR yet (first-contribution approval), so the checks tab does not show it; it will go red once it runs. Fix: drop the extra blank line at EOF of `test_framework.py` and restore the final newline in `analyze.py`.
2. **DCO red, no trailers.** The commit carries no `Signed-off-by:` and none of the Lore-style trailers `AGENTS.md` requires (Constraint / Rejected / Confidence / Scope-risk / Directive / Tested / Not-tested, with `Signed-off-by:` last). `git commit --amend -s` plus the trailer block and a force-push clears both.

### Nits

- `test_targeted_manifest_cli_rejects_mismatched_manifest` builds `other` from `_live_manifest()`, which already truncates to one live model, then removes that same model — so `other["models"]` is `[]` (checked: count 0). The test proves "rows outside an *empty* matrix are rejected", not "a *different* targeted manifest rejects foreign rows". Build `other` from the full `manifest.json` filtered to a different live model so the matrix is populated.
- The two rejection tests assert only `returncode != 0`, so they also pass against the pre-change CLI (argparse's `unrecognized arguments` exit satisfies them). Assert on the coverage error text in `stderr` so an argument error cannot stand in for the contract.
- Context, not on you: `benchmarks/live-model-tools/v1/tests` is not run by CI (`ci.yml` runs `unittest discover -s tests` only), so the contract these tests pin is enforced only when someone runs them by hand. Worth a follow-up issue.

PR body follows the template and its claims match the diff. Thanks — once the two Important items are fixed this is straightforward.

benchmarks/live-model-tools/v1/analyze.py
```diff
@@ -8,7 +8,7 @@
...
+ p=argparse.ArgumentParser(); p.add_argument("--baseline",type=Path,required=True); p.add_argument("--optimized",type=Path,required=True); p.add_argument("--bootstrap-repetitions",type=int,default=10000); p.add_argument("--seed",type=int,default=20260813); p.add_argument("--manifest",type=Path,default=BASE/"manifest.json"); p.add_argument("--output",type=Path,required=True); a=p.parse_args(argv)
  if a.bootstrap_repetitions<100: p.error("bootstrap repetitions must be at least 100")
- report=analyze(a.baseline,a.optimized,a.bootstrap_repetitions,a.seed,load_object(BASE/"manifest.json")); write_json(a.output,report); print(json.dumps(report,sort_keys=True,indent=2)); return 0
-if __name__=="__main__": raise SystemExit(main())
+ report=analyze(a.baseline,a.optimized,a.bootstrap_repetitions,a.seed,load_object(a.manifest)); write_json(a.output,report); print(json.dumps(report,sort_keys=True,indent=2)); return 0
+if __name__=="__main__": raise SystemExit(main())
```
sionic-khope, line 14: The final newline was dropped here (`\ No newline at end of file`). Please restore it.

sionic-khope:
cc @GentechLabs — thanks for this; the CLI change is exactly right. To land it without another round-trip, the maintainers carried it into #1312 with `Co-authored-by` credit and applied the review fixes there: the mismatched-manifest test now analyses against a populated matrix of a different live model, both rejection tests write a real optimized file and assert the coverage error text (while doing that it turned out they were failing on `optimized file contains wrong condition` — the baseline file was passed twice — before reaching the manifest check), and the EOF whitespace is fixed. Once #1312 is merged this PR will be closed as superseded; nothing further is needed from you.

sionic-khope:
Superseded by #1312, now merged with your change credited via Co-authored-by. Thanks again, @GentechLabs.

PR #1413, opened 2026-09-08
fix: keep delegation route edits in the active Hermes home

Joeywrz:
### What Changed

Default delegation route reads and writes now resolve the active Hermes home, rather than always targeting `~/.hermes`. An explicitly supplied home keeps precedence.

### Why This Exists

In a profile session, setting, reading, or clearing a route without an explicit home ignored `HERMES_HOME` and could address the default profile configuration. The shared `_config_path` was the common cause for route and parent-provider reads.

### User / Operator Impact

The route tool targets the active profile. Other profile configurations, unrelated YAML settings, and comments remain unchanged. Callers can still explicitly address another home.

### How It Works

Resolve an explicit home first. Otherwise use Hermes `get_hermes_home` when importable. The standalone bundle falls back to nonempty `HERMES_HOME`, then `~/.hermes`, with tilde expansion. Reuse this shared resolver for existing callers; no route lifecycle or dispatch protocol is added.

### Files And Contracts Touched

`src/plugin_bundle/omh/delegation_routing.py`: shared configuration path. `src/plugin_bundle/omh/tools/delegate_route_tool.py`: accurate schema description. `tests/test_delegate_route_tool.py`: seven profile-resolution roundtrip regressions through the real tool handler.

### Affected Surfaces

- [x] Hermes chat or wrapper
- [ ] Codex
- [ ] Claude Code
- [x] Hermes runtime or handoff
- [ ] Generic executor
- [ ] Not executor-specific

## Validation

Check only commands that completed successfully. Leave non-applicable checks
unchecked and explain why under **Not Tested / Not Applicable**.

- [x] `uv run --group lint ruff check src tests`
- [x] `PYTHONPATH=tests uv run python -m unittest discover -s tests -v`
- [x] `uv run python -m compileall -q src tests`
- [ ] `uv run python -m omh.cli docs workflows --check`
- [ ] `uv run python -m omh.cli docs roles --check`
- [ ] `uv run python -m omh.cli docs capability-families --check`
- [ ] `uv run python -m omh.cli harness validate`
- [ ] `uv run python -m omh.cli release checklist --json`
- [ ] `uv run python -m omh.cli release hermes-smoke`
- [x] `git diff --check`
- [ ] `omh --help`
- [ ] `omh --omh-home /tmp/omh-smoke --hermes-home /tmp/hermes-smoke release hermes-smoke --install-path setup --omh-command omh --include-command-smoke`
- [ ] Relevant workflow-learning review queue smoke, when learning contracts changed:
[...]

PR #1415, opened 2026-09-08
fix(memory): bound complete record prefetch serialization

Joeywrz:
### What Changed

- Bound the complete `<memory_records>` prefetch section to its character budget, including wrapper tags, newlines, XML escaping, and omission reports.
- Replace unbounded per-record omission lines with at most two aggregate reports carrying exact counts. Return only the number of complete record elements actually rendered.

### Why This Exists

- The old renderer budgeted record bodies but not wrappers or omission metadata. A synthetic store of 10,000 records produced a 649,038-character section under the default 2,400-character budget. A single-record boundary test also showed that wrapper overhead could exceed the requested budget.
- Both regressions fail against the unchanged base implementation and pass against this patch.

### User / Operator Impact

- Large reviewed stores no longer expand the record prefetch section beyond its advertised budget. Recall status counts admitted records actually included, not omitted records.
- Review eligibility, ranking, executor selection, and existing summary/date projections remain unchanged. This is a language-independent serialization fix.

### How It Works

- Start accounting with the opening/closing tags and separators. For multiple records, reserve space for both possible omission reports using the input count as their maximum counter width.
- Select complete escaped records in existing order and aggregate cuts by `render_budget_exhausted` or `record_limit_reached`.
- Check final serialized length defensively. If the custom budget cannot hold the report, return an empty section and zero rendered records.

### Files And Contracts Touched

- `src/plugin_bundle/omh/memory_records.py`: bounded serialization and aggregate omission accounting.
- `tests/test_memory_records.py`: large stores, tiny/negative budgets and limits, escaping and long metadata, exact single-record boundaries, ordering/provenance, mixed omission reasons, and isolated standalone-bundle import coverage.
- `tests/test_memory_provider.py`: admitted-record integration coverage for prefetch length and recall status; update the omission format expectation.
- `docs/MEMORY.md`: complete-section budget, aggregate format, conservative reservation, and tiny-budget behavior.

### Affected Surfaces

- [x] Hermes chat or wrapper
- [ ] Codex
- [ ] Claude Code
- [x] Hermes runtime or handoff
- [ ] Generic executor
- [x] Not executor-specific

## Validation

Check only commands that completed successfully. Leave non-applicable checks
unchecked and explain why under **Not Tested / Not Applicable**.

- [x] `uv run --group lint ruff check src tests`
[...]

PR #1414, opened 2026-09-08
fix: keep consolidation status queries read-only

Joeywrz:
### What Changed

The `omh_memory` consolidation action now reads the latest recorded reminder and scheduler counters without initializing a provider or evaluating triggers. It explicitly returns `evaluated: false`; `due` is absent if no readable brief exists.

### Why This Exists

The status handler initialized a provider session, consuming pending scheduler state, then evaluated a second time. A status request could therefore return `due: false` immediately after writing a due brief. This closes the plugin-side counterpart of the CLI status fix in #912.

### User / Operator Impact

Repeated status questions preserve pending counters, suppression state, and reminder bytes. Operators see the recorded brief rather than accidentally triggering or consuming another evaluation. A missing brief means no recorded result, not proof that no reminder is due.

### How It Works

Reuse the existing validated `read_latest_consolidation` and `read_dreaming_state` readers. Remove provider lifecycle calls from this read action. Explicit CLI `dream --evaluate`, provider lifecycle scheduling, reminder delivery from #1410, and normal tool-observation telemetry remain unchanged.

### Files And Contracts Touched

`src/plugin_bundle/omh/tools/memory_tool.py`: status behavior and schema description. `tests/test_memory_provider.py`: real handler regressions for pending state, repeated due-brief reads, and no provider lifecycle. `docs/MEMORY.md`: read-versus-evaluate contract.

### Affected Surfaces

- [x] Hermes chat or wrapper
- [ ] Codex
- [ ] Claude Code
- [x] Hermes runtime or handoff
- [ ] Generic executor
- [ ] Not executor-specific

## Validation

Check only commands that completed successfully. Leave non-applicable checks
unchecked and explain why under **Not Tested / Not Applicable**.

- [x] `uv run --group lint ruff check src tests`
- [x] `PYTHONPATH=tests uv run python -m unittest discover -s tests -v` — equivalent direct repository `.venv/bin/python` invocation, isolated HOME/HERMES_HOME/OMH_HOME.
- [x] `uv run python -m compileall -q src tests`
- [ ] `uv run python -m omh.cli docs workflows --check`
- [ ] `uv run python -m omh.cli docs roles --check`
- [ ] `uv run python -m omh.cli docs capability-families --check`
- [ ] `uv run python -m omh.cli harness validate`
- [ ] `uv run python -m omh.cli release checklist --json`
- [ ] `uv run python -m omh.cli release hermes-smoke`
- [x] `git diff --check`
- [ ] `omh --help`
[...]

PR #1416, opened 2026-09-08
fix(hud): keep native delegation activity in its conversation

Joeywrz:
### What Changed

Attribute native delegation activity to the reading conversation when Hermes supplies a mapped durable identity. Preserve global activity when ownership is unavailable, and render its scope explicitly instead of implying it belongs to this chat. Preserve the existing OMH/Maestro, graph and todo behavior.

### Why This Exists

The native HUD reader queried recent children across the profile even when called for a specific conversation. Synthetic SQLite records reproduce a chat receiving another chat's agents. The original scope fix overreached by hiding the global OMH lane and changing todo fallback. The review corrections remove those regressions rather than requiring an unreleased host update.

### User / Operator Impact

- Mapped conversations see only their own native children, filtered before the database cap and derived totals.
- Unambiguous compression continuations are the same conversation; branches and delegated children are not.
- Missing, malformed and unmapped references retain native global fallback, including its existing manifest context. Rows and aggregates explicitly carry their scope.
- The actual widget labels local rows `[this chat]`, global rows `[global]`, and mixed aggregates `[this chat + global]`.
- OMH executor/Maestro activity, workflow state and graph remain available and globally attributed.
- Existing todo selection, write-time and clear behavior is unchanged. No agents are stopped and dispatch/delivery is unchanged.

### How It Works

The widget passes its own active-session-file reference. The native reader resolves the durable conversation and compression lineage and applies `_delegate_from` ownership before `LIMIT`. Successful resolution remains session-scoped even when the conversation has no children. Unavailable identity uses the prior global native read with `scope: global`; it does not substitute another durable conversation as the owner.

Timestamp-only manifests and route metadata do not prove ownership, so mapped native reads still omit that optional context. Global reads retain it. OMH status and graph projections remain intact and marked global. The original `_todo_summary` policy is restored, removing the extra duplicated validation introduced by the earlier version.

- `src/plugin_bundle/omh/hermes_delegation.py`: native ownership resolution/filtering, scope metadata and explicit global fallback.
- `src/plugin_bundle/omh/runtime_reader.py`: reference plumbing, retained global lanes and scope attribution.
- `src/omh/tui_widgets/omh-status.mjs`: real reader invocation and visible scope labels.
- `tests/test_tui_subagent_dag_mode.py`: align existing mock with native scope; no assertion removed.
- `tests/test_hud_conversation_scope.py`: ownership/lineage/caps, empty mapped conversations, fallback and actual shipped-widget regressions.
- `docs/INSTALLATION.md`: native scope and host compatibility. Original todo policy retained.

### Affected Surfaces

- [x] Hermes chat or wrapper
- [ ] Codex
- [ ] Claude Code
- [ ] Generic executor
- [x] Not executor-specific

## Validation

- [x] `git diff --check`
- [ ] `omh --help`
- [ ] Relevant workflow-learning review queue smoke
- [x] Relevant dry-run or smoke command (synthetic host/reader and managed installer dry-run)
- [ ] Manual Hermes/TUI check

### Observed Evidence

- Final head `d11a5f1fb6e6db58436b7527273e9b4861488a6e`: 175 HUD/native/widget/DAG/todo regressions passed. Original native isolation, restored global lanes, explicit fallback and real JS scope labels remain covered.
- Final-head static shard plan: **9,887 unique tests accounted for: 9,885 passed, one failed, one Windows-only skip, zero errors**. Failure: `test_fanout_confinement.FanoutFilesystemConfinementTests.test_verification_command_is_confined_when_it_has_no_owner_receipt`. The identical failure was reproduced on unchanged base `4058bad5` with the same unaffected shard ordering (only the new HUD test IDs omitted). No sandbox policy or assertion was weakened. This is NOT an all-green full-suite claim.
- Before final width correction/fixture alignment, monolithic discovery ran 9,893 tests on `dcbbf137` and 10,142 on current main `5c8e024c` plus all four PRs (`ac3eebfd`). Both had only the outdated DAG mock missing scope (one error, one skip). The mock now matches the real native reader. These earlier runs are not final-head passes.
- Final current-main integration `4aa446af`: 351 focused HUD/DAG/native/widget/todo/memory tests passed together after the last corrections. The local portfolio also passed its combined focused checks. #1413–#1415 are unchanged.
- Independent full-diff review confirmed the functional obligations, then found a scope-label width regression. Post-fix review cleared it and exercised 48 renderer cases at 80/100 columns with mixed/no scope and heterogeneous routes; 11 previously fitting legacy cases were byte-identical. This is real widget JS with a test SDK, not live terminal UI.
- Ruff, compileall, Node syntax, docs workflows/roles/capability-families byte gates, harness validation and `git diff --check` passed.
- Synthetic integration uses actual Hermes create-session RPC results, SQLite and the OMH reader: mapped conversations remain distinct; unknown/missing references retain explicitly global rows. No LLM or private live session data involved.
- Hosted CI is separate; prior fork CI required maintainer authorization. Updated exact-head state is reported in the correction reply.

No manual live two-chat GUI acceptance or live release smoke is claimed; the reader and actual JS renderer are exercised with isolated synthetic state. No release, installer or workflow-learning contracts change. Their broad release/checklist/workflow-learning commands are not the acceptance surface for this PR. Local installer verification, where performed, is separate from production process activation.

## Risk

Mapped native rows still omit optional manifest/task labels lacking ownership. Global fallback intentionally remains visible and is not a privacy boundary between profiles: labels disclose attribution, not authorization. Activity recency is not proof that a process is alive. Other profile-global metadata is not transformed into per-conversation data by this patch.

## Compatibility / Rollout

Existing Hermes create-session paths may publish a transport ID. This now remains usable with explicitly global native activity, not an empty widget. Once a durable ID arrives, native rows become conversation-scoped. Existing host fixes NousResearch/hermes-agent#18476 and #66446 are complementary improvements, not prerequisites. NousResearch/hermes-agent#105065 independently fixes the built-in delegation counter, not this database-backed panel.

## Release / Claims

- [x] Known manual Hermes checks are listed, including the live release-smoke gap

## Follow-Up

Only a producer-supplied trustworthy owner may make OMH/manifest activity conversation-scoped; timestamps do not establish it. The approved fixes #1413, #1414 and #1415 remain separate and unchanged.

### DIFF
--- a/src/omh/tui_widgets/omh-status.mjs
+++ b/src/omh/tui_widgets/omh-status.mjs
@@ -343,7 +343,8 @@ export default function register(sdk) {
     return { compact: Math.max(width.compact, cellWidth(routeCompact(text))), full: Math.max(width.full, cellWidth(text)) }
   }, { compact: 0, full: 0 })
 
-  const activityLayout = (row, columns, main, extraSeconds, tokensColumn, routeColumn) => {
+  const scopeLabel = row => row.scope === 'global' ? '[global] ' : row.scope === 'session' ? '[this chat] ' : ''
+  const activityLayout = (row, columns, main, extraSeconds, tokensColumn, routeColumn, scopeWidth) => {
     const state = safeText(row.state) || 'running'
     const stateText = columns < 100 ? ({ running: 'run', blocked: 'block', failed: 'fail' })[state] || state : state
     const taskId = truncateCells(safeText(row.task_id) || safeText(row.role) || 'agent', 8).padEnd(8)
@@ -403,7 +404,8 @@ export default function register(sdk) {
     // so they render as two pieces: same cells as before, split at the dot.
     const tailRest = ` · ${padCells(elapsedText(elapsed) || '0s', 7)}`
     const tailTokens = tokensColumn ? tokensPiece : ''
-    const prefix = `${taskId} `
+    const scope = padCells(scopeLabel(row), scopeWidth)
+    const prefix = `${scope}${taskId} `
     const separator = '  ·  '
     const budget = Math.max(24, columns - 4)
     // The route column exists per LIST, exactly like the tokens column: a
@@ -417,6 +419,9 @@ export default function register(sdk) {
     // title is what shrinks. `architect(claude-fable-5-1:xhigh)` still says
     // which lane, model and effort ran; `category:architect(claude-` does not.
     const routeRoom = budget - cellWidth(prefix) - tailWidth - cellWidth(separator) - 8 - 2
+    // If even the compact identity cannot fit beside a four-cell title,
+    // omit the whole shared route column rather than clipping the token tail.
+    if (routeColumn.compact > routeRoom + 4) routeColumn = { compact: 0, full: 0 }
     const routeShedPrefix = routeColumn.full > routeRoom
     const routeCap = routeShedPrefix ? routeColumn.compact : routeColumn.full
     const routeWidth = routeCap ? cellWidth(separator) + routeCap : 0
@@ -447,12 +452,13 @@ export default function register(sdk) {
       tailRest,
       tailState,
       tailTokens,
+      scope,
       taskId: main ? 'MAIN'.padEnd(8) : taskId,
     }
   }
 
-  function ActivityRow({ columns, extraSeconds, frame, main, row, routeColumn, t, tokensColumn }) {
-    const layout = activityLayout(row, columns, main, extraSeconds, tokensColumn, routeColumn)
+  function ActivityRow({ columns, extraSeconds, frame, main, row, routeColumn, scopeWidth, t, tokensColumn }) {
+    const layout = activityLayout(row, columns, main, extraSeconds, tokensColumn, routeColumn, scopeWidth)
     const blocked = row.state === 'blocked' || row.state === 'failed'
     const done = row.state === 'done'
     const marker = blocked ? '▲' : done ? '✓' : SPINNER_FRAMES[frame % SPINNER_FRAMES.length]
@@ -461,7 +467,7 @@ export default function register(sdk) {
       Text,
       { wrap: 'truncate-end' },
       h(Text, { color: blocked ? t.color.error : done ? t.color.ok : t.color.warn }, `${marker} `),
-      h(Text, { color: t.color.muted }, `${row.scope === 'global' ? '[global] ' : row.scope === 'session' ? '[this chat] ' : ''}${layout.taskId} `),
+      h(Text, { color: t.color.muted }, `${layout.scope}${layout.taskId} `),
       h(Text, { color: t.color.text }, layout.action),
       // Identity, then the measured block, then the rest. The route column
       // and the state/elapsed/tokens tail are fixed widths, so those figures
@@ -530,6 +536,8 @@ export default function register(sdk) {
     // it is what the fixed tail is anchored against, so every row reserves
     // it as soon as one row has a route to show.
     const routeColumn = routeColumnWidth([...mainRows, ...rows])
+    // Scope is a shared column too: mixed ownership must not shift the tail.
+    const scopeWidth = Math.max(0, ...[...mainRows, ...rows].map(row => cellWidth(scopeLabel(row))))
     return h(
       Box,
       { flexDirection: 'column', width: '100%' },
@@ -541,6 +549,7 @@ export default function register(sdk) {
           key: `main-${index}`,
           main: true,
           routeColumn,
+          scopeWidth,
           row,
           t,
           tokensColumn,
@@ -553,6 +562,7 @@ export default function register(sdk) {
           frame,
           key: `${safeText(row.task_id)}-${index}`,
           routeColumn,
+          scopeWidth,
           row,
           t,
           tokensColumn,
--- a/src/omh/tui_widgets/omh-status.mjs
+++ b/src/omh/tui_widgets/omh-status.mjs
@@ -21,16 +21,16 @@ export default function register(sdk) {
     'import json,os,sys',
     "sys.path.insert(0, os.path.join(os.environ['HERMES_HOME'], 'plugins'))",
     'from omh.runtime_reader import read_omh_hud',
-    "print(json.dumps(read_omh_hud(os.environ.get('OMH_HOME'), os.environ.get('HERMES_HOME'), graph_preference=os.environ.get('OMH_SUBAGENT_GRAPH', 'auto'), tui_session_ref=os.environ.get('OMH_HUD_TUI_SESSION_REF', ''))))",
+    "print(json.dumps(read_omh_hud(os.environ.get('OMH_HOME'), os.environ.get('HERMES_HOME'), graph_preference=os.environ.get('OMH_SUBAGENT_GRAPH', 'auto'), tui_session_ref=os.environ.get('OMH_HUD_TUI_SESSION_REF', ''), session_scoped=True)))",
   ].join(';')
   // This TUI's own session id. The host writes it to the file named by
   // HERMES_TUI_ACTIVE_SESSION_FILE whenever it creates, resumes, or switches
   // a session, and this widget runs inside that same TUI process, so the
   // file is the one identity the poll can carry that no other TUI shares.
-  // The reader scopes the plan todo to it. After a resume or switch the
-  // file holds the durable session key; on a freshly created session it
-  // holds the gateway's transport id instead, which the reader detects
-  // (no live row, no record) and answers as an identity-less poll would.
+  // The reader scopes todos, agent rows and their derived metrics to it.
+  // A fresh host session may expose only an unmapped transport id: leave
+  // session-local activity empty until a durable id arrives, never use MRU.
+  // Even an identity-less widget poll explicitly requests session scope.
   // A missing, unreadable, or malformed value is passed as nothing rather
   // than as a mutated string that would select the wrong record.
   const ACTIVE_SESSION_FILE = process.env.HERMES_TUI_ACTIVE_SESSION_FILE || ''
--- a/src/omh/tui_widgets/omh-status.mjs
+++ b/src/omh/tui_widgets/omh-status.mjs
@@ -27,10 +27,9 @@ export default function register(sdk) {
   // HERMES_TUI_ACTIVE_SESSION_FILE whenever it creates, resumes, or switches
   // a session, and this widget runs inside that same TUI process, so the
   // file is the one identity the poll can carry that no other TUI shares.
-  // The reader scopes todos, agent rows and their derived metrics to it.
-  // A fresh host session may expose only an unmapped transport id: leave
-  // session-local activity empty until a durable id arrives, never use MRU.
-  // Even an identity-less widget poll explicitly requests session scope.
+  // A mapped durable id scopes native rows and their derived metrics.
+  // Otherwise native activity remains visibly global, as does the OMH lane.
+  // The reader preserves the separate existing todo fallback policy.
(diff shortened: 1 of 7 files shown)


src/plugin_bundle/omh/runtime_reader.py
```diff
@@ -763,22 +763,26 @@ def read_omh_hud(
...
     safe_preset = preset if preset in HUD_PRESETS else "focused"
     home = _expand_path(omh_home) if omh_home else _default_omh_home()
     hermes = _expand_path(hermes_home) if hermes_home else _default_hermes_home()
     safe_limit = _safe_limit(limit, default=3)
-    status_payload = status if status is not None else _read_omh_hud_status(home, limit=safe_limit)
+    scoped = session_scoped or bool(session_ref or tui_session_ref)
+    reference = session_ref or tui_session_ref
+    # OMH run/binding projections identify executors and wrapper targets, not
+    # the originating Hermes conversation. Do not invent that relationship.
+    status_payload = {} if scoped else (
```
sionic-khope: Important: every widget read is scoped now, so this line drops `maestro.rows`, executor subagent counts, `runtime.workflow`, and flips `active` to false for the widget. OMH-lane rows never had Hermes conversation ownership and were not the source of the cross-chat bleed; please keep `status_payload` intact for scoped reads and apply the ownership filter only to native `state.db` rows.

src/omh/tui_widgets/omh-status.mjs
```diff
@@ -21,16 +21,16 @@ export default function register(sdk) {
     'import json,os,sys',
     "sys.path.insert(0, os.path.join(os.environ['HERMES_HOME'], 'plugins'))",
     'from omh.runtime_reader import read_omh_hud',
-    "print(json.dumps(read_omh_hud(os.environ.get('OMH_HOME'), os.environ.get('HERMES_HOME'), graph_preference=os.environ.get('OMH_SUBAGENT_GRAPH', 'auto'), tui_session_ref=os.environ.get('OMH_HUD_TUI_SESSION_REF', ''))))",
+    "print(json.dumps(read_omh_hud(os.environ.get('OMH_HOME'), os.environ.get('HERMES_HOME'), graph_preference=os.environ.get('OMH_SUBAGENT_GRAPH', 'auto'), tui_session_ref=os.environ.get('OMH_HUD_TUI_SESSION_REF', ''), session_scoped=True)))",
```
sionic-khope, line 24: Important: on the installed host (`ui-tui/src/app/useSessionLifecycle.ts`) a freshly created session writes the transport `session_id` to the active-session file and only resume/switch writes the durable `session_key`. With `session_scoped=True` unconditional, a new chat shows no native rows, no todo, and no OMH lane until it is resumed. Suggest: when the reference maps to no durable session, keep the global native rows and tag them (for example `scope: "global"`) so the widget renders them distinctly instead of hiding them.

src/plugin_bundle/omh/runtime_reader.py
```diff
@@ -1778,19 +1785,21 @@ def _todo_summary(
...
-        # the owner of a record, which is what a fresh session's transport id
-        # looks like. Read as a widget with no identity would, rather than
-        # hiding the plan this TUI is most likely looking at.
-        session_id, session = _reading_session(hermes, "")
-        record, own_record = _own_todo_record(home, session_id)
     if not own_record:
         # No per-session record: the home-wide file answers, gated below by
         # the identity or write-time rule so another session's plan stays out.
         record = _read_hud_json(todo_path(home), root=home)
+    if scoped and session is None and not own_record and record.get("session_ref") != session_id:
```
sionic-khope: This reverses the documented rule in `docs/INSTALLATION.md` (hiding a current checklist on missing evidence is the worse failure) and flips two existing test expectations. That is an owner decision; please split it out or leave the todo behavior as it was in this PR.

sionic-khope (changes requested):
Thanks for this. The bug is real and the native-ownership part of the fix is well built: the compression-lineage CTE excludes branch and delegate children correctly, ownership is applied before `LIMIT 32`, and the widget test drives the shipped JS reader. Verified locally at `b84c05cc` merged onto current `main` together with #1413, #1414, #1415: full suite 9,909 tests OK (1 skip), Ruff, `git diff --check`, docs byte gates pass, so the missing full-suite claim in the body is covered.

Requesting changes for three product-level regressions that the PR treats as accepted consequences but that the owner has not decided. Inline comments carry the `file:line` for each.

**2 important, 1 nit.**

**Important 1 — the widget loses the OMH lane entirely.** `read_omh_hud` sets `status_payload = {}` whenever `scoped` is true (`runtime_reader.py:783`), and the widget now always passes `session_scoped=True` (`omh-status.mjs:24`). Reproduced: with one active `maestro` executor in status, the global read returns `maestro.rows=[...]`, `subagents.active=1`, `active=true`; the scoped read returns `maestro.rows=[]`, `subagents.active=0`, `active=false`, and `graph.reason="session_ownership_unavailable"` (`runtime_reader.py:866`). `omh coding fanout dispatch` and Maestro activity therefore disappear from the TUI widget, which is the primary consumer of those rows. "They remain available globally" does not help, because the widget never issues a global read any more. The native-ownership filter should not need to blank the OMH lane; those rows never had Hermes conversation ownership and were never the source of the cross-chat bleed.

**Important 2 — freshly created sessions render an empty widget on the installed host.** The PR acknowledges this for "older Hermes create-session paths", but that is the current path: `ui-tui/src/app/useSessionLifecycle.ts` writes `r.session_id` (transport id) on create and only writes `r.session_key` on resume/switch. The referenced host fixes (hermes-agent#18476, #66446) are not in the release users run. Combined with `_todo_summary` returning `empty` for any unmapped reference (`runtime_reader.py:1801`) and manifests being dropped for scoped reads (`hermes_delegation.py:1456`), the default experience becomes: open a new chat, dispatch agents, see nothing until you resume. `docs/INSTALLATION.md` previously recorded the opposite decision ("hiding a legitimately current checklist on missing evidence is the worse failure"); reversing it is an owner call, not something to settle inside a bug fix.

**Suggested shape.** Keep the ownership filter on native `state.db` rows (that part is right), but leave `status_payload`, the graph, and the todo fallback as they are for the widget. Where no durable id is available, fall back to the previous global native rows and mark them (for example `scope: "global"` on the row) so the widget can render them distinctly rather than hide them. That fixes the cross-chat attribution without removing information users currently rely on.

**Nit** — the session-reference sanitizer (`1 <= len <= N` plus the `_.:@-` alphabet) is duplicated in `hermes_delegation._conversation_session_ids` and `runtime_reader._todo_summary` with different length caps (160 vs `MAX_TODO_SESSION_REF_CHARS`). One helper would keep them from drifting.

Joeywrz:
Addressed the requested product regressions in `dcbbf137` and the final head `d11a5f1fb6e6db58436b7527273e9b4861488a6e`:

- OMH status/workflow, executor/Maestro rows and graph are retained, explicitly globally attributed. The widget renders `[global]`, `[this chat]`, and mixed aggregate labels.
- Missing/invalid/unmapped durable identity uses the existing native global fallback, including manifest context, with explicit row/aggregate scope. This no longer requires an unreleased Hermes host fix. A mapped conversation with no children remains empty rather than borrowing another conversation.
- `_todo_summary` and its original tests/policy are restored; the duplicated sanitizer introduced there is removed. Native state.db ownership remains before LIMIT and follows only the unambiguous compression lineage.
- Independent review caught scope badges exceeding the row width. `95c08de3` budgets a shared scope column, retains full routes when they fit, and sheds an oversized route column before clipping tokens. Actual-renderer 80/100-column global/local/mixed cases and legacy cases now pass. The final commit aligns an existing DAG test mock with the native reader's scope contract, without removing assertions.

Verification: 175 focused regressions; final current-main + all-four-PR integration: 351 focused tests. Final-head static suite accounts for all 9,887 planned tests: 9,885 passed, one Windows-only skip, one sandbox-confinement failure reproduced on unchanged base with the same unaffected shard ordering. Zero errors. Detailed earlier/full-suite boundaries are in the updated PR body; no all-green suite claim. Independent post-fix review found no remaining blocker/high/medium issue.

Exact-head hosted state just checked: **CI: completed/action_required**. Hosted CI approval is separate from the local results. No merge performed. #1413–#1415 remain unchanged.

sionic-khope (approved):
Round two, scoped to the delta since `b84c05cc` (`dcbbf137`, `95c08de3`, `d11a5f1f`). All three findings from round one are resolved; no new Important findings.

- **OMH lane retained.** `status_payload` is read unconditionally again (`runtime_reader.py:782`), and maestro/runtime/executor/graph carry `scope: "global"`. Reproduced: a scoped read with one active maestro executor now returns `maestro.rows=[...]`, `active=true`, `graph.scope="global"`.
- **Unmapped identity falls back to global native rows.** `_query_state_db` keeps the unfiltered query when the reference maps to no durable session and reports `scope: "global"` (`hermes_delegation.py:1304`); manifests are read in that mode. Reproduced: `tui_session_ref="3f9a1c2b"` with `session_scoped=True` returns both native children tagged `global`; a mapped id returns only its own child tagged `session`, with `subagents.scope="mixed"` when OMH rows are present.
- **Todo policy restored.** `_todo_summary` is back to the original fallback and the two flipped tests in `tests/test_hud.py` carry their original expectations.

Verified locally at `d11a5f1f` merged onto current `main` (after #1413, #1414, #1415 landed): full suite 10,163 tests OK (1 skip), Ruff, `git diff --check`, `docs workflows --check`, `node --check` on the widget all pass.

Nit (not blocking, carried from round one): the reference sanitizer now lives only in `_conversation_session_ids`, which resolves the duplication.

sionic-khope (changes requested):
Round three: hosted CI on `d11a5f1f` is red on `test-windows (1)` (run 34210090074), and both failures are in the new `tests/test_hud_conversation_scope.py`, so this blocks the merge despite the approval above. Linux, macOS, quarantine, and distribution jobs all pass, and the latest `main` runs are green on Windows, so this is PR-introduced. **1 important.**

**Important — new Node-driven tests fail on Windows.**

1. `test_widget_renders_owned_rows_and_explicit_global_fallback` errors with `CalledProcessError` from the `node --input-type=module -e ...` step. The script does `await import(process.argv[1])` with a bare Windows path (`C:\Users\...\widget.mjs`). ESM `import()` needs a `file://` URL on Windows; `pathToFileURL(process.argv[1]).href` from `node:url` fixes that. Please confirm on the log whether that is the whole story, since `--log-failed` truncates the Node stderr.

2. `test_scoped_activity_rows_fit_and_keep_token_columns` fails every 80/100-column subtest with `'12.3k tokens' not found in line[:cols]`. The rendered lines on Windows are 2–3 cells wider than on Linux/macOS (the tail lands at `12.3k to`), so this is not the import issue. Likely the width measurement differs on that runner (locale, `…`/`·` cell width) rather than the layout itself, but I have not verified which; either make the test platform-independent or pin the expected behavior on Windows explicitly. A blanket `skipIf(sys.platform == "win32")` is acceptable only if the widget itself is documented as untested on Windows.

These are the first tests in the repo that execute Node in CI (`tests/test_tui_widget_pack.py` and `tests/test_hud.py` never spawn it), so there is no existing pattern to copy; whatever guard you choose becomes the pattern.

Local verification on the current head is unchanged from round two (full suite 10,163 OK on macOS). I will re-approve once the Windows shard is green.

sionic-khope (approved):
Round three resolved at `f1941833`: test-only change (`pathToFileURL` for the ESM import, explicit UTF-8 decoding of the Node subprocess output, which also explains the Windows width mismatch since the default code page mangled the ellipsis and middle-dot glyphs). Hosted CI is green on every job including both Windows shards (run 34227015793). Product code unchanged since the round-two full-suite run. Approving and merging.

PR #1430, opened 2026-09-09
fix(install): handle symlinked plugin dirs in _copy_plugin_bundle

JuznemHub:
### What Changed

- Added `_discard_path()` helper in `src/install/plugin_pack.py` that removes symlinks, regular files, and directories uniformly (unlink for symlinks/files, rmtree for dirs).
- Replaced all four `shutil.rmtree(..., ignore_errors=True)` calls in `_copy_plugin_bundle` with `_discard_path()`.
- Added `target.is_symlink()` to the target-exists check so a symlinked plugin directory is properly backed up and replaced with a real directory.
- Added regression test `test_update_replaces_a_symlinked_bundle_without_crashing` in `tests/test_plugin_distribution.py`.

### Why This Exists

`omh update` crashed with `NotADirectoryError: [Errno 20]` when a bot-profile plugin directory was a symlink — a leftover from an older OMH layout where profiles symlinked `plugins/omh` to a shared location.

The crash sequence was:
1. `target.rename(backup)` renamed the symlink to `.omh.previous` (this works — rename on a symlink just moves it).
2. `shutil.rmtree(backup, ignore_errors=True)` silently failed to remove the symlink (rmtree cannot unlink symlinks; it descends into the target and fails, then swallows the error).
3. The stale `.omh.previous` symlink survived across updates.
4. On the next update, `target.rename(backup)` tried to rename a real directory over the existing symlink. Linux `rename(2)` refuses `rename(dir, symlink)` with `ENOTDIR`, crashing the entire `omh update` for every bot profile.

This affected any multi-profile setup where the `law` (or other bot) profile's plugin dir was originally a symlink, which is the case on this machine.

### User / Operator Impact

- `omh update` no longer crashes on profiles with symlinked plugin directories.
- `omh update` is now idempotent across consecutive runs even when a stale `.omh.previous` symlink exists from a prior layout.
- Operators no longer need to manually remove stale symlinks before running `omh update`.

### How It Works

`_discard_path(path)` checks the filesystem entry type:
- **Symlink** → `path.unlink()` (removes the link itself, not the target)
- **Directory** → `shutil.rmtree(path, ignore_errors=True)`
- **Regular file** → `path.unlink()`

Every temporary and backup path in `_copy_plugin_bundle` (`tmp`, `backup`, and the post-rename cleanup) now goes through `_discard_path`, so no stale leftover — regardless of type — can block a subsequent rename. The `target.exists()` guard also gains `target.is_symlink()` so a symlinked plugin dir enters the backup path instead of being skipped.

- `src/install/plugin_pack.py` — new `_discard_path()` helper + updated `_copy_plugin_bundle()`
- `tests/test_plugin_distribution.py` — new test + `import shutil`

### Affected Surfaces

- [ ] Hermes chat or wrapper
- [ ] Codex
- [ ] Claude Code
- [ ] Generic executor
- [x] Not executor-specific

## Validation

- [ ] `uv run --group lint ruff check src tests`
- [x] `PYTHONPATH=tests uv run python -m unittest discover -s tests -v` (plugin distribution suite: 52/52 pass)
- [ ] `uv run python -m compileall -q src tests`
- [ ] `git diff --check`
- [ ] `omh --help`
- [x] Manual Hermes/TUI check, or explicit reason it was not run: `omh update` and `omh doctor` run clean on a machine with a real stale `.omh.previous` symlink on the `law` profile.

### Observed Evidence

- Targeted tests: `PYTHONPATH=tests uv run python -m unittest tests.test_plugin_distribution -v` → 52/52 pass, including the new `test_update_replaces_a_symlinked_bundle_without_crashing`.
- Manual: `omh update` on the affected machine (law profile had a stale `.omh.previous` symlink → `/home/senor/.hermes/plugins/omh`). Before fix: `NotADirectoryError [Errno 20]`. After fix: exit 0, "Bot profiles: law (refreshed)".
- Manual: `omh doctor` → "Hermes targets: ok (2/2)".

- Full test suite (`unittest discover -s tests`): the CLI test module timed out (180s) due to subprocess-heavy tests unrelated to this change. The targeted plugin distribution suite (52 tests) fully exercises the changed code paths.
- `ruff check`: not run (ruff not available in this environment).
- `compileall`, `docs --check`, `harness validate`, `release` smoke: not run — this change touches only the plugin install path, not docs, skills catalog, generated artifacts, or release tooling.
- Windows: `rename(2)` semantics differ, but the symlink code path is POSIX-specific.

## Risk

Low. The change is scoped to `_copy_plugin_bundle` internals. The atomic-rename contract is preserved — the backup is still created via rename, and the rollback path still restores it on error. The only behavioral change is that symlinks/files in the tmp/backup paths are now correctly removed instead of silently surviving.

## Compatibility / Rollout

No compatibility concerns. Older layouts that used symlinks will now be transparently replaced with real directories on the next `omh update` — which is the intended behavior. No config or schema changes.

## Release / Claims

- [x] Release-channel impact considered (`stable`, `preview`, or `local`) — this is a bugfix for `preview` channel; no release-channel change.
- [x] Runtime/native capability claims are backed by evidence or marked as not observed — plugin install smoke is local only; Hermes runtime load not claimed.
- [x] Prepared handoffs or plans are labeled `prepared_not_observed`, never as execution, review, CI, or merge evidence — N/A (no handoffs in this PR).
- [x] Evidence contains no credentials, raw prompts, raw platform events, transcripts, or unrelated private data.
- [x] Known manual Hermes checks are listed — `omh update` and `omh doctor` run manually; `omh release hermes-smoke --live` gap noted.

## Follow-Up

- Consider adding a Windows-specific path in `_discard_path` if OMH ever supports Windows plugin dirs (currently POSIX-only for symlinks).

### DIFF
--- a/src/install/plugin_pack.py
+++ b/src/install/plugin_pack.py
@@ -251,23 +251,44 @@ def _collect_resource_records(root: Any, rel: Path, records: list[PluginFileReco
             records.append(PluginFileRecord(str(item_rel), sha256_text(item.read_text(encoding="utf-8"))))
 
 
+def _discard_path(path: Path) -> None:
+    """Remove *path* whether it is a symlink, a regular file, or a directory.
+
+    ``shutil.rmtree`` alone cannot remove a symbolic link — it errors (or
+    silently no-ops with ``ignore_errors=True``) and leaves the link behind.
+    A stale ``.omh.previous`` symlink from an older layout then causes the
+    subsequent ``target.rename(backup)`` to fail with ``ENOTDIR``, because
+    Linux ``rename(2)`` refuses to rename a directory over an existing
+    non-directory.  Every temporary/backup path in ``_copy_plugin_bundle``
+    goes through this helper so the install is robust against leftovers from
+    any prior layout.
+    """
+    if path.is_symlink():
+        path.unlink()
+    elif path.is_dir():
+        shutil.rmtree(path, ignore_errors=True)
+    elif path.exists():
+        path.unlink()
+
+
 def _copy_plugin_bundle(target: Path, file_records: list[dict[str, str]]) -> None:
     root = resources.files("omh.plugin_bundle.omh")
     parent = target.parent
     tmp = parent / f".{target.name}.installing"
     backup = parent / f".{target.name}.previous"
     ensure_dir(parent)
-    shutil.rmtree(tmp, ignore_errors=True)
-    shutil.rmtree(backup, ignore_errors=True)
+    _discard_path(tmp)
+    _discard_path(backup)
     try:
         _copy_resource_tree(root, tmp)
         atomic_write_json(tmp / PLUGIN_MANAGED_MANIFEST, _new_plugin_manifest(target, file_records))
-        if target.exists():
+        if target.exists() or target.is_symlink():
+            _discard_path(backup)
             target.rename(backup)
         tmp.rename(target)
-        shutil.rmtree(backup, ignore_errors=True)
+        _discard_path(backup)
     except OSError:
-        shutil.rmtree(tmp, ignore_errors=True)
+        _discard_path(tmp)
         if backup.exists() and not target.exists():
             backup.rename(target)
         raise
--- a/tests/test_plugin_distribution.py
+++ b/tests/test_plugin_distribution.py
@@ -5,6 +5,7 @@ import importlib.resources as resources
 import importlib.util
 import json
 import os
+import shutil
 import subprocess
 import sys
 import tomllib
@@ -1312,6 +1313,35 @@ class UpdateRefreshesTheBundleTests(unittest.TestCase):
             status, _, stderr = run_cli(base + ["update"])
             self.assertEqual(status, 0, stderr)
             self.assertEqual(marker.read_text(encoding="utf-8"), "# edited outside OMH\n")
+
+    def test_update_replaces_a_symlinked_bundle_without_crashing(self) -> None:
+        # Older OMH installs sometimes left the plugin directory as a symlink
+        # to a shared location.  A subsequent `omh update` renamed that symlink
+        # to ``.omh.previous``, but ``shutil.rmtree`` cannot unlink a symlink,
+        # so the stale backup survived.  The next update then hit
+        # ``rename(dir, symlink)`` → ``ENOTDIR`` and crashed.  The fix
+        # (``_discard_path``) removes symlink/file/directory leftovers before
+        # every rename, so two consecutive updates on a symlinked bundle both
+        # succeed and leave a real directory.
+        with TemporaryDirectory() as tmp:
+            root = Path(tmp)
+            base = ["--omh-home", str(root / ".omh"), "--hermes-home", str(root / ".hermes")]
+            run_cli(base + ["setup"])
+            bundle = self._bundle_dir(root / ".hermes")
+            shared = root / ".hermes" / "shared-omh"
+            shutil.move(str(bundle), str(shared))
+            bundle.symlink_to(shared)
+
+            status, _, stderr = run_cli(base + ["update"])
+            self.assertEqual(status, 0, stderr)
+            self.assertFalse(bundle.is_symlink())
+            self.assertTrue((bundle / "memory_provider.py").is_file())
+
+            # Second update proves the stale-symlink path is gone — no ENOTDIR.
+            status, _, stderr = run_cli(base + ["update"])
+            self.assertEqual(status, 0, stderr)
+            self.assertTrue((bundle / "memory_provider.py").is_file())
+            self.assertFalse((bundle.parent / ".omh.previous").is_symlink())
 
 
 class UpdateCarriesRegistrationTests(unittest.TestCase):


src/install/plugin_pack.py
```diff
@@ -251,23 +251,44 @@ def _collect_resource_records(root: Any, rel: Path, records: list[PluginFileReco
...
     ensure_dir(parent)
-    shutil.rmtree(tmp, ignore_errors=True)
-    shutil.rmtree(backup, ignore_errors=True)
+    _discard_path(tmp)
+    _discard_path(backup)
     try:
         _copy_resource_tree(root, tmp)
         atomic_write_json(tmp / PLUGIN_MANAGED_MANIFEST, _new_plugin_manifest(target, file_records))
-        if target.exists():
+        if target.exists() or target.is_symlink():
```
sionic-khope, line 285: **Important — a dangling symlink here is destroyed without `--force`.** Carried forward from round one (Codex P2 on this line), independently reproduced.

`or target.is_symlink()` only changes behavior for a symlink that does **not** resolve; one that resolves already passes `target.exists()`. A dangling link also slips past the ownership guard at `src/install/plugin_pack.py:49`, because `Path.exists()` follows the link and reports `False`. So this branch renames the operator's link to `.omh.previous` and `_discard_path(backup)` deletes it, with no `--force` and no message. Observed: `omh update` exits 0 and the link is gone.

On the parent commit the same input crashed instead (`NotADirectoryError [Errno 20]` on `tmp.rename(target)`), so the guard was never bypassed before this line existed.

The new regression test uses a symlink that resolves, so it does not exercise this clause at all. I removed the clause and the test still passes (`Ran 1 test ... OK`).

Fix the guard rather than this line — see Important 1 in the review body:

```python
# src/install/plugin_pack.py:49
unmanaged = (target.exists() or target.is_symlink()) and existing_manifest is None
```

Verified: with that applied, a dangling link survives `omh update` untouched and all 52 tests in this module still pass.

src/install/plugin_pack.py
```diff
@@ -251,23 +251,44 @@ def _collect_resource_records(root: Any, rel: Path, records: list[PluginFileReco
...
+            _discard_path(backup)
             target.rename(backup)
         tmp.rename(target)
-        shutil.rmtree(backup, ignore_errors=True)
+        _discard_path(backup)
     except OSError:
-        shutil.rmtree(tmp, ignore_errors=True)
+        _discard_path(tmp)
         if backup.exists() and not target.exists():
```
sionic-khope, line 292: **Important — this rollback cannot restore what line 285 now backs up.**

Line 285 gained `or target.is_symlink()`; this guard kept the bare `exists()`. The one case the new branch admits is therefore the one case the rollback cannot undo. Reproduced by forcing an `OSError` on `tmp.rename(target)`:

| Original target | After a failed install |
| --- | --- |
| symlink to a real directory | restored, no leftovers |
| dangling symlink | **target gone**, stale `.omh.previous` left behind |

That contradicts this PR's Risk section ("the rollback path still restores it on error"), and it leaves behind the exact stale `.omh.previous` the change set out to eliminate.

```suggestion
        if (backup.exists() or backup.is_symlink()) and not (target.exists() or target.is_symlink()):
```

Verified: with this applied the dangling target is restored with no leftover, and the 52 tests in this module still pass.

sionic-khope (commented):
**2 important, 0 nits posted** (2 new nits withheld under the round-two suppression rule)

**Round 2**, ratcheting against `eb4a60c609e60789822b5b40bbdf42b72b495051` — the same commit this review is posted against. Round one was the automated Codex connector review at that SHA. Its body claims no scope and it named exactly one finding, so under `REVIEW.md` **Settled ground** it settles only that one point. Everything else below is ground nobody examined, not ground being re-opened. Both Important findings carry the novelty admission the ratchet requires: **I did not examine this PR last round.**

## Shape of the work

Two changes travel together in `_copy_plugin_bundle`:

1. Four `shutil.rmtree(..., ignore_errors=True)` calls become `_discard_path()`, which unlinks symlinks instead of silently failing on them.
2. The target-exists check gains `or target.is_symlink()`.

**Part 1 is correct, and it alone fixes the reported crash. Part 2 is unexercised by the new test and is the sole cause of both findings below.**

## Gates

Run in a private worktree at `eb4a60c6`, nothing else applied:

```
PYTHONPATH=tests uv run python -m unittest discover -s tests
    Ran 9592 tests in 614.262s
    OK (skipped=1)

uv run --group lint ruff check src tests                     All checks passed!
uv run python -m compileall -q src tests                     exit 0
uv run python -m omh.cli docs workflows --check              {"ok":true,"tap_skills":{"checked":190,"expected":190,"extra":[],"missing":[],"ok":true,"stale":[]}}
uv run python -m omh.cli docs roles --check                  {"ok":true}
uv run python -m omh.cli docs capability-families --check     {"ok":true}
git diff --check                                             exit 0
```

The full suite is green, so the Validation checkbox for `unittest discover` is substantiated even though the Not-Tested section says that run timed out locally. Nothing to fix there. Note that GitHub CI on this PR has run only DCO so far, so the `windows-latest` leg in `.github/workflows/ci.yml` is unobserved by anyone.

## The reported bug and Part 1 of the fix are real

I reproduced the entire causal chain from the PR body. It holds on macOS as well as Linux:

- `shutil.rmtree(symlink, ignore_errors=True)` leaves the link in place. Observed.
- After one `omh update` on a symlinked bundle, `.omh.previous` survives as a symlink. Observed.
- The next update's `rename(dir, symlink)` fails `ENOTDIR`. Observed.

The regression test genuinely fails without the source change. Reverting only `src/install/plugin_pack.py` to its parent commit and running the new test twice gives the exact claimed failure, deterministically, on the second update:

```
NotADirectoryError: [Errno 20] Not a directory:
  '.../.hermes/plugins/omh' -> '.../.hermes/plugins/.omh.previous'
```

That is a well-built regression test for a real bug. The findings below are about Part 2 only.

## Important 1 — a dangling plugin symlink is destroyed without `--force`

Carried forward from round one, where the Codex connector raised it as P2 on the same line. It is unresolved, so it stays at its original severity. I reproduced it independently rather than taking it on trust.

`install_plugin_bundle` refuses to replace an entry it cannot prove it owns:

```python
# src/install/plugin_pack.py:49
unmanaged = target.exists() and existing_manifest is None
if unmanaged and not force:
    raise PluginPackError(f"{target} exists without an OMH plugin manifest; use --force to replace it")
```

`Path.exists()` follows the link, so a **dangling** symlink reports `False` and never reaches that guard. Before this PR that was harmless: `_copy_plugin_bundle` skipped the backup, `tmp.rename(target)` hit the dangling link, and the command crashed. Observed on the parent commit:

```
NotADirectoryError: [Errno 20] Not a directory:
  '.../.hermes/plugins/.omh.installing' -> '.../.hermes/plugins/omh'
```

With `or target.is_symlink()` the branch is now entered, the link is renamed to `.omh.previous`, and `_discard_path(backup)` deletes it. Observed on this branch: `omh update` exits 0 and the operator's link is gone, with no `--force` and no message.

The guard still works for a symlink that resolves to a foreign directory, because `exists()` follows it and the manifest read comes back empty. Only the dangling case slips through, and this PR is what makes it reachable.

The repository's own posture for the sibling installers is stricter still: `_reject_symlink_path()` at `src/omh/tui_widget_pack.py:26` and `src/omh/skin_pack.py:161` refuse a symlinked destination outright rather than converting it.

Fix at the guard, not in the copy helper (outside the diff hunk, so it cannot be a suggestion block):

```python
# src/install/plugin_pack.py:49
unmanaged = (target.exists() or target.is_symlink()) and existing_manifest is None
```

I applied that and re-ran: a dangling link now survives `omh update` untouched instead of being deleted, and all 52 tests in `tests/test_plugin_distribution.py` still pass. `_refresh_installed_plugin_bundle` already swallows `PluginPackError` by design (`src/commands/setup.py:637`), so update stays non-fatal and `omh doctor` reports the drift — which is the behavior that comment describes.

## Important 2 — the rollback cannot restore what the new branch backs up

Novelty admission, per the ratchet: this was not raised last round because **I did not examine this PR last round**. It is a defect in the delta, in code this PR adds.

The PR extended one `exists()` check and left the matching one four lines below unchanged:

```python
if target.exists() or target.is_symlink():   # line 285, updated by this PR
    ...
except OSError:
    _discard_path(tmp)
    if backup.exists() and not target.exists():   # line 292, not updated
        backup.rename(target)
```

So the one case Part 2 newly admits into the backup branch is exactly the case the rollback cannot undo. Reproduced by forcing an `OSError` on `tmp.rename(target)`:

| Original target | After a failed install |
| --- | --- |
| symlink to a real directory | restored correctly, no leftovers |
| dangling symlink | **target gone**, stale `.omh.previous` left behind |

That contradicts the Risk section of this PR body, which states that "the atomic-rename contract is preserved ... the rollback path still restores it on error." For the dangling case it does not, and it also leaves behind the very stale `.omh.previous` this PR set out to eliminate. A false behavioral claim in the body is what `REVIEW.md` reserves Important for, independently of the ratchet.

A suggestion is attached inline at line 292. I verified it: with it applied, the dangling target is restored and no leftover remains, and the 52 plugin-distribution tests still pass.

## One clause, both defects

Worth stating plainly, because it makes the fix cheap. `or target.is_symlink()` changes behavior **only** for a dangling symlink — a symlink that resolves already passes `target.exists()`. The new test uses a symlink that resolves.

I removed `or target.is_symlink()` and left everything else in place. The new regression test still passes:

```
Ran 1 test in 0.933s
OK
```

So the clause is untested, is not needed for the bug this PR fixes, and is the only source of both findings. Dropping it plus fixing the guard as in Important 1 gets the reported crash fixed with neither defect.

## Withheld nits

Two new Nits are suppressed under the round-two rule, which limits this round to Important findings. Neither blocks; both are consistency points with an existing `file:line` alternative in this repository.

## PR #1439 overlap

#1439 (`Som0111`) edits the same two files. I checked for a real collision rather than a textual one. It touches `_SmokeContext`, `_register_smoke`, and `inspect_plugin_bundle`; this PR touches `_copy_plugin_bundle`. No shared state. Merging `origin/main` + this PR + #1439 in sequence auto-merges with no conflict, and `tests/test_plugin_distribution.py` then runs 60 tests, all passing. Reporting only; sequencing is the maintainer's call.

## Verification boundaries

- Windows is unverified by me. I ran nothing on Windows and CI has not run its Windows leg on this PR.
- Every behavioral claim above was reproduced in a worktree at this SHA, except where labeled otherwise.

Posted as a comment. This procedure reports; the maintainer decides.

sionic-khope:
Superseded by #1441, which is now merged.

Thank you for this one. The diagnosis was correct and it reproduced exactly as written, on macOS as well as Linux: `rmtree` leaves a symlink in place, the stale `.omh.previous` survives one update, and the next update's `rename(dir, symlink)` gives ENOTDIR. Your regression test genuinely went red without the source change, with the exact error. That analysis is what made the fix cheap, and #1441 carries your `Co-authored-by` for it.

Two things changed on the way in, both from the review on this PR:

- The symlink-aware discard is kept, but as a shared `discard_path` in `src/system/local_store.py` rather than a new private helper. The type dispatch already existed in the installer's removal collector and the junction probe already existed in the self-update platform, so the shared form replaces both instead of a third one existing. It is junction-aware, which the original was not.
- The `or target.is_symlink()` clause was unsafe on its own: a dangling link was destroyed without `--force`, because `exists()` follows the link and never reaches the ownership guard, and the rollback could not restore that one case. #1441 keeps the clause but pairs it with a link-aware ownership guard and a link-aware rollback, all spelled `is_directory_link` so a dangling junction is covered too. Each half was reverted separately to confirm it is load-bearing on its own.

All four cells now come out right, with no `.omh.previous` or `.omh.installing` left behind:

| case | result |
| --- | --- |
| resolvable, no `--force` | exit 0, replaced |
| resolvable, `--force` | exit 0, replaced |
| dangling, no `--force` | exit 0, link kept, note names `--force` |
| dangling, `--force` | exit 0, replaced |

Closing this one as superseded rather than as rejected.

PR #1439, opened 2026-09-09
feat(plugin): validate plugin tool schemas before registration smoke passes

### LINKED ISSUE #1437: [Feature]: Validate plugin tool schemas before registration smoke passes
## Problem

OMH can report a successful plugin registration smoke even when a registered tool has a missing or malformed parameter schema. The smoke context records only tool names and discards each toolset, schema, handler, and description, so the release gate cannot distinguish a callable registration from a provider-safe tool definition. A malformed definition can therefore ship past setup and doctor and fail only when Hermes prepares a model request.

## Complete user goal

Ensure every OMH plugin tool is rejected before installation is reported healthy unless its registered definition satisfies the minimum Hermes tool contract. Users should receive a named, actionable setup or doctor failure instead of encountering a later model-request failure caused by one bad tool schema.

## Intended surfaces

Hermes runtime or handoff

Hermes chat or wrapper

## Observed evidence

On current OMH main:

- `src/install/plugin_pack.py::_SmokeContext.register_tool()` appends only the tool name and accepts all positional and keyword arguments without inspecting them.
- `src/install/plugin_pack.py::_register_smoke()` marks `register_smoke` successful when the expected tool and hook names are present; it has no schema-conformance result.
- `src/plugin_bundle/omh/__init__.py::register()` passes one schema object for each bundled tool, but there is no shared validation at the OMH packaging boundary for schema name, description, or an object-shaped `parameters` field.
- A process-local check registered a synthetic tool definition with no `parameters` field through `_SmokeContext`; the context accepted it and recorded the name.
- All nine currently bundled OMH tool schemas have object-shaped `parameters`, so this is a release-regression blind spot rather than a confirmed failure in the current bundle.

Existing plugin registration and host-load work verifies that tools are present and distinguishes local smoke from observed Hermes loading. It does not validate the model-facing schema contract described here.

## Proposal

Strengthen the existing plugin registration smoke instead of adding a parallel validator:

1. Capture the registered tool name, toolset, schema, handler, and public description in `_SmokeContext`.
2. Validate each bundled tool definition during `_register_smoke()` before `plugin_register_smoke` can pass.
3. At minimum require a mapping-shaped schema whose `name` matches the registered name, a non-empty description, and a mapping-shaped `parameters` schema with `type: object`. Validate `properties` and `required` coherence when those fields are present.
4. Return bounded per-tool failures from `inspect_plugin_bundle()` and surface them through setup and doctor with the failing tool name and violated rule.
5. Exercise the bundled `register(ctx)` path with behavior-level fixtures so a malformed schema proves the gate red without reading source text.

Keep this as deterministic local conformance. It should not execute a provider request and must not upgrade local registration smoke into proof that Hermes loaded or invoked the plugin.

## Alternatives considered

- Rely on type hints and the current schema constants. Rejected because Python runtime registration does not enforce the annotations and the existing smoke explicitly accepts arbitrary arguments.
- Validate only that every expected tool name registered. This is the current behavior and cannot catch malformed model-facing definitions.
- Add a separate schema-lint command. Rejected because installation and doctor already own the plugin health decision; splitting the check would allow the primary health signal to remain green.
- Require a live model request in setup or CI. Rejected because it introduces credentials, network variability, cost, and provider coupling for a deterministic contract check.

## Success criteria

- A fixture with one registered tool missing `parameters` makes `plugin_register_smoke` false and reports that tool by name.
- Fixtures with `parameters` set to `null`, a list, or a non-object schema fail with stable rule identifiers and bounded messages.
- A schema whose internal `name` differs from the registered name fails before installation is reported ready.
- All currently bundled OMH plugin tools pass the same validator through the real bundled `register(ctx)` path.
- `omh setup --with-plugin` and `omh doctor` surface the schema failure and do not report `plugin_distribution_ready`.
- The test suite asserts behavior through imported definitions and registration calls rather than source-text matching.
- The result continues to say that local schema and registration smoke are not observed Hermes load or tool-use evidence.

## Boundaries and risks

This change is limited to OMH-owned plugin packaging, setup, doctor, and tests. It must not patch Hermes, call a model provider, add network access, or execute tool handlers. Validation should cover only the stable contract OMH depends on; over-validating optional JSON Schema features could reject definitions that Hermes accepts. Error output must remain bounded and must not include handler internals, prompts, credentials, or unrelated plugin data.

Som0111:
Fixes #1437

### What Changed

- `_SmokeContext.register_tool()` in `src/install/plugin_pack.py` now captures name, toolset, schema, handler, and description for every registered tool, instead of just the name.
- Added `validate_tool_definitions()`, which checks each captured tool against the minimum Hermes tool contract: mapping-shaped schema, schema `name` matches the registered name, non-empty description, `parameters` is an object-shaped mapping with `type: object`, and coherent `properties`/`required` when both are present.
- `_register_smoke()` now runs this validation and folds any failures into the `register_smoke` result instead of reporting success on name-presence alone.
- `inspect_plugin_bundle()` surfaces bounded per-tool schema failures (tool name + violated rule id only — no handler internals, prompts, or credentials), which flow through to `plugin_distribution_ready` for both `setup` and `doctor` via the existing shared code path (no duplicated validation logic).

### Why This Exists

- Previously, `register_tool()` accepted arbitrary positional/keyword arguments without inspecting them, so a tool with a missing or malformed `parameters` schema could pass registration smoke and only fail later when Hermes prepared an actual model request. This closes that release-regression blind spot at the setup/doctor gate, before installation is ever reported healthy.

### User / Operator Impact

- A malformed plugin tool schema now fails `setup --with-plugin` / `doctor` immediately, with a named tool and a stable rule id, instead of surfacing as a confusing runtime failure during a live Hermes session.

### How It Works

- Validation is deterministic and local: it inspects the schema shape captured at registration time. It does not execute any tool handler, make a model/provider request, or touch the network. It only proves conformance to OMH's own registration contract — it is explicitly not evidence that Hermes loaded or invoked the plugin.

- `src/install/plugin_pack.py` — `_SmokeContext.register_tool()`, `validate_tool_definitions()`, `_register_smoke()`, `inspect_plugin_bundle()`
- `tests/test_plugin_distribution.py` — new `PluginToolSchemaValidationTests`

### Affected Surfaces

- [ ] Hermes chat or wrapper
- [ ] Codex
- [ ] Claude Code
- [ ] Generic executor
- [x] Not executor-specific

## Validation

- [x] `git diff --check`
- [ ] `uv run --group lint ruff check src tests`
- [x] `PYTHONPATH=tests uv run python -m unittest discover -s tests -v` — 59/60 pass
- [ ] `omh --help`

### Observed Evidence

- `PluginToolSchemaValidationTests` exercises the real bundled `omh.plugin_bundle.omh.register(ctx)` path, covering: all 15 currently bundled tool schemas passing unchanged; missing/`null`/list/non-object `parameters` failing with stable rule ids; internal schema `name` mismatch failing before `plugin_distribution_ready`; setup/doctor surfacing the failure and withholding `plugin_distribution_ready`; and an evidence-boundary wording check confirming this smoke is not claimed as observed Hermes load or tool-use evidence.
- Full targeted test file (`test_plugin_distribution.py`) run: 59/60 pass. The 1 failure (`test_installed_plugin_status_tool_and_hook_keep_evidence_boundary`) is pre-existing and unrelated — it fails identically on unmodified `main` (confirmed via `git stash`); root cause is a missing `python3` executable on this machine's PATH inside a subprocess call, not this change.

- Lint, docs-check, harness, release-checklist, hermes-smoke, and live `omh` CLI commands were not run in this pass — need to be run before merge to fully satisfy the checklist.

## Risk

- Low: validation is additive, deterministic, local-only, and does not touch Hermes-owned code, network access, or tool handler execution. The only behavior change is that previously-undetected malformed schemas now fail smoke instead of passing silently.

## Compatibility / Rollout

- No breaking change for correctly-formed plugin tools — all 15 currently bundled OMH schemas pass validation unchanged. Only malformed/future schemas are newly rejected.

## Release / Claims

- [ ] Release-channel impact considered (`stable`, `preview`, or `local`)
- [ ] Known manual Hermes checks are listed, including any `omh release hermes-smoke --live` gap

## Follow-Up

- Run the remaining unchecked validation commands (lint, docs checks, harness validate, release checklist, hermes-smoke) before merge.
- Investigate/fix the pre-existing `python3` PATH issue causing the one unrelated test failure, separately from this PR.

### DIFF
--- a/src/install/plugin_pack.py
+++ b/src/install/plugin_pack.py
@@ -34,12 +34,82 @@ class _SmokeContext:
     def __init__(self) -> None:
         self.tools: list[str] = []
         self.hooks: list[str] = []
+        self.tool_definitions: list[dict[str, object]] = []
 
     def register_tool(self, name: str, *args: object, **kwargs: object) -> None:
         self.tools.append(name)
+        toolset = args[0] if len(args) > 0 else None
+        schema = args[1] if len(args) > 1 else None
+        handler = args[2] if len(args) > 2 else None
+        description = kwargs.get("description")
+        if description is None and isinstance(schema, dict):
+            description = schema.get("description")
+        self.tool_definitions.append(
+            {
+                "name": name,
+                "toolset": toolset,
+                "schema": schema,
+                "handler": handler,
+                "description": description,
+            }
+        )
 
     def register_hook(self, name: str, *args: object, **kwargs: object) -> None:
         self.hooks.append(name)
+
+
+# Stable rule identifiers for `validate_tool_definitions` failures. These are
+# part of the contract callers (doctor, setup, tests) key off of, so treat
+# renames as breaking.
+TOOL_SCHEMA_RULE_NOT_A_MAPPING = "tool_schema_not_a_mapping"
+TOOL_SCHEMA_RULE_NAME_MISMATCH = "tool_schema_name_mismatch"
+TOOL_SCHEMA_RULE_DESCRIPTION_MISSING = "tool_description_missing"
+TOOL_SCHEMA_RULE_PARAMETERS_MISSING = "tool_parameters_missing"
+TOOL_SCHEMA_RULE_PARAMETERS_NOT_A_MAPPING = "tool_parameters_not_a_mapping"
+TOOL_SCHEMA_RULE_PARAMETERS_TYPE_NOT_OBJECT = "tool_parameters_type_not_object"
+TOOL_SCHEMA_RULE_REQUIRED_NOT_A_LIST = "tool_parameters_required_not_a_list"
+TOOL_SCHEMA_RULE_REQUIRED_FIELD_UNKNOWN = "tool_parameters_required_field_unknown"
+
+
+def validate_tool_definitions(tool_definitions: list[dict[str, object]]) -> list[dict[str, str]]:
+    """Validate captured tool registrations against the minimum Hermes tool contract.
+
+    Schema-only: never touches ``handler`` and never calls it. Each failure
+    carries a stable ``rule`` id plus the ``tool`` name so callers can surface
+    a bounded message without leaking handler internals or unrelated plugin
+    data.
+    """
+    failures: list[dict[str, str]] = []
+    for definition in tool_definitions:
+        name = str(definition.get("name", ""))
+        schema = definition.get("schema")
+        if not isinstance(schema, dict):
+            failures.append({"tool": name, "rule": TOOL_SCHEMA_RULE_NOT_A_MAPPING})
+            continue
+        if schema.get("name") != name:
+            failures.append({"tool": name, "rule": TOOL_SCHEMA_RULE_NAME_MISMATCH})
+        description = definition.get("description")
+        if not str(description or "").strip():
+            failures.append({"tool": name, "rule": TOOL_SCHEMA_RULE_DESCRIPTION_MISSING})
+        if "parameters" not in schema:
+            failures.append({"tool": name, "rule": TOOL_SCHEMA_RULE_PARAMETERS_MISSING})
+            continue
+        parameters = schema["parameters"]
+        if not isinstance(parameters, dict):
+            failures.append({"tool": name, "rule": TOOL_SCHEMA_RULE_PARAMETERS_NOT_A_MAPPING})
+            continue
+        if parameters.get("type") != "object":
+            failures.append({"tool": name, "rule": TOOL_SCHEMA_RULE_PARAMETERS_TYPE_NOT_OBJECT})
+        required = parameters.get("required")
+        if required is not None:
+            if not isinstance(required, (list, tuple)):
+                failures.append({"tool": name, "rule": TOOL_SCHEMA_RULE_REQUIRED_NOT_A_LIST})
+            else:
+                properties = parameters.get("properties")
+                known = set(properties) if isinstance(properties, dict) else set()
+                if any(field not in known for field in required):
+                    failures.append({"tool": name, "rule": TOOL_SCHEMA_RULE_REQUIRED_FIELD_UNKNOWN})
+    return failures
 
 
 def install_plugin_bundle(paths: OmhPaths, *, force: bool = False, dry_run: bool = False) -> dict[str, Any]:
@@ -73,6 +143,7 @@ def install_plugin_bundle(paths: OmhPaths, *, force: bool = False, dry_run: bool
             "register_smoke": smoke["plugin_register_smoke"],
             "registered_tools": smoke["registered_tools"],
             "registered_hooks": smoke["registered_hooks"],
+            "tool_schema_failures": smoke["tool_schema_failures"],
         }
     )
     return result
@@ -111,6 +182,7 @@ def inspect_plugin_bundle(paths: OmhPaths) -> dict[str, Any]:
         errors.append(str(smoke["error"]))
     missing_tools = [str(item) for item in smoke.get("missing_registered_tools", [])]
     missing_hooks = [str(item) for item in smoke.get("missing_registered_hooks", [])]
+    schema_failures = [item for item in smoke.get("tool_schema_failures", []) if isinstance(item, dict)]
     if target.exists() and import_smoke and not register_smoke and (missing_tools or missing_hooks):
         detail = []
         if missing_tools:
@@ -118,6 +190,11 @@ def inspect_plugin_bundle(paths: OmhPaths) -> dict[str, Any]:
         if missing_hooks:
             detail.append(f"missing hooks={missing_hooks}")
         errors.append("plugin register smoke is incomplete: " + "; ".join(detail))
+    if target.exists() and import_smoke and schema_failures:
+        errors.append(
+            "plugin tool schema validation failed: "
+            + "; ".join(f"{item.get('tool', '')}={item.get('rule', '')}" for item in schema_failures)
+        )
     return {
         "schema_version": PLUGIN_SCHEMA_VERSION,
         "plugin_name": PLUGIN_NAME,
@@ -136,6 +213,7 @@ def inspect_plugin_bundle(paths: OmhPaths) -> dict[str, Any]:
         "registered_hooks": smoke.get("registered_hooks", []),
         "missing_registered_tools": missing_tools,
         "missing_registered_hooks": missing_hooks,
+        "tool_schema_failures": schema_failures,
         "plugin_distribution_ready": bool(
             target.exists()
             and manifest_current
@@ -359,13 +437,15 @@ def _register_smoke(plugin_dir: Path) -> dict[str, Any]:
         required_hooks = set(REQUIRED_HOOKS)
         missing_tools = sorted(required_tools.difference(ctx.tools))
         missing_hooks = sorted(required_hooks.difference(ctx.hooks))
+        schema_failures = validate_tool_definitions(ctx.tool_definitions)
         return {
             "import_smoke": True,
-            "register_smoke": not missing_tools and not missing_hooks,
+            "register_smoke": not missing_tools and not missing_hooks and not schema_failures,
             "registered_tools": sorted(ctx.tools),
             "registered_hooks": sorted(ctx.hooks),
             "missing_registered_tools": missing_tools,
             "missing_registered_hooks": missing_hooks,
+            "tool_schema_failures": schema_failures,
         }
     except Exception as exc:
         return {"import_smoke": False, "register_smoke": False, "error": f"plugin smoke failed: {exc}"}
--- a/tests/test_plugin_distribution.py
+++ b/tests/test_plugin_distribution.py
@@ -23,6 +23,7 @@ from omh.commands import setup as _setup_module
 from omh.paths import resolve_paths
 from omh.install.plugin_loader_observation import observe_real_loader_registration
 from omh.plugin_pack import inspect_plugin_bundle
+from omh.install.plugin_pack import _SmokeContext, validate_tool_definitions
 from omh.plugin_bundle.omh.tools import evidence_tool
 from omh.plugin_bundle.omh.metadata import PROVIDED_HOOKS, PROVIDED_TOOLS, TOOL_FILE_STEMS
 from omh.release_smoke_core import CommandResult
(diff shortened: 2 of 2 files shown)


src/install/plugin_pack.py
```diff
@@ -34,14 +34,84 @@ class _SmokeContext:
     def __init__(self) -> None:
         self.tools: list[str] = []
         self.hooks: list[str] = []
+        self.tool_definitions: list[dict[str, object]] = []
 
     def register_tool(self, name: str, *args: object, **kwargs: object) -> None:
         self.tools.append(name)
+        toolset = args[0] if len(args) > 0 else None
+        schema = args[1] if len(args) > 1 else None
```
sionic-khope, line 42: Nit (two, folded together since they sit on the same three lines, 41-43).

**Dead fields.** `toolset` and `handler` are stored into `tool_definitions` and never read. `validate_tool_definitions` reads only `name`, `schema` and `description`. Its docstring says it "never touches `handler`", which is true and is exactly why capturing it earns nothing. Dropping both keeps the captured record equal to what the validator actually validates.

**Positional-only extraction fails closed.** Both live call sites are positional today, `src/plugin_bundle/omh/__init__.py:108` and `src/plugin_bundle/omh/egress_attempts.py:251`, so this is correct as written. But if a call site ever moves `schema` to a keyword, `args[1]` is absent, `schema` becomes `None`, and validation raises `tool_schema_not_a_mapping` for a tool that is perfectly well formed. `register_smoke` then goes false and `plugin_distribution_ready` is withheld, so `omh setup` and `omh doctor` report a healthy install as broken. A keyword fallback costs one expression and removes the trap.

I did not reproduce this: no current call site triggers it. It is a latent shape assumption, which is why it is a nit and not a finding.

The suggestion below covers only the second point, because dropping `toolset` and `handler` also means removing them from the dict literal just below, and a suggestion block cannot reach that far.

```suggestion
        schema = args[1] if len(args) > 1 else kwargs.get("schema")
```

src/install/plugin_pack.py
```diff
@@ -34,14 +34,84 @@ class _SmokeContext:
...
     def register_hook(self, name: str, *args: object, **kwargs: object) -> None:
         self.hooks.append(name)
 
 
+# Stable rule identifiers for `validate_tool_definitions` failures. These are
+# part of the contract callers (doctor, setup, tests) key off of, so treat
+# renames as breaking.
+TOOL_SCHEMA_RULE_NOT_A_MAPPING = "tool_schema_not_a_mapping"
+TOOL_SCHEMA_RULE_NAME_MISMATCH = "tool_schema_name_mismatch"
+TOOL_SCHEMA_RULE_DESCRIPTION_MISSING = "tool_description_missing"
```
sionic-khope, line 66: Nit. The comment above this block calls these ids part of the contract and says renames are breaking, so the naming is worth settling now rather than after something keys off it. Seven of the eight ids carry a `tool_schema_` or `tool_parameters_` prefix; this one is bare `tool_description_missing`. Since the description is read off the registration kwargs rather than the schema body, either name is defensible, but the odd one out should be the deliberate choice, not the accident.

```suggestion
TOOL_SCHEMA_RULE_DESCRIPTION_MISSING = "tool_schema_description_missing"
```

src/install/plugin_pack.py
```diff
@@ -34,14 +34,84 @@ class _SmokeContext:
...
+        if parameters.get("type") != "object":
+            failures.append({"tool": name, "rule": TOOL_SCHEMA_RULE_PARAMETERS_TYPE_NOT_OBJECT})
+        required = parameters.get("required")
+        if required is not None:
+            if not isinstance(required, (list, tuple)):
+                failures.append({"tool": name, "rule": TOOL_SCHEMA_RULE_REQUIRED_NOT_A_LIST})
+            else:
+                properties = parameters.get("properties")
+                known = set(properties) if isinstance(properties, dict) else set()
+                if any(field not in known for field in required):
```
sionic-khope, line 110: Nit, and it is the PR body that is wrong rather than this code.

The body says the check covers "coherent `properties`/`required` when both are present". This branch runs whenever `required` is present, whatever `properties` does: with `properties` absent, `known` is the empty set and every entry in `required` is reported unknown.

That looks like the behavior you want. A schema that requires a field it never declares is malformed whether or not `properties` exists at all, and catching it is the point of the rule. Just reword the body so the sentence matches, since the rule ids are advertised as a stable contract and this one's trigger condition is part of it.

tests/test_plugin_distribution.py
```diff
@@ -23,6 +23,7 @@
 from omh.paths import resolve_paths
 from omh.install.plugin_loader_observation import observe_real_loader_registration
 from omh.plugin_pack import inspect_plugin_bundle
+from omh.install.plugin_pack import _SmokeContext, validate_tool_definitions
```
sionic-khope, line 26: Nit. Line 25 imports `inspect_plugin_bundle` through the public shim `omh.plugin_pack`, which is `from .install.plugin_pack import *` at `src/omh/plugin_pack.py:3`; this line reaches into `omh.install.plugin_pack` for the same module.

Nothing is broken. The star re-export carries `validate_tool_definitions` and every `TOOL_SCHEMA_RULE_*` constant with it, and the `mock.patch("omh.install.plugin_pack.validate_tool_definitions")` in the setup/doctor test does have to name the defining module to bind the reference `_register_smoke` calls. But `_SmokeContext` is private and `validate_tool_definitions` is public, and importing them together through the private path makes two adjacent lines disagree about which name this module has.

```suggestion
from omh.plugin_pack import validate_tool_definitions
from omh.install.plugin_pack import _SmokeContext
```

sionic-khope (commented):
**Round 1.** `gh api repos/rlaope/oh-my-hermes/pulls/1439/reviews` returns `[]`, so this is the first review on this PR. The `REVIEW.md` re-review ratchet does not apply and I reviewed the whole diff.

**2 important, 5 nits.**

Neither Important finding is in the code. The change itself is small and well shaped: `_SmokeContext.register_tool` now records the schema next to the name, `validate_tool_definitions()` checks it against eight named rules, and `_register_smoke` folds any failure into `register_smoke` so `plugin_distribution_ready` is withheld. The positional argument order it assumes (`name, toolset, schema, handler`) matches both live call sites, `src/plugin_bundle/omh/__init__.py:108` and `src/plugin_bundle/omh/egress_attempts.py:251`, and the existing test harness `FakeHermesContext` already indexes `args[2]` for the handler at `tests/test_plugin_distribution.py:814`. The two Important findings are about the commit trailers and one checked box in the PR body.

## Gates

The PR leaves seven validation boxes unchecked and says under Not Tested that they "need to be run before merge". I ran all of them, in a clean private worktree at this head.

| Gate | Result |
| --- | --- |
| `unittest discover -s tests` | `Ran 10486 tests in 782.402s` / `OK (skipped=1)` |
| `unittest tests/test_plugin_distribution.py` | `Ran 59 tests in 33.401s` / `OK` |
| `ruff check src tests` | `All checks passed!` |
| `compileall -q src tests` | exit 0 |
| `docs workflows --check` | `{"checked":".../docs/WORKFLOWS.md","ok":true,"tap_skills":{"checked":206,"expected":206,...,"ok":true,...}}` |
| `docs roles --check` | `{"checked":".../docs/ROLES.md","ok":true}` |
| `docs capability-families --check` | `{"checked":".../src/plugin_bundle/omh/tools/capability_families.json","ok":true}` |
| `git diff --check` | exit 0, no output |
| `harness validate` | `{"counts":{"harnesses":77,"skills":128},"errors":[],"ok":true,...}` |
| `release checklist --json` | `"ok":true`, `"required_item_count":36`, `"tag":"v2.0.2"` |
| `release hermes-smoke` | exit 0, plan mode, `Status: ok` |
| `omh --help` | exit 0 |
| `release hermes-smoke --install-path setup --omh-command omh --include-command-smoke --json` | exit 0 against an isolated `--omh-home` / `--hermes-home` |

Every gate is green. Nothing in this diff moves a generated artifact, a routing case, a skill, or a demo card, so no exact-count assertion is stale.

I also ran the real end-to-end path the change gates, against isolated homes:

```
omh setup --with-plugin   ->  [ok] ready   (plugins/omh)
omh doctor                ->  exit 0
                              "Plugin bridge: import/register smoke passed against OMH's fake context."
```

All 15 bundled schemas pass the new validation and a healthy install is still reported ready.

## The reported local failure does not reproduce here

The PR reports `test_installed_plugin_status_tool_and_hook_keep_evidence_boundary` failing locally, attributed to a missing `python3` on PATH inside a subprocess call.

I could not reproduce it. On this head, in a clean worktree, `test_plugin_distribution.py` runs 59 tests and all 59 pass, that test included, and the full 10486-test suite is `OK`.

The mechanism you describe is real, though. That test reaches `omh_gather_evidence` with `{"commands": ["python3 -m compileall -q ."]}` at `tests/test_plugin_distribution.py:990` and asserts `evidence["all_pass"]`, so the command genuinely shells out to `python3`; on a machine where `python3` is not on PATH it would fail exactly as you say. This machine has `python3` at `/opt/homebrew/bin/python3`. So: the attribution is plausible and environmental, and there is nothing here for you to fix, but the arithmetic does not line up either way. The file holds 59 tests on this branch, not 60, so "59/60" cannot describe a run of it.

## Overlap with PR #1430

#1430 (`fix(install): handle symlinked plugin dirs in _copy_plugin_bundle`) touches the same two files. I checked for a semantic collision and found none.

The two changes occupy disjoint functions. #1430 rewrites `_copy_plugin_bundle` and adds `_discard_path`, which is filesystem replacement; this PR touches `_SmokeContext`, `validate_tool_definitions`, `_register_smoke` and `inspect_plugin_bundle`, which is schema inspection. Neither reads the other's state, and neither one's behavior undercuts the other.

I verified rather than assumed. `git merge-tree --write-tree --messages` on the two heads reports `Auto-merging src/install/plugin_pack.py` and `Auto-merging tests/test_plugin_distribution.py` with no conflict, and running the merged tree gives:

```
Ran 60 tests in 29.250s

OK
```

59 from this PR plus the one #1430 adds. Reporting only; the merge order is the maintainer's call.

## Untrusted-input scan

This is a fork PR, so I scanned the title, body, commit message and diff for hidden or injected instructions. Clean: no HTML comments, no zero-width or bidi characters, no non-ASCII bytes anywhere in the diff, and no text directing a reviewer to approve, skip a check, or ignore `REVIEW.md`. I did not run anything the branch adds; every command above is the repository's own suite or CLI.

## Important 1 of 2 — the commit is missing every Lore-style trailer but the signoff

`AGENTS.md`, under Git And Commits, says every commit must include the DCO signoff **and** the local Lore-style trailers, with `Signed-off-by:` last: `Constraint`, `Rejected`, `Confidence`, `Scope-risk`, `Directive`, `Tested`, `Not-tested`. `REVIEW.md` lists this under Always check.

The single commit on this branch carries only the signoff:

```
$ git log origin/main..HEAD --format='%an <%ae>%n%b'
Som0111 <soumyaswarup07@gmail.com>
Signed-off-by: Som0111 <soumyaswarup07@gmail.com>
```

The signoff itself is correct and matches the author. An amend adds the rest; most of the content is already written in the PR body, and your Not Tested section is close to a ready `Not-tested:` line.

## Important 2 of 2 — the full-suite box is checked with a single-file result

The Validation section carries:

```
[x] PYTHONPATH=tests uv run python -m unittest discover -s tests -v - 59/60 pass
```

That box is the whole-suite gate. On this head it runs **10486** tests, not 60. Your own Observed Evidence section scopes the number correctly ("Full targeted test file (`test_plugin_distribution.py`) run"), and the Not Tested section does not list the suite, so the checked box is the only place claiming the suite ran. `REVIEW.md` treats evidence language claiming execution that did not happen as Important, which is why this is here rather than in the nits.

This is a labeling fix, not a work item. The suite does pass at this head, and I have quoted the run above. Either move the annotation to the targeted-file line and check the suite box on my evidence, or leave the suite box unchecked and say the targeted file is what you ran.

## Nits

1. `toolset` and `handler` are captured into `tool_definitions` at `src/install/plugin_pack.py:41` and `:43` but nothing ever reads them. `validate_tool_definitions` uses only `name`, `schema` and `description`. Inline.
2. The positional-only extraction at `src/install/plugin_pack.py:41-43` fails closed. Inline.
3. Rule-id prefix drift at `src/install/plugin_pack.py:66`. Inline.
4. The PR body says the required/properties check applies "when both are present", but `src/install/plugin_pack.py:108-110` also fails a schema that has `required` and no `properties` at all, because `known` becomes the empty set and every required field is then unknown. That is probably the behavior you want; the sentence in the body is what is wrong. Inline.
5. `tests/test_plugin_distribution.py:26` imports through `omh.install.plugin_pack` while line 25 imports the same module through the public `omh.plugin_pack` shim. Inline.

Plus 2 similar items:

- The egress surface is never schema-validated by this smoke. `src/plugin_bundle/omh/__init__.py:80-84` gates `register_egress_attempts` on `ctx.get_config`, and `_SmokeContext` does not define it, so those `register_tool` calls never run under smoke. Not a defect today, and worth knowing before anyone adds `get_config` to the smoke context: `src/plugin_bundle/omh/egress_attempts.py:251` forwards `original.schema` from Hermes' own tool registry, so OMH's contract would then be applied to schemas OMH does not author.
- `docs/INSTALLATION.md:1615-1617` describes what `omh doctor` checks on the plugin bundle. It now has a new failure mode and `inspect_plugin_bundle` a new `tool_schema_failures` field, neither mentioned there. No gate enforces this.

PR #1442, opened 2026-09-09
feat: native Kanban coordination, resumable executor sessions, capacity-aware fanout, bounded failure diagnostics and lifecycle launch controls

### LINKED ISSUE #1399: [Feature]: Make lifecycle-growth launch decisions audience-aware and exposure-safe
## Problem

`lifecycle-growth` can judge a lifecycle idea from aggregate results without proving that the intended audience is reachable, that exposure was assigned correctly, or that users were not repeatedly contacted across overlapping experiments. This can turn measurement noise or delivery failure into a rollout recommendation.

## Complete user goal

A growth operator should receive a launch decision that separates audience eligibility, exposure assignment, actual reach, repeated-contact pressure, and outcome validity before recommending expansion.

## Intended surfaces

- `lifecycle-growth` experiment and launch review
- Audience and exposure evidence contracts
- Rollout, hold, and stop dispositions

## Observed evidence

The current workflow covers data health, significance, guardrails, and rollout decisions, but it does not require a complete audience-reachability and exposure-integrity record across channels and concurrent lifecycle actions.

## Proposal

Add a provider-neutral audience and exposure contract. Record eligibility and exclusion reasons, assignment identity, attempted and observed reach, channel-specific failures, repeated-contact pressure, and overlap with other experiments. Block launch recommendations when these fields are unknown or materially inconsistent.

## Alternatives considered

- Rely on aggregate conversion metrics: rejected because unreachable or misassigned users can bias the result.
- Treat delivery as a separate operations concern: too narrow because delivery determines whether the experiment was actually experienced.
- Require one universal provider schema: rejected in favor of a normalized evidence contract.

## Success criteria

- The review distinguishes eligible, assigned, attempted, reached, and converted populations.
- Unknown audience or exposure fields produce a hold rather than a launch recommendation.
- Repeated-contact and experiment-overlap checks are explicit.
- Channel failures cannot silently appear as neutral outcomes.
- Fixtures cover sample mismatch, unreachable cohorts, duplicate exposure, and partial delivery.

## Boundaries and risks

Provider receipts and audience models vary, so confidence and missing evidence must remain explicit. The workflow must not claim delivery, exposure, or launch success from a prepared plan alone.

### LINKED ISSUE #1417: [Feature]: Connect agent-board workflows to native Hermes Kanban tools
## Problem

OMH routes multi-agent coordination requests to `agent-board`, but the workflow stops at a prepared card even when the Hermes host exposes its native durable Kanban toolset. Users can be told that a board, handoff lane, heartbeat policy, or blocker state has been prepared without receiving a capability-aware path that creates or updates the actual shared task queue.

This leaves a product gap between OMH's routing and Hermes execution surfaces: the documentation says native Kanban is the preferred mechanism for real multi-profile work distribution, while the generated skill and wrapper actions expose only prepare, refresh, and show behavior.

## Complete user goal

A user should be able to ask Hermes to coordinate persistent profiles on a multi-step objective and have OMH select the right native primitive, prepare safe task contracts, invoke the host's Kanban actions when those capabilities are available and authorized, and then report observed task, handoff, heartbeat, review, blocker, and completion state. The same request must degrade honestly to a prepared-only board when the host cannot expose or authorize those actions.

## Intended surfaces

- Hermes chat or wrapper
- Hermes runtime or handoff
- All executors (executor-neutral)

## Observed evidence

At local revision `87916cf8254f902c701ad9ee084a16618c82ebc0`:

- `skills/agent-board/SKILL.md` defines the expected output as an `agent-board/v1` card or guidance and says to prepare a runtime handoff only after a later accepted step. It does not map any native Kanban tool call or result into the workflow.
- `src/routing/recommend.py` assigns `agent-board` the single next action `prepare_agent_board_card`.
- `src/wrapper/contract.py` renders only `prepare_agent_board_card`, `refresh_status`, and `show_status` actions for this route. It has no create, comment, heartbeat, block, review, complete, link, or unblock action contract.
- `docs/MULTI_AGENT_OPERATIONS.md` identifies the Hermes Kanban toolset as the preferred target for actual multi-profile work distribution and explicitly distinguishes that from OMH narration.
- `docs/HERMES_AGENT_INTEGRATION_RUNBOOK.md` likewise says OMH should prepare handoffs onto native Kanban rather than implement another queue.
- `src/workflows/hermes_readiness_catalog.py` recognizes `kanban.db` and broad Kanban capability descriptions, but the routed `agent-board` response does not consume a live tool-capability observation to select an executable host action.

This is source-level evidence of an integration gap. No Kanban task creation, worker dispatch, review, or completion was executed or observed during this investigation.

## Proposal

1. Add a capability-aware `agent-board` action contract that distinguishes:
   - prepared-only board design,
   - native Kanban actions available but not yet authorized,
   - native Kanban mutation requested,
   - observed Kanban result,
   - unavailable or failed host action.
2. Map the workflow to the host's structured Kanban operations for task creation, listing/showing, dependency linking, comments, heartbeat, block/unblock, review requests and change requests, attachment references, and completion. The model-facing path should use native structured tools rather than shell commands.
3. Keep primitive selection explicit:
   - use Kanban for durable cross-profile work, restart survival, human intervention, or peer handoffs;
   - use Hermes delegation for bounded fork-and-join work whose result returns to the parent context;
   - use a prepared card only when neither execution surface is available or authorized.
4. Carry stable board, task, profile, parent, revision, and idempotency references in the handoff so retries cannot create duplicate tasks and read-back results cannot update the wrong board or task.
5. Map host results into evidence-bounded OMH states. A prepared action remains `prepared_not_observed`; task creation becomes observed only after an exact host result; heartbeat, review, blocker, and completion require their own read-back evidence.
6. Update the canonical catalog and wrapper sources, then regenerate the skill and workflow documentation rather than editing generated artifacts independently.

## Alternatives considered

- Keep `agent-board` permanently read-only and require users to operate Kanban separately. This preserves a narrow boundary but leaves the primary multi-agent route unable to reach the native mechanism OMH already recommends.
- Build an OMH-owned task queue. Rejected because Hermes already owns the durable queue, dispatcher, profiles, review lifecycle, and workspace management.
- Shell out to `hermes kanban` from model-driven workflows. Rejected because structured host tools provide better capability detection, typed arguments, remote-terminal portability, and result evidence.
- Treat every multi-agent request as delegation. Rejected because delegation does not provide the durable, peer-oriented, restart-safe, human-editable lifecycle required by persistent profile collaboration.

## Success criteria

- [ ] With a host fixture that exposes native Kanban tools, an authorized `agent-board` request produces typed create/link/comment/heartbeat/review/block/complete actions rather than only `prepare_agent_board_card`.
- [ ] With Kanban unavailable, the same route returns a prepared-only card, names the missing capability, and makes no execution claim.
- [ ] Durable cross-profile requests select Kanban while bounded parent-child research or review requests can select delegation under deterministic tests.
- [ ] Retrying an accepted task-creation action with the same idempotency reference returns the existing task instead of creating a duplicate.
- [ ] Observed host results bind to the exact board and task before OMH advances status; stale, foreign, malformed, or failed results cannot mark work active or complete.
- [ ] Task creation, worker dispatch, heartbeat, blocker resolution, review, and completion remain separate evidence states in chat cards and status output.
- [ ] Canonical catalog, wrapper actions, generated `agent-board` skill, workflow documentation, and examples remain synchronized through existing drift checks.
- [ ] Tests cover capability present/absent, authorization denied, mutation failure, retry, stale result, board mismatch, task lifecycle, and prepared-versus-observed claims.

## Boundaries and risks

OMH should adapt to Hermes Kanban rather than own its database, dispatcher, worker processes, worktree lifecycle, or task-state machine. The feature must not read or write `kanban.db` directly when a structured host action is available. It must not expose raw task prompts, attachments, transcripts, credentials, or private profile metadata in OMH artifacts.

Host versions and enabled toolsets vary, so capability detection must be version-aware and fail closed. Mutating actions still require normal authority checks. A successful prepared action is not task creation, a created task is not worker dispatch, a heartbeat is not progress proof, and a completed task is not review, CI, merge-readiness, or merge evidence.

### LINKED ISSUE #1421: [Feature]: Persist resumable executor session IDs for local fanout units
## Problem

OMH's local fanout dispatcher launches Codex and Claude Code as plain subprocesses, but it does not capture the executor's resumable session identifier. After a unit succeeds, fails, or needs follow-up, operators can see the unit and worktree but cannot reliably reopen the exact executor conversation that produced the result. This breaks continuity during review fixes, interrupted runs, and partial reruns.

## Complete user goal

Make every local fanout unit resume-capable when its selected executor exposes a stable session or thread identifier. Persist the observed identifier with the unit's runtime evidence, show a copyable executor-specific resume action, and preserve the association across status reads and partial reruns without treating a prepared resume action as resumed execution.

## Intended surfaces

- Codex
- Claude Code
- Hermes runtime or handoff

## Observed evidence

At local revision `87916cf8254f902c701ad9ee084a16618c82ebc0`:

- `src/coding/fanout_dispatch.py` defines `codex exec {prompt}` and `claude -p {prompt}` command templates without structured output flags.
- `_dispatch_unit` records the unit ID, executor, command, timestamps, exit code, and worktree, but does not parse or persist a session or thread identifier.
- The fanout journal and status projection therefore cannot associate a dispatched unit with an executor session.
- Other OMH runtime paths already model observed session references and copyable resume contracts, so the missing fanout association is a local integration gap rather than an absent product concept.

No executor process was launched during this review. The evidence above is repository inspection, not execution, review, CI, or merge evidence.

## Proposal

1. Add executor adapters that request supported structured output and extract stable identifiers, such as a Codex thread ID or Claude Code session ID, without persisting raw event streams.
2. Extend per-unit dispatch and result records with an optional typed executor-session reference, its observation state, and executor identity.
3. Project the reference into fanout status and completion summaries and generate an executor-specific, copyable resume contract only when an identifier was actually observed.
4. Keep unit, worktree, revision, and executor-session lineage bound together so a reference cannot be reused by another unit accidentally.
5. Preserve compatibility for generic executors and older CLI versions by recording an explicit `not_available` or `not_observed` resumability state rather than failing otherwise valid work.
6. Add a bounded migration path for existing fanout artifacts that do not contain a session reference.

## Alternatives considered

- Rely on the worktree alone. Rejected because the filesystem does not contain the executor conversation or its review context.
- Store complete JSONL output. Rejected because it increases privacy and retention risk and is unnecessary when a stable identifier is sufficient.
- Require operators to copy identifiers manually. Rejected because it is error-prone and cannot preserve deterministic unit-to-session lineage.
- Make missing session IDs fail every unit. Rejected because some executors or versions cannot expose one; resumability should degrade explicitly without converting successful work into failure.

## Success criteria

- A fixture for each supported executor emits a structured session identifier, and the completed fanout unit exposes the exact observed identifier with the correct executor, worktree, and unit ID.
- A copyable resume action is available only for executor/reference combinations OMH knows how to resume.
- A follow-up or partial rerun can select the stored unit and produce the correct resume contract without launching it implicitly.
- Malformed output, unsupported versions, and generic executors produce explicit `not_available` or `not_observed` state and never reuse a stale identifier.
- Raw prompts, transcripts, hidden reasoning, and complete event streams are not persisted in the fanout artifact.
- Existing fanout plans and status readers remain compatible with artifacts that lack the new optional fields.
- Focused fanout tests and the full repository test suite pass.

## Boundaries and risks

Session identifiers may be sensitive operational metadata and must follow existing redaction, retention, and artifact-permission rules. CLI event schemas differ by executor and version, so parsing must be adapter-specific and fail closed. A resume contract is prepared guidance, not evidence that resumed execution, verification, review, CI, or merge occurred. This issue does not add autonomous review-fix execution or persist full executor transcripts.

### LINKED ISSUE #1422: [Feature]: Make local fanout dispatch capacity-aware
## Problem

OMH's local fanout dispatcher treats every non-zero worker exit as an ordinary unit failure and submits all currently ready units before it can react to a runtime-capacity rejection. When an executor refuses a spawn because its active-agent or session limit is temporarily exhausted, OMH can start additional futile processes, mark unaffected work as failed, and block dependent units even though no implementation attempt occurred.

## Complete user goal

Make local fanout dispatch aware of temporary executor-capacity failures. Stop futile submissions, distinguish admission failure from implementation failure, preserve runnable work for a safe retry or resume, and expose a clear operator action without weakening concurrency bounds or claiming unobserved execution.

## Intended surfaces

- Codex
- Claude Code
- Hermes runtime or handoff
- Generic executor

## Observed evidence

At local revision `87916cf8254f902c701ad9ee084a16618c82ebc0`:

- `src/coding/fanout_dispatch.py` performs a one-time binary executable-readiness probe before dispatch but has no runtime admission state.
- `dispatch_fanout` submits every dependency-ready unit to a `ThreadPoolExecutor` wave immediately.
- `_dispatch_unit` maps all non-zero process exits to `failed` without a typed capacity or admission classification.
- After a failed result, downstream units are recorded as `blocked_by_dependency`; there is no capacity breaker, queued state, backoff policy, or safe sequential fallback.
- Existing tests cover success, ordinary failure, dependency blocking, worktree setup, timeouts, and verification, but no capacity-rejection scenario.

No executor process was launched during this review. The evidence above is repository inspection, not execution, review, CI, or merge evidence.

## Proposal

1. Add executor-specific classification for temporary admission and capacity failures, with a conservative unknown-error fallback to ordinary failure.
2. Introduce a dispatch-scoped capacity circuit breaker. Once tripped for an executor, stop submitting new units to that executor and cancel queued futures that have not started.
3. Record affected units as a distinct resumable state such as `blocked_by_capacity`, not `failed` or `blocked_by_dependency`.
4. Preserve unit/worktree lineage and expose a bounded retry or `--only-units` resume action after capacity changes. Do not retry indefinitely or poll blindly.
5. Where the executor exposes reliable capacity telemetry, allow the scheduler to lower effective concurrency for later waves; otherwise stop with an actionable HOLD state.
6. Keep real worker failures isolated: an implementation/test failure must not trip the capacity breaker, and a capacity failure must not erase unrelated completed evidence.

## Alternatives considered

- Lower the default concurrency globally. Rejected because fixed conservative limits waste capacity and do not handle limits imposed outside OMH.
- Retry every non-zero exit. Rejected because deterministic implementation failures would loop and consume resources.
- Treat capacity rejection as an ordinary failure. Rejected because it misstates whether the unit executed and unnecessarily blocks recovery.
- Implement unbounded dynamic polling. Rejected because it can hang scheduled jobs and obscure the stop condition.

## Success criteria

- A deterministic fixture with executor capacity below requested fanout proves that the first classified admission rejection trips the breaker and no not-yet-started unit launches afterward.
- Capacity-blocked units retain their original plan, dependency, executor, and worktree lineage and can be selected for a later bounded resume.
- Completed units remain completed, genuinely failed units remain failed, and dependents distinguish failed prerequisites from capacity-blocked prerequisites.
- Ordinary non-zero exits, timeouts, missing executables, and verification failures do not trip the capacity breaker unless an executor adapter positively classifies them as admission failures.
- The run summary exposes requested concurrency, effective concurrency, breaker state, affected unit IDs, and the next safe operator action without persisting raw prompts or credentials.
- Tests cover mixed executors, concurrent in-flight units, cancellation races, malformed errors, retry exhaustion, and compatibility with existing fanout artifacts.
- Focused fanout tests and the full repository test suite pass.

## Boundaries and risks

Capacity error text and exit behavior vary across executors and versions, so classifiers must be narrowly scoped and version-tested. Futures already running when the breaker trips may finish and must be reconciled honestly. Automatic retries need strict bounds and should never turn a prepared resume action into an execution claim. This issue does not change provider quotas or create additional executor capacity.

### LINKED ISSUE #1423: [Feature]: Preserve bounded failure diagnostics for local fanout workers
## Problem

OMH's local fanout dispatcher captures both subprocess streams in memory but records only a short tail of standard output in the worker-result journal. Many executors and shell tools emit actionable failure diagnostics exclusively to standard error. In those cases, the unit is marked failed while the persisted summary can be empty, forcing operators to reproduce the failure before they can understand it.

## Complete user goal

Preserve enough bounded, sanitized failure diagnostics for every local fanout unit to explain why it stopped, regardless of which subprocess stream carried the error. Keep the diagnostic associated with the exact unit and phase while avoiding raw transcript retention, secret leakage, or unbounded artifacts.

## Intended surfaces

- Codex
- Claude Code
- Hermes runtime or handoff
- Generic executor

## Observed evidence

At local revision `87916cf8254f902c701ad9ee084a16618c82ebc0`:

- `_dispatch_unit` in `src/coding/fanout_dispatch.py` uses `capture_output=True`, so standard output and standard error are available after process completion.
- The `worker_result` journal summary is derived only from `(result.stdout or "")[-300:]`.
- Standard error is not projected into the returned unit record, journal summary, completion summary, or a referenced diagnostic artifact.
- Tests in `tests/test_fanout_dispatch.py` assert exit status and selected standard-output summaries but do not cover an executor that fails with diagnostics only on standard error.
- Worktree, launch, timeout, and verification exceptions are represented separately, so subprocess diagnostic preservation can be added without changing those failure categories.

No executor process was launched during this review. The evidence above is repository inspection, not execution, review, CI, or merge evidence.

## Proposal

1. Add a typed, bounded fanout failure-diagnostic projection that considers standard error and standard output with explicit stream provenance.
2. Prefer a concise sanitized diagnostic in status and journal records; when more retained detail is justified, store it in a permission-bounded artifact and reference it instead of embedding raw output everywhere.
3. Apply deterministic byte and line limits before persistence and include truncation metadata, stream name, phase, exit code, and a content digest.
4. Run the existing redaction policy over retained diagnostics and refuse persistence when safe sanitization cannot be established.
5. Preserve diagnostics for worker-process failures separately from worktree setup, timeout, verification, and dispatcher exceptions.
6. Keep successful-output summaries compatible and avoid changing a failed unit into a successful one merely because diagnostics were captured.

## Alternatives considered

- Persist standard error in full. Rejected because output can contain secrets, source fragments, or very large traces.
- Continue recording only standard output. Rejected because it loses the conventional error channel used by many CLIs.
- Merge both streams before process completion. Rejected because it destroys stream provenance and can complicate structured executor output parsing.
- Require operators to rerun with shell redirection. Rejected because the original environment and failure may not be reproducible.

## Success criteria

- A fixture that exits non-zero and writes only to standard error produces a non-empty, correctly attributed diagnostic in the fanout result.
- Mixed standard-output/standard-error failures preserve deterministic ordering metadata or separate bounded fields without conflating the streams.
- Oversized output is truncated at declared limits with original-size and truncation indicators; raw output is never copied unbounded into the journal.
- Representative secrets and control characters are redacted or rejected according to policy, and successful runs do not gain unnecessary raw-output retention.
- Timeout, worktree, verification, and dispatcher failures keep their existing typed reason while exposing only diagnostics relevant to that phase.
- Older fanout artifacts remain readable, and tests cover empty streams, non-UTF-8 replacement behavior, truncation, redaction, and malformed structured output.
- Focused fanout tests and the full repository test suite pass.

## Boundaries and risks

Diagnostic text is untrusted and may contain credentials, prompts, filesystem paths, source content, terminal control sequences, or misleading instructions. Retention must be minimal, sanitized, permission-bounded, and rendered as data rather than executable guidance. A better error record is debugging evidence only; it is not implementation, verification, review, CI, merge-readiness, or merge evidence.

sionic-khope:
### What Changed

- **Native Hermes Kanban coordination (#1417).** `omh_agent_board` now prepares typed native Kanban actions (create/link/comment/heartbeat/block/unblock/request-review/request-changes/complete plus show/list/attachment references) and projects exact observed identities back from the normal Hermes tool loop. The installed plugin's pre/post hooks correlate each native call by host session/task identity and argument digest; results are metadata-only receipts with `prepared` / `observed` / `denied` / `failed` / `unavailable` kept distinct. OMH-origin creates are serialized per board with an idempotency reference; stale, foreign, replayed and malformed results cannot advance task state.
- **Lifecycle-growth safety and upstream contracts (#1399).** Public audience, promotion and graduation operations remain provider-neutral. A separate exposure-evidence companion now binds eligible, assigned, attempted, actually reached and converted populations to observed audience, contact-pressure, overlap and channel checks. Missing or inconsistent observations hold expansion; channel failures remain visible, and an independently justified rollback is preserved. The six historical artifact schemas remain readable, but incomplete legacy inputs no longer produce a new safe-to-ship decision.
- **Resumable executor sessions (#1421).** Local fanout negotiates structured output per executor binary/version, feeds stdout native events to a fresh per-attempt session decoder, and persists validated `fanout_executor_session/v1` receipts (executor, protocol, reference, unit, attempt, worktree, revision, contract digest) into unit results, journal and status. `omh coding fanout status --fanout-id FANOUT_ID --unit UNIT_ID --json` and the wrapper follow-up projection return copy-only resume commands with cwd and required stdin; nothing is ever launched implicitly. Stale, duplicate, foreign and unsupported associations are refused; legacy artifacts read compatibly.
- **Capacity-aware fanout (#1422).** Every actual launch crosses the existing per-owner atomic gate. Default CLI dispatch now recognizes the exact official Codex 0.154.0 Darwin ARM64 and Linux ARM64/musl builds documented in `docs/FANOUT-CODEX-BUILD-PROVENANCE.md`; unknown builds remain ordinary failures. A source-qualified fresh initial-request rejection closes the owner breaker before subsequent launches while preserving in-flight work, dependency reasons and explicit bounded reselection. Definition and release-source revisions remain distinct; no global quota or native saturation is inferred.
- **Bounded failure diagnostics (#1423).** Worker, worktree, launch, timeout, verification and dispatcher failures persist a sanitized `failure_diagnostic` with exact unit/phase/stream/exit provenance, declared byte/line limits, original sizes, truncation state and digest. Streams are captured incrementally in bounded binary chunks; secrets and prompt/source sentinels are screened before truncation; raw sidecars live only in invocation-owned ephemeral intake. Journal, status, brief and status-board project the current attempt's diagnostic without borrowing stale failures.

### Why This Exists

Five open issues described real operator gaps: prepared-only Kanban cards that never touched the native board, lifecycle decisions missing assigned-population and observed contact/reach evidence, fanout units whose executor sessions were lost after the process exited, futile launches after an executor refused admission, and failed workers whose stderr vanished or spilled raw. Each fix builds on existing foundations (adaptive admission, retries, wrappers, safe evidence helpers) rather than adding a parallel runtime.

### User / Operator Impact

- Hermes can coordinate durable work through its own authorized Kanban tools and report the exact landed lifecycle state.
- A failed fanout unit can be resumed in the original executor conversation from a copyable command instead of starting over.
- For a supported source-qualified rejection, remaining units are capacity-blocked with their lineage; unrelated executors continue.
- Failure output is readable, bounded and safe in `status` / `brief` output.
- Lifecycle-growth launch decisions now carry explicit reachability, promotion and graduation states with provider-neutral vocabulary.

### How It Works

See per-component notes above. Key boundaries: no SDK dispatch bypass, no JSON grants, no direct Kanban DB access, no OMH-owned queue, no raw prompt/transcript/event-stream retention, no implicit resume, no unbounded retry, no Hermes core patch.

- `src/workflows/agent_board.py`, `src/plugin_bundle/omh/{agent_board_bridge.py,tools/agent_board_tool.py,hooks/tool_hooks.py,__init__.py,metadata.py,plugin.yaml}`
- `src/workflows/{lifecycle_growth_launch.py,lifecycle_growth_exposure.py,lifecycle_growth_contracts.py,workflow_artifact_operations.py,observation_journal.py}`
- `src/coding/{fanout_dispatch.py,fanout_output.py,fanout_failure_diagnostics.py,fanout_executor_sessions.py,fanout_capacity.py,fanout_admission.py,fanout_retry.py,fanout_journal.py,fanout_status.py,status_board.py,worktree_creator.py,executor_readiness.py,unit_telemetry.py,fanout_unit_results.py,fanout_confinement.py,dispatch_failure_recovery.py,hermes_child_dispatch.py}`
- `src/commands/coding.py`, `src/wrapper/executor_sessions.py`, `src/runtime/artifacts.py`
- Wrapper/routing/catalog registration and generated projections (see final diff)
- Tests: component regression modules plus `tests/five_issue_cases/*` real-process case producers and `tests/five_issue_process_fixture.py`

### Affected Surfaces

- [x] Hermes chat or wrapper
- [x] Codex
- [x] Claude Code
- [x] Generic executor
- [ ] Not executor-specific

## Validation

Verified head: `99408c8d`. [CI 34447438085](https://github.com/rlaope/oh-my-hermes/actions/runs/34447438085) passed all 13 jobs, including both Windows shards and aggregate reconciliation; all 14 PR checks, including DCO, are successful.

- [x] `uv run --group lint ruff check src tests` (ran over `src tests tools packaging`)
- [x] `PYTHONPATH=tests uv run python -m unittest discover -s tests -q` - **10829 local unittest executions, OK, exit 0**, at `99408c8d` (2 explicit platform/capability skips; not a count of independent scenarios).
- [x] `uv run python -m omh.cli docs ulw-inventory --check` and `docs ulw-site --check`
- [x] `uv run python -m omh.cli docs claims --check --json`
- [x] `uv run python -m omh.cli release drift --json` - `drift_count: 0`
- [x] `git diff --check`
- [x] `omh harness validate`, `release checklist --json`, `release hermes-smoke`, `omh --help`, and isolated installed-command smoke (help, skill content, setup-plan). Plan-only Hermes smoke does not prove live Hermes execution.
- [ ] Manual Hermes/TUI check - not run; the native Kanban path was exercised through an isolated Hermes host in QA instead (see Observed Evidence).

### Observed Evidence

- Final wheel/sdist builds, generated checks, Ruff, compilation, release drift and installed-command smoke passed. Native Windows diagnosis confirmed missing bootstrap variables and text-mode CRLF fixture differences; the final repairs preserve runtime isolation, exact byte contracts and existing safety checks.

- All 36 canonical criterion surfaces passed again on clean tree `9ce4bcf285457d2e9b31468e67ec568559846e40` at `99408c8d`, with successful cleanup and independently checked path absence. K1/K4/K6/K8 use a real isolated Hermes host; K7/L1-L7 are local contract/source/generator checks; remaining surfaces use real process fixtures. Windows CI additionally passed the actual S1/C2/D5 smoke and the full test shards. No live chat, provider inference, native saturation or positive review dispatch is claimed.
- Per-node RED/GREEN, mutation and diagnostic evidence under `.omc/artifacts/five-issues/phases/C/<node>/`.

- Live provider execution of Codex/Claude capacity rejection: classification is source-qualified and fixture-observed; `native_positive_admission_executed=false` is recorded.

## Risk

- Shared dispatcher (`fanout_dispatch.py`) carries inherited typing debt; new code paths are covered by tests but strict whole-file diagnostics remain non-zero as before.

## Compatibility / Rollout

- Old artifacts, legacy string runners and generic executors remain readable. Lifecycle callers need the new companion evidence before receiving a fresh expansion recommendation; unsafe legacy decision equality is deliberately not preserved. Capacity detection supports only the explicitly documented executable builds.

## Release / Claims

- [x] Known manual Hermes checks are listed, including the unrun live-profile smoke.

## Follow-Up

- Typing debt in `fanout_dispatch.py` / `test_fanout_dispatch.py` (pre-existing).

Fixes #1399
Fixes #1417
Fixes #1421
Fixes #1422
Fixes #1423


sionic-khope (commented):
**6 important, 5 nits (plus 7 similar items).** Round one; no prior review exists on this PR.

Reviewed at head `a88a8526`, by four reviewers splitting the diff by dimension: generated artifacts and counts, the fanout execution boundary, the evidence contracts, and gates and claims. Each worked in its own worktree and ran the gates rather than reading the patch. This is a 13k-line, 71-file, 14-commit change spanning five issues, so the coverage list at the bottom says what we did and did not establish.

The shape of the work is strong. The five result states are genuinely distinct, the receipts are metadata-only by construction rather than by redaction, the failure-diagnostic allowlist screens secrets across chunk boundaries before truncation, the byte-identical compatibility promise holds, no closed-key `/vN` record gained a required field, and executor-neutrality is clean throughout the capacity vocabulary. Where the native Kanban lane is unavailable it blocks honestly instead of passing, and that behaviour is pinned by a test. The findings below are about a red build, one inverted evidence claim, one unreachable feature, and citation and trailer accuracy.

**Credit where it is due first.** Commit `a88a8526` independently found and fixed the defect our execution reviewer had reproduced: pre-spawn failures wrote a `worker_result` with no preceding dispatch event, and the journal's `ValueError` aborted the whole fanout. We re-verified at this head. A missing `codex` now returns `status: failed, exit_code: 127` with the journal in order, matching `origin/main`. All three pre-spawn branches are covered (`src/coding/fanout_dispatch.py:4140`, `:4153`, `:4158`), `TimeoutExpired` is correctly excluded, and the per-attempt idempotency really is exercised by `test_d6_retry_uses_fresh_capture_attempt_and_clears_live_old_failure`. That finding is **resolved**. Worth recording that the defect was reachable on POSIX too, not only on Windows — we reproduced it on macOS at the previous head. POSIX CI stayed green only because six test files use the production runner and the rest inject a runner that takes a different branch.

---

## Important

### 1. CI is red at this head, and the LF pin is incomplete

`test-windows (1)`: `Ran 5345 tests`, `FAILED (failures=11, skipped=134)`. `test-windows (0)` cancelled by matrix fail-fast, `aggregate` failed. Run 34378165298.

The ordering errors are gone — `errors=0`, down from 15 — so the fix landed. Eleven failures remain, in four groups:

| Cause | Count | Tests |
| --- | --- | --- |
| `dispatcher receipt missing` | 5 | `test_fanout_executor_sessions` S1, S3, S5, S7; `test_wrapper_contract` S2 |
| `A did not cross Popen` | 3 | `test_fanout_capacity` C2, C5, C7 |
| CRLF in captured subprocess bytes | 2 | C6 `(0, b'ready\r\n', b'')` vs `b'ready\n'`; S4 `(b'1.0\r\n', 'observed')` vs `b'1.0\n'` |
| Generated-artifact byte mismatch | 1 | K7: `['src\\plugin_bundle\\omh\\tools\\capability_families.json'] != []` |

The last one is the tractable one and it points at an incomplete fix. This head added a `.gitattributes` LF pin because git's Windows checkout rewrote generated files to CRLF and broke the K7 byte characterization. The pin covers the generated skill bodies and two generated docs. It does not cover `src/plugin_bundle/omh/tools/capability_families.json`, which K7 also compares byte-for-byte, so K7 still fails on the same mechanism the pin was written to close.

The two CRLF-in-subprocess failures are a different thing: those compare raw bytes captured from a real Windows child process, so a `.gitattributes` pin cannot help them. They need the assertion to normalise line endings or the capture to declare its newline handling.

Every failing test file is new in this PR and `main`'s Windows shards are green, so this is the PR's own regression, not inherited.

Two process points. `test-windows (0)` was cancelled by fail-fast (`ci.yml:156` has no `fail-fast: false`), and at the previous head it was shard 1 that was cancelled — so across both runs no single run has measured both shards, and the true blast radius is still unknown. And this is the **second consecutive commit** whose `Not-tested:` trailer reads "the Windows CI shards on this head are the proof". Both times the proof came back negative. A trailer that stakes its evidence on a pending run is a promise, not evidence; consider running the shards before writing the claim.

### 2. An observed rollback is rewritten to `insufficient_data`

`src/workflows/lifecycle_growth_contracts.py:135` spreads `lifecycle_growth_evaluation_context(...)` as the last entry of the returned dict, so its `disposition` and `interpretation_state` overwrite the ones computed above. The overlay sets them whenever any reason code fires (`src/workflows/lifecycle_growth_launch.py:296-298`), and `derive_readout_disposition` returns `rollback` for an observed guardrail breach or a triggered rollback (`src/workflows/lifecycle_growth_readout.py:88-89`). That is exactly what gets overwritten.

The module's own docstring at `src/workflows/lifecycle_growth_launch.py:277` states the opposite: "Resolved/observed context never overrides the existing data-health, runtime, approval, or rollback decisions."

Reproduced through the public API, on the fixture the PR's own tests use:

```
base disposition        : rollback | HOLD
with-context disposition: insufficient_data | HOLD
reason codes            : ['baseline_exposure_absent'] blocked= False
```

An observed guardrail breach reads as "not enough data yet", which is the softer claim, and softening an observed finding into a prepared-sounding one is the boundary this product exists to hold.

The existing guard `test_l6_context_overlay_never_upgrades_existing_hold` (`tests/test_lifecycle_growth_upstream.py:331-341`) passes only a clean context, so the overlay returns no disposition key and the case never runs. It can be extended in place rather than replaced: it already iterates the `{"guardrail_state": "failed"}` case that produces `rollback`, so parametrizing the inner context over a clean one and a reason-bearing one fails today. Calling `evaluate_lifecycle_growth(..., evaluation_context=...)` is the stronger form than composing `base | overlay(...)` by hand.

Suggested shape, hoisting the overlay before the return:

```python
    overlay = lifecycle_growth_evaluation_context(
        evaluation_context, displayed_count=max(0, _count(readout, "displayed_count")))
    if result["disposition"] == "rollback":
        overlay = {key: value for key, value in overlay.items()
                   if key not in ("disposition", "interpretation_state")}
```

then `**overlay,` at line 135. This keeps the additive `evidence_reason_codes` and `blocked` keys on a rollback record. We did not apply or run this patch; treat it as a suggestion, not a verified fix.

### 3. The capacity breaker cannot trip on any shipped path

`omh coding fanout dispatch --help` exposes no flag that supplies admission sources, and the only suppliers of `capacity_sources` anywhere are `tests/five_issue_cases/capacity.py:287,364,398`. With an empty sequence, `codex_admission_source()` returns `None` (`src/coding/fanout_capacity.py:272-280`), `CodexAdmissionObserver.receipt()` returns `None` (`:304`), and the only production trip call at `src/coding/fanout_dispatch.py:4115` is guarded by `if receipt is not None`. The other trip route, the `admission_check` preflight in `LaunchContext.start` (`fanout_capacity.py:173-183`), has no caller that passes it.

The split matters, so to be precise rather than sweeping. **Live on every dispatch:** `OwnerLaunchGate` is constructed and really does wrap every `Popen`, on both the direct CLI path (`fanout_dispatch.py:4044-4046`) and the Hermes-child recovery path (`:2864`). Every worker spawn is now serialized through one per-dispatch lock — a real behavioural change. **Dead:** no unit is ever marked blocked, no lineage is retained, `capacity_fields` never lands on a unit row, no `capacity_admission_observed` event is written, and the worktree-reuse path never runs.

So the body's bullet "When an executor is out of capacity, remaining units stay queued with their lineage instead of failing one by one" describes behaviour no `omh` invocation can reach. The code is honest about this — `capacity_summary` emits `native_support: 'unsupported_without_source_qualified_build'` (`fanout_capacity.py:392`) and Not Tested discloses the gap — which is why this is a body correction rather than a code change. But User/Operator Impact is where a reader looks to learn what shipped.

### 4. The Observed Evidence citations do not support the claim as cited

The capability claim itself is real: we re-ran the tracked case producers at head and 36 of 36 criterion surfaces pass, with K1, K4, K6 and K8 returning `provenance.kind == 'native'` and `native_available: True`. The native Kanban path is not a prepared claim wearing observed clothes.

The citations are the problem.

```
archived kanban.py        fc318511…    head 1c7e7b18…
archived kanban_native.py 4aa213ad…    head 36fd6fe4…
```

The archived proof was produced against different source than the head, and the author's own `evidence-index.json` labels itself `"classification": "pre_fix_native_execution"` with `"native_rerun_after_parent_notice": false`. Separately, the cited runner `.omc/qa/five_issues.py` and the artifact tree under `.omc/artifacts/five-issues/phases/C/` are both inside gitignored `.omc/`, existing only in the author's worktree. No reviewer can run the cited command or read the cited artifacts. The claim survives because we reproduced it independently; the citation does not.

### 5. Absolute paths into one developer's home are committed

`tests/five_issue_cases/kanban_native.py:28-30`:

```python
_HOST = Path('/Users/<user>/.hermes/hermes-agent')
_PYTHON = _HOST / 'venv/bin/python'
_CLI = Path('/Users/<user>/.local/bin/hermes')
```

Three consequences: the native lane can never run in CI or on any other machine, a user-identifying path ships in the repository, and the PR's only native evidence is permanently unreproducible off one host. The fallback is honest — `tests/five_issue_cases/kanban.py:266-271` returns `pass: False` with `blocked_reason: native_normal_loop_and_review_dispatch_not_observed`, pinned by `tests/test_agent_board_kanban.py:595-628` — so this is not a correctness defect. It should be an environment variable or a discovery probe.

### 6. A `Tested:` trailer claims a number it cannot support

`c2900539` and `90c874bc` both state "full suite 10606 tests exit 0", the identical total. Static test-method counts between them: `786d23b0` (merge base) 10561, `c2900539` 10697, `90c874bc` 10762. The suite grew by 65 methods between the two commits, so one runtime total cannot describe both; the number was carried forward rather than re-measured. For contrast, `ea59f0db`'s "236 tests" is exact, and all 14 commits are otherwise structurally correct with `Signed-off-by:` last and matching the author.

---

## Nits

1. **A reinvented secure-open helper, in the area that is currently failing on Windows.** `read_unit_result_input` (`src/coding/fanout_unit_results.py:87-107`) hand-rolls a bounded no-follow read. `src/system/secure_regular_file.py` already provides `open_regular_read` (`:193`) and `read_bounded` (`:255`), and it carries the Windows reparse-point handling that the POSIX-flag guard has to approximate with a device/inode recheck.
2. **A new generated-artifact family has no map row and no reachable regen command**, so the next person to touch its source has no documented way to regenerate it and the byte gate will find the drift instead.
3. **The output-spill removal is not named in the PR body.** The deleted block at `fanout_dispatch.py` (old lines 3865-3910) removed the journal summary's output tail and its content-addressed spill pointer. The body describes only additions, so a reader would not learn that a failed unit's summary now carries no output text at all.
4. **A generically named capacity reader hard-codes one adapter's build revision.** `src/coding/fanout_capacity.py:357` accepts `definition_revision` only in `(None, CODEX_ADMISSION_REVISION)`; a second adapter that pins its own revision has its whole `fanout_capacity_unit/v1` record silently dropped from journal and status, with no error. The neutral `None` path stays open, so nothing shipped is blocked.
5. **Two files listed as touched are unchanged.** The body lists `fanout_admission.py` and `fanout_retry.py`; `git diff --stat origin/main...HEAD` is empty for both.

Plus 7 similar items, mostly wording and dead plumbing: a `spill_dir` still passed but no longer read while the new docstring says "no raw spill"; a never-started unit closing its progress row as `failed` (pre-existing pattern, not a new break); a successful bounded-research delegation receipted as `failed` (unreachable today, since the bridge gates on `kanban_` prefixes); `status_card/v1` gaining a severity value outside its existing vocabulary, with no consumer switching on it; a boolean receipt field backticked as if it were a receipt state in always-loaded prompt text; a command synopsis omitting a newly shipped flag; and "serialized per board" in the body describing what is really deduplication of one create — `idempotency_key` is forced equal to `request_id`, and two distinct request ids can arm concurrently.

---

## What we checked and found clean

Stated so this review does not imply coverage nobody established.

- **Generated artifacts.** All six byte gates green, tap-skills 206/206, demo cards parse-equal, `release drift --json` `drift_count: 0`. We regenerated every family ourselves and the tree stayed clean, so both edited generated files genuinely match their producers.
- **No new installable skill ships.** `skills/omh-agent-board/` exists at the merge base and is only regenerated here; no `SkillDefinition` was added and `catalog_definitions.py`, all of `src/routing/`, and all of `src/quality/` are untouched. The new surface is the plugin tool `omh_agent_board`, and every registration surface moved with it. It passes the schema validation that landed in the merge base.
- **Counts and budgets.** Re-derived from the producer: 202 negative controls, 345 interventions, 547 total, `overroute_count` 0, `missed_intervention_count` 0, `routing_precision_errors()` `[]`. Nothing needed to move because no routing case was added. Both raised character budgets are exact, which is this repo's existing convention.
- **Execution boundary.** Every `Popen`/spawn site in `src/coding/` is behind the launch gate on both paths. Nothing runs by default, no automatic retry, no resume that executes — the resume surface returns copy-only commands. No network, LLM, or API call in the new modules. Per-unit worktree isolation holds and no branch is merged by the bridge. Every new broad `except` is classified in `tests/test_broad_exception_policy.py`.
- **Evidence contracts.** The five result states stay distinct and no path promotes prepared into observed. We constructed all four adversarial inputs — stale, foreign, replayed, malformed — and each was refused; tampered snapshot restores promoting `failed` to `observed` were rejected as `invalid_state`. Receipts persist only schema version, digests, an observation ref, the operation, the landed identity, state and the claim boundary; a sentinel string and a fake `sk-` key in a comment body reached the ephemeral tool result and never the disk. Correlation degrades to `unavailable` or a block under a shared argument digest, an absent session id, and a reused host id, with no cross-task attribution.
- **Compatibility.** The six existing artifact shapes are defined in two files this PR does not touch; absent-context output is a genuine no-op, including the eagerly-evaluated argument, which cannot raise. No closed-key `/vN` record gained a required field — `runtime_observation/v1` is explicitly protected with a comment and a separate journal record.
- **Gates on POSIX.** Full suite 10792 tests OK with one pre-existing skip, matching the body. ruff, compileall, all seven doc checks, `harness validate` (77 harnesses, 128 skills), `release checklist --json`, `release hermes-smoke`, the installed-CLI smoke and `git diff --check` all pass, including the four boxes left unchecked in the body.
- **No clock-sensitive fixtures** in the new test surface: no `time.sleep`, and the two `time.monotonic()` uses are deadline computation rather than assertions.

**Not established:** why the five `dispatcher receipt missing` and three `A did not cross Popen` Windows failures occur, or why the D5 verification case does not fail on Windows as its case requires; whether the previous head's 15 ordering errors shared one upstream trigger or several; a live call site for `build_agent_board_status_interaction` or `build_fanout_status_interaction`, so nit 5's reachability is unknown in both directions; `pre_agent_board` end to end; and an audit of `tests/five_issue_cases/*`, so the "36 criterion surfaces" figure is ours by re-running the producers, not by reading them.

---

## Recommendation

Not mergeable at this head, on finding 1 alone. The tractable next step is the `.gitattributes` pin extended to `capability_families.json`, and a newline-normalising assertion for the two subprocess byte compares; the eight remaining Windows failures need a mechanism before they need a fix. Findings 2 and 3 are small and independent of the build. Findings 4, 5 and 6 are accuracy fixes to the body, the trailers, and one test file.

On delivery grain, for the record rather than as a request: five issues in one PR is a lot, and commit `90c874bc` — 26 files, +2892, wiring all five features through the wrapper contract, the catalog, the render layer and the generated projections in one pass — is the honest argument for why splitting them afterwards would have been worse. The cost is visible in this review: the areas we could examine least carefully are exactly where the Windows failures landed.

sionic-khope:
Two refinements to the review above, both from re-checking run 34378165298 after posting. Neither changes the tally.

**A candidate mechanism for the eight failures I said needed one.** The five `dispatcher receipt missing` and three `A did not cross Popen` failures look like one upstream cause rather than two: on Windows the units are not reaching spawn at all. That makes the pair consistent — after `a88a8526` the dispatch observation is now recorded on the pre-spawn path, which is why the `ValueError` is gone, but the receipt that would follow a real spawn is still absent, and the capacity assertion that a unit crossed `Popen` fails for the same reason. So the fix corrected the journal ordering without changing whether the process starts. If that reading holds, the eight are one defect and the question is why the spawn does not happen on `windows-latest` — the commit message's own note about the confinement backend being unsupported there is the obvious place to look. I have not confirmed this; it is a hypothesis that fits all eight, offered because it is cheaper to check than eight separate investigations.

**A correction on D5.** My review listed the D5 verification case as failing on Windows. That was observed at the previous head `ea59f0db`. At this head its status is **unobserved**, not passing and not failing: D5 runs in shard 0, and shard 0 was fail-fast cancelled 20 seconds after shard 1 failed. Shard 0 did run for 20m12s, but `unittest` buffers its summary until the end, so the truncated log carries no failure headers and no totals — absence of failures there is not evidence of success. The honest statement is that D5 is unresolved and unmeasured at `a88a8526`, and the commit touches only `.gitattributes` and 14 lines of `fanout_dispatch.py`, none of it the verification path.

This is also the sharper form of the fail-fast point in the review: it is not merely that one shard was cancelled each time, it is that the cancelled shard **swapped** between the two runs. Neither head has ever been fully measured on Windows. Adding `fail-fast: false` to the matrix at `.github/workflows/ci.yml:156` would make the next run answer both shards at once, which seems worth doing before the next fix attempt rather than after.

sionic-khope:
Note for future readers, not a reopened argument.

While reviewing this PR I diagnosed the CR-as-unsafe-control behaviour in `SanitizedStream` as a product defect — the reasoning being that every Windows child writing text emits CRLF, so a diagnostic whose lines are all on the safe-line allowlist would still come back `withheld / unsafe_control`. `99408c8d` reached the opposite conclusion, that the guard was right and the native fixture framing was wrong, fixed it fixture-side, and has green Windows shards behind it. That is the correct diagnosis for the cases exercised here and I withdrew mine.

The narrow thing worth recording: every case I examined was a fixture whose framing was the actual bug, so **the guard was never tested against real, non-fixture Windows output** — an actual coding CLI writing text-mode stderr on Windows. My withdrawal should not be read as evidence that the guard is correct in that case, only that it was correct for these. If someone later sees a real Windows unit come back `withheld / unsafe_control` for output that looks allowlisted, that is the untested corner and not a regression from this PR.

Recorded so the absence of a test is visible rather than inferred from the retraction.

PR #1461, opened 2026-09-10
docs(readme): point model routing to /omh-model-setup

frirenai:
### What Changed

- Kept the existing `omh doctor` block unchanged and added a new shell block directly below it in the English, Korean, Japanese, and Chinese READMEs.
- The new block uses a localized `#` comment followed by `/omh-model-setup` for model onboarding and model-routing configuration.
- Added regression coverage for the separate-block placement, exact localized text, and no-link contract in all four READMEs.

### Why This Exists

The Quick Start diagnostics step did not expose the Hermes workflow that owns model onboarding and model-routing configuration.

### User / Operator Impact

Users can discover and copy `/omh-model-setup` from its own block directly below `omh doctor`.

### How It Works

Documentation only. `omh doctor` stays in its own block; each localized README adds a second block containing a localized shell comment and `/omh-model-setup`. No documentation link is added.

- `README.md`, `README.ko.md`, `README.ja.md`, `README.zh.md`
- `tests/test_readme_highlights.py`

### Affected Surfaces

- [x] Hermes chat or wrapper
- [ ] Codex
- [ ] Claude Code
- [ ] Generic executor
- [x] Not executor-specific

## Validation

- [x] Relevant dry-run or smoke command
- [x] `git diff --check`

### Observed Evidence

- 171 targeted README, router-content, model-recommendation, explicit-invocation, and distribution tests passed.
- The focused localization/placement regression failed for all four READMEs before the correction, then passed afterward.
- The first current-head CI run exposed stale exact-block assertions in `test_router_content`; those assertions and the localized README line budget were updated, and all 111 router-content tests now pass locally.
- Ruff and compileall passed for the changed test.
- Workflow docs and all 206 tap skills were current.
- `/omh-model-setup` selected `model-setup` and returned `setup_guide` through the real chat interaction adapter.

- Full repository suite and live Hermes UI were not run; this is a localized README-only behavior change with focused coverage and a route smoke.

## Risk

The installed skill label or localized wording could drift. The regression pins the current label, placement, languages, and no-link requirement.

## Compatibility / Rollout

Additive documentation-only change; no runtime or schema behavior changes.

## Release / Claims

- [x] Release-channel impact considered
- [x] Runtime/native capability claims are backed by evidence
- [x] Evidence contains no credentials or private data

## Follow-Up

- Keep all four Quick Start blocks aligned if the model-setup skill label changes.

### DIFF
--- a/README.ja.md
+++ b/README.ja.md
@@ -127,6 +127,11 @@ omh update
 omh doctor
 ```
 
+```sh
+# モデルのオンボーディングやモデルルーティングの設定には、Hermes でこのスキルを使ってください:
+/omh-model-setup
+```
+
 <details>
 <summary><b>その他のインストール方法</b> — Homebrew、Bun、npm、Hermes skill tap、手動フォールバック</summary>
 
--- a/README.ko.md
+++ b/README.ko.md
@@ -127,6 +127,11 @@ omh update
 omh doctor
 ```
 
+```sh
+# 모델을 온보딩하거나 모델 라우팅을 설정하려면 Hermes에서 이 스킬을 사용하세요:
+/omh-model-setup
+```
+
 <details>
 <summary><b>다른 설치 경로</b> — Homebrew, Bun, npm, Hermes skill tap, 수동 대체 경로</summary>
 
--- a/README.md
+++ b/README.md
@@ -135,6 +135,11 @@ registration.
 omh doctor
 ```
 
+```sh
+# To onboard models or configure model routing, use this skill in Hermes:
+/omh-model-setup
+```
+
 <details>
 <summary><b>Other installation paths</b> — Homebrew, Bun, npm, Hermes skill tap, manual fallback</summary>
 
--- a/README.zh.md
+++ b/README.zh.md
@@ -124,6 +124,11 @@ omh update
 omh doctor
 ```
 
+```sh
+# 如需接入模型或配置模型路由，请在 Hermes 中使用此技能：
+/omh-model-setup
+```
+
 <details>
 <summary><b>其他安装方式</b> — Homebrew、Bun、npm、Hermes skill tap、手动回退</summary>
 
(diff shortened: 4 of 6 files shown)


sionic-khope:
The CI failure is real and the diagnosis is short: the guidance line went *inside* the fenced block.

```
 ```sh
 omh doctor
+# To onboard models or configure model routing, use the /omh-model-setup skill in Hermes.
 ```
```

`tests/test_router_content.py:3745` requires each of the four READMEs to contain the block exactly:

```py
self.assertIn("\`\`\`sh\nomh doctor\n\`\`\`", localized_readme)
```

Adding a line between `omh doctor` and the closing fence breaks that exact match, which is why all four localized READMEs fail at once even though the change looks additive. The neighbouring assertions at `:3823-3828` say what the rule is protecting, in the test's own words: *"`omh doctor` belongs in its own block, never bundled into the install"* — hence both the positive `assertIn` and the negative `assertNotIn("\`\`\`sh\nomh setup\nomh doctor\n\`\`\`")`.

Verified rather than inferred: on `origin/main` all four READMEs contain the bare block; on this head none of them do, and the only change in each is the inserted comment line.

The fix is to move the sentence out of the fence and keep it as prose. That is what an earlier revision of this PR did, and it read better anyway — a shell fence that a reader might paste should contain only what is safe to paste, and `# To onboard models...` is documentation, not a command.

Two smaller things while you are in here, neither blocking:

- The four-language parity is good to see. Worth confirming the Japanese file still fits its line ceiling — `README.ja.md` was at 489 lines against a hard `< 490` in `tests/test_router_content.py` earlier today, so it has almost no headroom.
- The link target checks out: `docs/INSTALLATION.md` has a `## Guided Model Setup` heading, so the anchor resolves. The documentation navigation gate that landed in #1451 will now verify that kind of link on every run.

Separately, the three cancelled `test` legs are matrix fail-fast, not independent failures — only `test (3.11, 1)` and `test-windows (1)` actually failed.

PR #1512, opened 2026-09-12
fix: bind OMH runtime state to the active Hermes profile

Joeywrz:
## Current-head review follow-up

Head: `4339deca27d2967941b022d0d41cd725c169335f`.

Addressed all five Important findings and four Nits from sionic-khope's review of `2e40351e`; the point-by-point response is in the PR conversation. Changes preserve the original profile-isolation goal rather than redesigning it:

- Classify binding faults before hook I/O, and block pre-tool calls when their rules owner cannot be established. Direct reader and rule-error semantics remain explicit.
- Distinguish actual native registration/task scopes from a colocated standalone CLI; probe required host APIs without caching identity. Test ordinary single-owner callbacks as well as routed profiles.
- Reject native model home overrides before home/config resolution or observation. Restore real tool-cycle tests and verify malformed-config/missing-API refusals produce JSON without file changes.
- Compare both literal values and presence of the winning raw/effective configuration leaf. Preserve managed/null precedence.
- Restore trusted standalone observer overrides; document artifact-location defaults, canonical store identity, logical cwd, and intentional blank-home refusal. Reject named-user expansion and classify child binding failures without spawning; tolerate genuinely missing optional memory exports.

### Observed verification for this follow-up

- **Complete local Python 3.11 suite at this exact head: 11,568 tests run, OK, 3 skips.** Clean child environment and synthetic homes; unchanged candidate SHA/tree and clean worktree after execution. The three skips are platform/filesystem-specific, not failed cases silently removed.
- **Complete opt-in actual-Hermes native suite at this head: 74 passed, zero failures, no retries.** Both native test files used the canonical per-file runner against an isolated Hermes source copy. No inference or live platform delivery was performed.
- Independent review covered the effective PR and finite feedback obligations. Its one remaining probe-order finding was fixed, followed by a bounded independent re-review with the original reproductions passing; no Important findings remained in the reviewed scope. The subsequent one-line fixture-only adjustment preserves the static CI inventory and all assertions.
- Static shard planner passed, assigning 11,530 planned IDs exactly once. Planner and unittest-discovery totals are different inventories, not additive test counts.
- Ruff (`src tests tools packaging`), compileall, whitespace and all eight documentation checks passed. Final product sources match the separately built/installed wheel smoke (standalone roots and the real status handler); the only later diff is the CI-planner-compatible test fixture.
- **[Hosted CI run 34703960318](https://github.com/rlaope/oh-my-hermes/actions/runs/34703960318): SUCCESS at `4339deca27d2967941b022d0d41cd725c169335f`. All 13 workflow jobs plus DCO passed**, including both Linux Python versions, both Windows shards, quarantine lanes, distribution on Linux/macOS and the fail-closed aggregate. Previous-head CI remains historical evidence.

The first complete follow-up run exposed a static-planner fixture shape and a local macOS test-environment issue. The fixture was corrected without relaxing the planner; preserving non-secret `TMPDIR` restored the existing macOS alias test. The complete frozen suite was then rerun successfully as recorded above.

No merge, productive installation update, live model/platform delivery or deployment was performed. Historical implementation evidence below stays attributed to its original head.

### What Changed

Bind OMH's implicit runtime state to the **active Hermes profile**, rather than the gateway process's environment or working directory. The copied plugin and installed package share one uncached root resolver; providers, tools, hooks, optional observers and evidence subprocesses use the same owner binding.

This is an OMH-only fix: **no Hermes core patch, new dependency, global environment switching, or automatic gateway/data migration**. Opened following private maintainer coordination under `SECURITY.md`.

### Why This Exists

Hermes multiplexed gateways switch profile, secret and working-directory context per task. Several OMH components instead used process `OMH_HOME`/`HERMES_HOME`, literal OS-home defaults, or process cwd. Synthetic tests with real Hermes loaders reproduced reads and writes reaching the launch profile's store despite a correctly selected secondary profile.

Principal/audience filtering and store containment operate **after** root selection and do not fix a wrong root. This complements #1511: stable project identity remains intact, and the provider now derives that identity from its bound logical cwd rather than a foreign launch checkout.

**Impact boundary:** no production credential leak or live model disclosure is claimed. Direct provider/tool probes exposed a synthetic private marker; the ordinary native turn-start path cleared that pack before return but still wrote state to the wrong store. Both read and write paths are covered rather than treating empty output as isolation.

### User / Operator Impact

- Correctly configured profiles retain their own OMH/Hermes state when sharing a gateway process.
- Explicitly shared stores and standalone operator overrides remain supported.
- Missing, malformed or unverifiable routed bindings fail explicitly before foreign-store I/O.
- Existing profile settings and scoped legacy `OMH_HOME` bindings are supported; managed configuration keeps native precedence and empty-section semantics.
- Model-facing evidence paths cannot expand arbitrary scoped credential variables.

### How It Works

- A small bundled `runtime_paths` resolver with uncached owner binding selects trusted explicit pairs, effective profile settings and scoped legacy values. Package defaults reuse it; no additional current-profile singleton or ContextVar is introduced.
- Long-lived provider instances bind before availability checks and retain their homes for background work. Anonymous scope values are checked against the native home-owned snapshot when home-only registration could otherwise mix owners.
- Tools, hook journals and optional observer/egress/browser registrations use consistent roots. Stateful session-correlation keys include the relevant owner identity; pure caches are unchanged.
- Trusted configuration expansion is separate from model-facing path validation. Existing command allowlists, `shell=False`, timeouts, secret stripping, admission and integrity checks remain in place.
- Project root/identity uses initialized logical cwd. A bound absent cwd remains unresolved during later rendering instead of invoking the upstream resolver's ambient-cwd fallback. #1511 identity refresh, selector behavior, migration and v3 receipts are preserved.
- Evidence subprocesses receive only the bound non-secret homes in the existing minimal environment.

`src/plugin_bundle/omh/runtime_paths.py`; provider/runtime readers; tool and hook defaults; observer registration and scoped correlation caches; `src/system/paths.py`; operator documentation in `docs/INSTALLATION.md`; standalone unit tests and real-native integration tests under `tests/native/`.

### Affected Surfaces

- [x] Hermes chat or wrapper
- [ ] Codex — no executor-specific implementation or live execution
- [ ] Claude Code — no executor-specific implementation or live execution
- [ ] Generic executor — no executor dispatch change
- [x] Not executor-specific

## Validation

Prior implementation validation head (historical evidence below): `2e40351ee799f6b30ee6623393030f68cbcfe77c`, based on `e2844979c083aceee417923d39d17f75c9385999` (#1511).

- [ ] `uv run python -m omh.cli harness validate` — not run; no harness contract change
- [ ] `uv run python -m omh.cli release checklist --json` — not a release operation
- [ ] `uv run python -m omh.cli release hermes-smoke` — not run; isolated native/bundle smokes below
- [x] `git diff --check`
- [x] `omh --help` — candidate CLI module entry point
- [ ] `omh --omh-home /tmp/omh-smoke --hermes-home /tmp/hermes-smoke release hermes-smoke --install-path setup --omh-command omh --include-command-smoke` — full setup/release smoke not run
- [ ] Relevant workflow-learning review queue smoke — N/A; learning behavior is unchanged
- [x] Relevant dry-run or smoke command — wheel installation and actual copied-plugin bundle/source-byte checks
- [ ] Manual Hermes/TUI check — no interactive session or live platform/model delivery; actual native integration exercised without inference

### Observed Evidence

| Check | Result |
|---|---|
| Complete local unittest suite | **11,547 tests run; OK, 3 skips; no failures/errors** |
| Real-native profile integration | **63 cases passed** |
| Affected standalone selection | **1,684 passed across 64 modules**; overlaps full suite, not an additional total |
| Ruff, compileall, whitespace | Passed |
| Documentation projections | All eight checks passed: workflows, roles, claims, navigation, capability-families, ulw-inventory, ulw-site, chain-table |
| Wheel and copied plugin | Built, installed in isolation, source-byte checks passed |
| Patch application | Applied to the exact base in a disposable clone; resulting tree equals tested tree |
| Independent source reviews | Reported Important findings fixed and rechecked; bounded #1511 integration review found none remaining |

The full suite ran on macOS/Python 3.13.5 in an independent Git clone with a separate strict-editable environment and synthetic homes; tracked source remained unchanged. The skips were Linux dumpability, Windows process cleanup, and a case-collision check unsupported by the host filesystem.

Native tests used Hermes 0.21.2 APIs on macOS/Python 3.11. They cover both plugin/provider loading orders, both profile orders, concurrent task scopes, explicit sharing, ambiguous/missing bindings, partial registration, managed configuration, untrusted inputs, minimal child environments, bound-provider background work, and logical cwd/project identity. Same-basename projects, explicit identities without `.git`, normalized remotes, missing identities and later foreign-profile rendering have positive and negative controls.

To run the opt-in native tests, set `HERMES_SOURCE` to an absolute checkout with the supported native APIs, then from this repository run:

```sh
"$HERMES_SOURCE/scripts/run_tests.sh" \
  "$PWD/tests/native/test_profile_runtime.py" \
  "$PWD/tests/native/test_review_native.py" \
  "--native-source=$HERMES_SOURCE" \
  "--rootdir=$PWD/tests/native"
```

The historical validation rows above refer to the prior implementation head. Current-head local/native evidence and hosted-CI status are recorded at the top. Older-host compatibility, live Discord/model delivery, optional live room/browser/egress actions and exhaustive concurrent unload/first-use races remain untested. No external coding executor was started. Ordinary upstream `on_turn_start` pack invalidation is intentionally unchanged; positive correct-store recall is verified through the real queued prefetch path, not claimed as a repair of that separate lifecycle behavior.

## Risk

Shared defaults have broad integration reach. An unconfigured secondary profile or an ambiguous process-expanded template can now fail explicitly rather than inherit another owner's store. Use an absolute profile binding where required; `docs/INSTALLATION.md` explains precedence, sharing and failure behavior. Hosted operation requires the relevant native scope/config APIs; genuinely standalone OMH remains host-independent. Snapshot ownership checks add a native profile snapshot read, not a new global cache or network hydration.

## Compatibility / Rollout

Keep public/private gateway processes separate until the patched package passes the deployment's own checks. This PR does not migrate data, rewrite every profile binding, broaden tool permissions or combine gateways automatically. Explicit same-store configuration still intentionally shares that store and is not an OS sandbox. Existing principal, retention, containment and integrity guards remain necessary and unchanged.

## Release / Claims

- [x] Prepared artifacts are not presented as execution, hosted CI, merge or deployment evidence
- [x] Evidence contains only synthetic fixtures, not credentials, transcripts or unrelated private setup data
- [x] Known manual/live gaps are listed above

## Follow-Up

Current-head local/native verification and exact-head hosted CI passed, as recorded above. Review the documented compatibility choices before release. Unrelated memory lifecycle behavior and sandbox test-location portability are intentionally outside this contribution. No automatic merge or production deployment is requested.

### DIFF
--- a/src/plugin_bundle/omh/hooks/llm_hooks.py
+++ b/src/plugin_bundle/omh/hooks/llm_hooks.py
@@ -1,5 +1,7 @@
 from __future__ import annotations
 
+from .. import runtime_paths
+
 import errno
 from datetime import datetime, timezone
 import hashlib
@@ -159,7 +161,7 @@ def pre_llm_call(**kwargs) -> dict[str, object] | None:
     # Turn start is the freshest in-process view of the Shift+Tab yolo flag:
     # a toggle shows on the HUD at the user's next message, not only at the
     # next tool call.
-    record_approval_bypass(omh_home=str(kwargs.get("omh_home", "") or ""))
+    record_approval_bypass(omh_home=str(runtime_paths.plugin_home(kwargs.get("omh_home"))))
     context_parts: list[str] = []
     payload: dict[str, object] = {}
     user_message = "" if _tracker_event_is_present(kwargs) else str(kwargs.get("user_message", "") or "")
@@ -198,7 +200,7 @@ def pre_llm_call(**kwargs) -> dict[str, object] | None:
             if not claim_route_guidance_delivery(
                 session_id=session_id,
                 route_fingerprint=route_fingerprint,
-                omh_home=str(kwargs.get("omh_home", "") or ""),
+                omh_home=str(runtime_paths.plugin_home(kwargs.get("omh_home"))),
             ):
                 route_hint_context = ""
                 message_matches_awareness = False
@@ -249,26 +251,26 @@ def pre_llm_call(**kwargs) -> dict[str, object] | None:
     # with the HUD checklist unnoticed. Honors the caller's awareness opt-out.
     if include_awareness:
         outcomes = unacknowledged_outcomes(
-            str(kwargs.get("omh_home", "") or ""),
-            str(kwargs.get("hermes_home", "") or ""),
+            str(runtime_paths.plugin_home(kwargs.get("omh_home"))),
+            str(runtime_paths.plugin_home(kwargs.get("hermes_home"), hermes=True)),
             session_id,
         )
         todo_reminder = open_todo_reminder(
-            omh_home=str(kwargs.get("omh_home", "") or ""),
-            hermes_home=str(kwargs.get("hermes_home", "") or ""),
+            omh_home=str(runtime_paths.plugin_home(kwargs.get("omh_home"))),
+            hermes_home=str(runtime_paths.plugin_home(kwargs.get("hermes_home"), hermes=True)),
             session_ref=session_id,
             outcomes=outcomes,
         )
         if todo_reminder:
             context_parts.append(todo_reminder)
         workflow_context = active_workflow_context(
-            str(kwargs.get("omh_home", "") or ""), session_id
+            str(runtime_paths.plugin_home(kwargs.get("omh_home"))), session_id
         )
         if workflow_context:
             payload["omh_active_workflow"] = workflow_context
             context_parts.append(render_active_workflow_context(workflow_context))
         budget_context = context_budget_continuation(
-            str(kwargs.get("omh_home", "") or ""), session_id, str(kwargs.get("model", "") or "")
+            str(runtime_paths.plugin_home(kwargs.get("omh_home"))), session_id, str(kwargs.get("model", "") or "")
         )
         if budget_context:
             payload["omh_context_budget"] = budget_context
@@ -286,8 +288,8 @@ def pre_llm_call(**kwargs) -> dict[str, object] | None:
                 ""
                 if outcomes
                 else _todo_stall_status(
-                    str(kwargs.get("omh_home", "") or ""),
-                    str(kwargs.get("hermes_home", "") or ""),
+                    str(runtime_paths.plugin_home(kwargs.get("omh_home"))),
+                    str(runtime_paths.plugin_home(kwargs.get("hermes_home"), hermes=True)),
                     session_id,
                 )
             ),
@@ -298,8 +300,8 @@ def pre_llm_call(**kwargs) -> dict[str, object] | None:
 
     omh_home: str | None = None
     try:
-        omh_home = str(kwargs.get("omh_home", "") or "") or None
-        hermes_home = str(kwargs.get("hermes_home", "") or "") or None
+        omh_home = str(runtime_paths.plugin_home(kwargs.get("omh_home"))) or None
+        hermes_home = str(runtime_paths.plugin_home(kwargs.get("hermes_home"), hermes=True)) or None
         try:
             activity = read_omh_activity(omh_home=omh_home, limit=3)
         except Exception as exc:
--- a/src/plugin_bundle/omh/hooks/llm_hooks.py
+++ b/src/plugin_bundle/omh/hooks/llm_hooks.py
@@ -19,6 +19,7 @@ from ..degradation import (
     COMPONENT_RUNTIME_STATUS_READ,
     degradation_payload,
     safe_error_type,
+    runtime_binding_degradation,
 )
 from ..active_workflow_context import active_workflow_context, render_active_workflow_context
 from ..approval_bypass import record_approval_bypass
@@ -161,14 +162,8 @@ def pre_llm_call(**kwargs) -> dict[str, object] | None:
     try:
         omh_home = str(runtime_paths.plugin_home(kwargs.get("omh_home")))
         hermes_home = str(runtime_paths.plugin_home(kwargs.get("hermes_home"), hermes=True))
-    except (OSError, RuntimeError) as exc:
-        error_type = "RuntimeError" if getattr(exc, "errno", None) == errno.ELOOP else type(exc).__name__
-        degradation = degradation_payload([(COMPONENT_RUNTIME_STATUS_READ, safe_error_type(error_type))])
-        return {
-            "omh_degradation": degradation,
-            "context": "[OMH Degraded] components=" + COMPONENT_RUNTIME_STATUS_READ
-            + ". Runtime home binding failed; no runtime state was read or written.",
-        }
+    except (runtime_paths.RuntimeBindingError, OSError, RuntimeError) as exc:
+        return runtime_binding_degradation(exc)
     record_active_main_agent_model(kwargs.get("model"))
     observe_plugin_hook_call("pre_llm_call", kwargs)
     # Turn start is the freshest in-process view of the Shift+Tab yolo flag:
--- a/src/plugin_bundle/omh/hooks/llm_hooks.py
+++ b/src/plugin_bundle/omh/hooks/llm_hooks.py
@@ -156,12 +156,25 @@ def _tracker_event_is_present(kwargs: dict) -> bool:
 
 def pre_llm_call(**kwargs) -> dict[str, object] | None:
     """Inject bounded OMH role/status context without storing prompts."""
+    # Bind before any observer or awareness I/O. A failed root must not reach
+    # downstream defaults as None, or fail before the status classifier runs.
+    try:
+        omh_home = str(runtime_paths.plugin_home(kwargs.get("omh_home")))
+        hermes_home = str(runtime_paths.plugin_home(kwargs.get("hermes_home"), hermes=True))
+    except (OSError, RuntimeError) as exc:
+        error_type = "RuntimeError" if getattr(exc, "errno", None) == errno.ELOOP else type(exc).__name__
+        degradation = degradation_payload([(COMPONENT_RUNTIME_STATUS_READ, safe_error_type(error_type))])
+        return {
+            "omh_degradation": degradation,
+            "context": "[OMH Degraded] components=" + COMPONENT_RUNTIME_STATUS_READ
+            + ". Runtime home binding failed; no runtime state was read or written.",
+        }
     record_active_main_agent_model(kwargs.get("model"))
     observe_plugin_hook_call("pre_llm_call", kwargs)
     # Turn start is the freshest in-process view of the Shift+Tab yolo flag:
     # a toggle shows on the HUD at the user's next message, not only at the
     # next tool call.
-    record_approval_bypass(omh_home=str(runtime_paths.plugin_home(kwargs.get("omh_home"))))
+    record_approval_bypass(omh_home=omh_home)
     context_parts: list[str] = []
     payload: dict[str, object] = {}
     user_message = "" if _tracker_event_is_present(kwargs) else str(kwargs.get("user_message", "") or "")
@@ -200,7 +213,7 @@ def pre_llm_call(**kwargs) -> dict[str, object] | None:
             if not claim_route_guidance_delivery(
                 session_id=session_id,
                 route_fingerprint=route_fingerprint,
-                omh_home=str(runtime_paths.plugin_home(kwargs.get("omh_home"))),
+                omh_home=omh_home,
             ):
                 route_hint_context = ""
                 message_matches_awareness = False
@@ -251,26 +264,26 @@ def pre_llm_call(**kwargs) -> dict[str, object] | None:
     # with the HUD checklist unnoticed. Honors the caller's awareness opt-out.
     if include_awareness:
         outcomes = unacknowledged_outcomes(
-            str(runtime_paths.plugin_home(kwargs.get("omh_home"))),
-            str(runtime_paths.plugin_home(kwargs.get("hermes_home"), hermes=True)),
+            omh_home,
+            hermes_home,
             session_id,
         )
         todo_reminder = open_todo_reminder(
-            omh_home=str(runtime_paths.plugin_home(kwargs.get("omh_home"))),
-            hermes_home=str(runtime_paths.plugin_home(kwargs.get("hermes_home"), hermes=True)),
+            omh_home=omh_home,
+            hermes_home=hermes_home,
             session_ref=session_id,
             outcomes=outcomes,
         )
         if todo_reminder:
             context_parts.append(todo_reminder)
         workflow_context = active_workflow_context(
-            str(runtime_paths.plugin_home(kwargs.get("omh_home"))), session_id
+            omh_home, session_id
         )
         if workflow_context:
             payload["omh_active_workflow"] = workflow_context
             context_parts.append(render_active_workflow_context(workflow_context))
         budget_context = context_budget_continuation(
-            str(runtime_paths.plugin_home(kwargs.get("omh_home"))), session_id, str(kwargs.get("model", "") or "")
+            omh_home, session_id, str(kwargs.get("model", "") or "")
         )
         if budget_context:
             payload["omh_context_budget"] = budget_context
@@ -288,8 +301,8 @@ def pre_llm_call(**kwargs) -> dict[str, object] | None:
                 ""
                 if outcomes
                 else _todo_stall_status(
-                    str(runtime_paths.plugin_home(kwargs.get("omh_home"))),
-                    str(runtime_paths.plugin_home(kwargs.get("hermes_home"), hermes=True)),
+                    omh_home,
+                    hermes_home,
                     session_id,
                 )
             ),
@@ -298,10 +311,7 @@ def pre_llm_call(**kwargs) -> dict[str, object] | None:
         if claim_finding:
             context_parts.append(f"[OMH continuation claim] {claim_finding}")
 
-    omh_home: str | None = None
     try:
-        omh_home = str(runtime_paths.plugin_home(kwargs.get("omh_home"))) or None
-        hermes_home = str(runtime_paths.plugin_home(kwargs.get("hermes_home"), hermes=True)) or None
         try:
             activity = read_omh_activity(omh_home=omh_home, limit=3)
         except Exception as exc:
--- a/src/plugin_bundle/omh/host_observation.py
+++ b/src/plugin_bundle/omh/host_observation.py
@@ -1,5 +1,7 @@
 from __future__ import annotations
 
+from . import runtime_paths
+
 from contextlib import contextmanager
 from datetime import datetime, timezone
 import errno
@@ -132,7 +134,7 @@ def _observation_metadata(args: dict[str, Any], kwargs: dict[str, Any]) -> dict[
     for source in (args.get("observation"), kwargs.get("observation"), kwargs.get("omh_observation")):
         if isinstance(source, dict):
             metadata.update(source)
-    for key in ("host", "session_id", "source", "message", "omh_home", "hermes_home"):
+    for key in ("host", "session_id", "source", "message"):
         if args.get(key) is not None:
             metadata.setdefault(key, args.get(key))
         if kwargs.get(key) is not None:
@@ -186,8 +188,8 @@ def _record_observation(metadata: dict[str, Any], *, event: str, tool: str, hook
             from omh.plugin_observations import record_plugin_host_observation
 
             paths = resolve_paths(
-                omh_home=str(metadata.get("omh_home", "") or "") or None,
-                hermes_home=str(metadata.get("hermes_home", "") or "") or None,
+                omh_home=runtime_paths.default_omh_home(),
+                hermes_home=runtime_paths.default_hermes_home(),
             )
             return record_plugin_host_observation(
                 paths,
@@ -270,7 +272,7 @@ def _record_standalone_observation(
     evidence_refs: list[str],
     message: str,
 ) -> dict[str, Any]:
-    omh_home = _expand_path(str(metadata.get("omh_home", "") or "") or os.environ.get("OMH_HOME", "~/.omh"))
+    omh_home = runtime_paths.default_omh_home()
     runtime_dir = omh_home / "runtime"
     runtime_dir.mkdir(parents=True, exist_ok=True)
     try:
--- a/src/plugin_bundle/omh/host_observation.py
+++ b/src/plugin_bundle/omh/host_observation.py
@@ -147,6 +147,14 @@ def _observation_metadata(args: dict[str, Any], kwargs: dict[str, Any]) -> dict[
         metadata.setdefault("evidence_refs", args.get("evidence_refs"))
     if kwargs.get("evidence_refs") is not None:
         metadata.setdefault("evidence_refs", kwargs.get("evidence_refs"))
+    # Only standalone operator arguments may name homes. Nested observation
+    # metadata never grants store authority, even on a standalone invocation.
+    for key in ("omh_home", "hermes_home"):
+        metadata.pop(key, None)
+        if kwargs.get(key) is not None:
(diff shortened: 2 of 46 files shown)


src/plugin_bundle/omh/hooks/llm_hooks.py
```diff
@@ -159,7 +161,7 @@ def pre_llm_call(**kwargs) -> dict[str, object] | None:
     # Turn start is the freshest in-process view of the Shift+Tab yolo flag:
     # a toggle shows on the HUD at the user's next message, not only at the
     # next tool call.
-    record_approval_bypass(omh_home=str(kwargs.get("omh_home", "") or ""))
+    record_approval_bypass(omh_home=str(runtime_paths.plugin_home(kwargs.get("omh_home"))))
```
sionic-khope: Important (1): this call sits outside the `try` that classifies filesystem faults into a `runtime_status_read` degradation. `plugin_home()` → `expand_path()` → `Path.resolve()` raises `RuntimeError("Symlink loop …")` on 3.11 (the CI failure in `tests/test_runtime_reader_filesystem_faults.py::test_symlink_loop_home_is_classified_not_read_as_idle`) and can raise `OSError` generally; `RuntimeBindingError` escapes the same way, so an unbound profile kills every turn instead of degrading. Same shape at :200, :254, :291.

src/plugin_bundle/omh/runtime_paths.py
```diff
@@ -0,0 +1,224 @@
...
+            raise RuntimeBindingError("OMH runtime path variable is unavailable in this profile")
+        return resolved
+
+    expanded = _VARIABLE.sub(replace, str(value))
+    if _VARIABLE.search(expanded) or "${" in expanded:
+        raise RuntimeBindingError("OMH runtime path contains an unresolved variable")
+    path = Path(expanded).expanduser()
+    if relative_to is not None and not path.is_absolute():
+        path = relative_to / path
+    return path.resolve()
```
sionic-khope: `Path.resolve()` raises `RuntimeError` on a symlink loop (3.11) and `OSError` on other faults. Wrap it and re-raise as `RuntimeBindingError` so every caller has one failure type to classify.

src/plugin_bundle/omh/runtime_paths.py
```diff
@@ -0,0 +1,224 @@
...
+from pathlib import Path
+import re
+
+
+class RuntimeBindingError(ValueError):
+    """No safe runtime store can be selected (messages contain no path values)."""
+
+
+_VARIABLE = re.compile(r"\$(?:\{(?:env:)?([A-Za-z_][A-Za-z_0-9]*)\}|([A-Za-z_][A-Za-z_0-9]*))|%([A-Za-z_][A-Za-z_0-9]*)%")
+_MISSING = object()
```
sionic-khope, line 22: Important (2): this is an importability probe, not a bound-profile probe. In any environment where `hermes_constants` imports (a plain `omh` CLI installed beside hermes-agent), the standalone lane is left silently: `$OMH_HOME` is read from `secrets.get_secret` (:61-65, :186-191) instead of the process env, and five native APIs are used without a capability check. `src/plugin_bundle/omh/__init__.py:17-50` (`_host_supports_hook`, `_register_optional_surface`) is the repo's existing pattern for this.

src/plugin_bundle/omh/runtime_paths.py
```diff
@@ -0,0 +1,224 @@
...
+    # This host validator only reads. The normal behavioral loader deliberately
+    # tolerates malformed/unreadable YAML and can reuse last-known-good data;
+    # neither is safe for selecting a state owner. Validate before using it.
+    try:
+        raw = config.require_readable_config_before_write(home / "config.yaml")
+    except (RuntimeError, OSError, ValueError) as exc:
+        raise RuntimeBindingError("OMH profile configuration is unreadable or invalid") from exc
+    original = _setting(_overlay_config(raw, import_module("hermes_cli.managed_scope").load_managed_config()))
+    if original is not _MISSING:
+        original_path = expand_path(original, hermes_home=home, relative_to=home)
```
sionic-khope, line 238: Important (4): the raw leaf is cross-checked against the process-active config only when it contains a `$VAR`; a literal disagreement passes. The PR text says mismatching values are refused before I/O — compare `original_path != path` unconditionally.

src/plugin_bundle/omh/runtime_paths.py
```diff
@@ -0,0 +1,224 @@
...
+
+
+def plugin_home(value: object = None, *, hermes: bool = False) -> Path:
+    """Native callbacks cannot choose a store through arguments/observations.
+
+    Retain explicit standalone reader APIs for operator integrations. Within a
+    native host, these same legacy fields are not authority to cross profiles.
+    """
+    if _host() is None and value:
+        return expand_path(value)
```
sionic-khope, line 301: Important (3): in-host the explicit value is discarded without an error, while the model-facing tool schemas still advertise `omh_home`/`hermes_home` (e.g. `tools/delegate_route_tool.py:142-152`, `tools/probe_tool.py:25-32`). Either drop those fields or surface the binding error.

src/system/paths.py
```diff
@@ -814,14 +824,17 @@ def resolve_paths(
...
     return OmhPaths(
-        omh_home=expand_path(omh_home) if omh_home else default_omh,
-        hermes_home=expand_path(hermes_home) if hermes_home else default_hermes,
+        omh_home=resolved_omh,
+        hermes_home=resolved_hermes,
         # Recorded here, while the caller's intent is still known: comparing the
         # resolved home against the default later cannot tell a named home from
         # an unnamed one that happens to match it.
-        omh_home_named=omh_home is not None,
+        omh_home_named=omh_home is not None or (normalized_scope == "user" and runtime_paths._host() is not None),
```
sionic-khope: Important (2b): forcing `omh_home_named=True` whenever a host is importable changes where `project_artifact_dir` (:785) writes for an ordinary single-profile user — repo-scoped artifacts land in the profile store instead of the repo `.omh/`. Not mentioned in the docs addition; either gate on an actual binding or document it.

src/plugin_bundle/omh/tools/delegate_route_tool.py
```diff
@@ -157,8 +159,8 @@ def _chain_entry(
 def omh_delegate_route_handler(args: dict[str, Any], **kwargs) -> str:
     observation = observe_plugin_tool_call("omh_delegate_route", args, kwargs)
     action = str(args.get("action", "") or "set").strip().lower()
-    hermes_home = str(args.get("hermes_home", "") or "") or None
-    omh_home = str(args.get("omh_home", "") or "") or None
+    hermes_home = runtime_paths.plugin_home(args.get("hermes_home"), hermes=True)
```
sionic-khope, line 161: Important (3): the schema at :142-152 still offers `omh_home`/`hermes_home`, but in-host `plugin_home()` ignores them silently. `tests/test_delegate_route_tool.py:107-120` also swapped the tool-cycle assertion for a direct writer call, so the tool's home handling is no longer exercised.

src/plugin_bundle/omh/host_observation.py
```diff
@@ -186,8 +188,8 @@ def _record_observation(metadata: dict[str, Any], *, event: str, tool: str, hook
             from omh.plugin_observations import record_plugin_host_observation
 
             paths = resolve_paths(
-                omh_home=str(metadata.get("omh_home", "") or "") or None,
-                hermes_home=str(metadata.get("hermes_home", "") or "") or None,
+                omh_home=runtime_paths.default_omh_home(),
```
sionic-khope: Important (5): caller-supplied homes are ignored here (and at :272), so standalone `--omh-home` no longer reaches plugin observations; `tests/test_plugin_capabilities.py:191` and `tests/test_plugin_distribution.py:1012` were changed to set env instead of pinning that. `plugin_home()` still honours explicit values standalone — the two should agree.

src/plugin_bundle/omh/memory_provider.py
```diff
@@ -41,7 +43,9 @@
 try:  # Present only inside the Hermes process.
     from agent.memory_provider import MemoryProvider as _MemoryProviderBase
     from agent.memory_provider import RecallStatus as RecallStatus
-except ImportError:  # pragma: no cover - exercised by the repo's own test run
+except ModuleNotFoundError as exc:  # Standalone has no Hermes dependency.
+    if exc.name not in {"agent", "agent.memory_provider"}:
```
sionic-khope: Nit: narrowing to `ModuleNotFoundError` means a Hermes that renames `MemoryProvider` inside an existing module raises plain `ImportError` and hard-fails plugin import, where the previous fallback degraded.

sionic-khope (changes requested):
**5 important, 4 nits.** Round 1 — reviewed the full diff (42 files) against `origin/main` (which already carries #1511), per `REVIEW.md`. Read-only; I ran `tests/test_runtime_paths.py` and an import probe on an extracted copy of the head, nothing else.

The resolver's shape is sound and the trusted/untrusted split is real (`runtime_paths.py:47-51` rejects every variable form before `expand_path`, and `tests/test_runtime_paths.py:105-130` proves a scoped secret cannot reach tool output). Per-profile keying of `toolcall_rules._claim_fire`, `_PREPARE_CALLS` and `_BRIDGES` closes a genuine cross-profile collision. The bundle stays standalone-importable, no broad `except` was added, no generated artifact or pinned count moved, wording is executor-neutral, and the `docs/INSTALLATION.md` claims are hedged where they should be. What follows is what still blocks.

**CI.** The 3.11 shard fails for a real reason: `tests/test_runtime_reader_filesystem_faults.py::test_symlink_loop_home_is_classified_not_read_as_idle` — `runtime_paths.expand_path` (`:88`) ends in `Path.resolve()`, which raises `RuntimeError("Symlink loop …")` on 3.11, and the call sits outside the hook's degradation guard (`hooks/llm_hooks.py:164`). The two other red Linux shards are fail-fast cancellations, not independent failures.

### Important

1. **Turn-start hook crashes instead of classifying a path fault** — `src/plugin_bundle/omh/hooks/llm_hooks.py:164` (also `:200`, `:254`, `:291`). `plugin_home()` runs outside the `try` that turns filesystem faults into a `runtime_status_read` degradation, so a symlink loop (the CI failure) or any `OSError` from `resolve()` kills the turn; `RuntimeBindingError` escapes the same way, so an unbound profile kills every turn rather than degrading. Fix: wrap `.resolve()` and re-raise as `RuntimeBindingError`, and resolve homes inside the guarded block.
2. **`_host()` means "hermes_constants is importable", not "bound to a profile"** — `src/plugin_bundle/omh/runtime_paths.py:21-27`. Three consequences. (a) A plain `omh` CLI installed in the same environment as hermes-agent silently leaves the standalone lane: `$OMH_HOME` is read from `secrets.get_secret` (`:61-65`, `:186-191`) instead of the process env. (b) `src/system/paths.py:838` forces `omh_home_named=True` whenever a host is importable, so `project_artifact_dir` (`:785`) writes repo-scoped artifacts into the profile store instead of the repo `.omh/` — a default-path change for a normal single-profile user that the docs addition does not mention. (c) A hard, unprobed dependency on five native APIs (`build_profile_secret_scope`, `agent.runtime_cwd`, `hermes_cli.managed_scope.load_managed_config`, `require_readable_config_before_write`, the config loader): a version-skewed Hermes raises `AttributeError`/`ModuleNotFoundError` from every path call. The repo already solves host-shape tolerance at `src/plugin_bundle/omh/__init__.py:17-50` (`_host_supports_hook`, `_register_optional_surface`); please probe the same way.
3. **Model-facing schemas advertise an override that is now discarded** — `src/plugin_bundle/omh/tools/delegate_route_tool.py:142-152`, `tools/probe_tool.py:25-32` (same shape in chat/hud/status/todo). `plugin_home()` (`runtime_paths.py:205-211`) ignores the argument in-host with no error. Either drop the fields or return the binding error; if the schemas change, rerun `docs capability-families --check`. Related: `tests/test_delegate_route_tool.py:107-120` replaced the tool-cycle assertion with a direct writer call, so the tool's home handling is now untested.
4. **A literal disagreement between the winning leaf and the active config is accepted** — `runtime_paths.py:150-160`. The returned path comes from `load_config_readonly()` (process-active); the profile's own raw leaf is cross-checked only when it contains a `$VAR`. The PR text says mismatching values are refused before I/O; compare `original_path != path` unconditionally so that is true.
5. **"Fail before foreign-store I/O" is not uniform across call sites** — `src/plugin_bundle/omh/host_observation.py:188-192` and `:272` now ignore caller-supplied homes entirely, so standalone `--omh-home` no longer reaches plugin observations; the tests were changed to set env instead (`tests/test_plugin_capabilities.py:191`, `tests/test_plugin_distribution.py:1012`) rather than pinning the behaviour. Meanwhile `plugin_home()` still honours explicit values standalone.

### Nits

- `expand_path` always resolves symlinks, so homes change form and tests had to add `.resolve()` (`tests/test_toolcall_rules.py:155`, `tests/test_delegate_route_tool.py:410`), while project scope keeps the unresolved core `expand_path` — two forms of one home per process.
- `resolve_paths(omh_home="")` now raises (verified) where it used to default.
- `expand_input_path` blocks `$VAR`/`%VAR%` but still `expanduser()`s model input (`~root/…`); and `_minimal_child_environment` (`tools/evidence_tool.py:265-267`) resolves homes outside the new `RuntimeBindingError` guard, so that error escapes as an exception rather than JSON.
- `memory_provider.py:46-48`: a Hermes that renames `MemoryProvider` inside an existing module raises plain `ImportError` and now hard-fails plugin import.

### Also checked

`tests/native/` has no `__init__.py`, so `unittest discover -s tests` skips it and `pytest_ignore_collect` covers pytest — safe in the default suite. Coverage gaps: no test for the hook-level behaviour when a binding error is raised (1), for the literal-disagreement case (4), or for the `omh_home_named` / `project_artifact_dir` change (2b). For a machine with no profile settings and no Hermes importable, defaults are unchanged.

### Verdict

Request changes — the listed fixes, not a redesign. Top three asks: make every path fault a `RuntimeBindingError` and resolve homes inside the hook's degradation guard (1); replace `_host()`-as-importability with an explicit bound-profile probe plus a capability check for the five native APIs, and state what `omh_home_named` / `project_artifact_dir` do for a single-profile user (2); remove or reject the now-ignored `omh_home`/`hermes_home` schema fields and restore a tool-cycle test (3).

Joeywrz:
@sionic-khope Thanks for the detailed review. I checked every Important and Nit against the current candidate and pushed the follow-up in this same PR. The original review was against `2e40351e`; the current head is `4339deca27d2967941b022d0d41cd725c169335f`.

### Important findings

1. **Hook/path faults:** runtime path-resolution faults are now bounded `RuntimeBindingError`s. Turn-start, pre/post-tool and session-end hooks bind both homes before scoped observers/I/O. Failed pre-tool binding returns the native block directive because the rules owner cannot be established. Direct readers still raise; rule failures are not swallowed, and no alternate store is selected. The original symlink case and both-home/no-I/O controls pass.
2. **Native detection and compatibility:** importability alone no longer selects native mode. Required callable APIs are probed, and an active/ambiguous unsupported host fails closed. A colocated standalone CLI keeps process-env behavior. Real-host tests exposed an important lifecycle detail: ordinary single-owner memory loading and later callbacks can lack task scopes, so actual `PluginContext`/memory-collector registration also marks that bundle's execution lane. That flag stores **no home, profile or secret**; owner selection stays dynamic. Both loading orders, later owner switches and unscoped-multiplex refusal were exercised. Ordinary single-owner repo artifacts remain repo-local; routed/explicit stores are named. The invocation matrix is documented and tested, including absent logical cwd and #1511 identity behavior.
3. **Tool overrides:** retained the fields for standalone operator compatibility, but schemas explicitly say native calls reject them. All seven handlers reject native home arguments before resolving homes/configuration or observing the call. The real delegate status/set/status/clear cycle is covered again. Independent review caught a remaining ordering error in `omh_probe`; it is fixed, with both malformed-config and missing-API native controls proving bounded JSON and unchanged active/private files. Capability-family projection checks pass without generated-file edits.
4. **Configuration winner:** winning raw/effective paths are compared unconditionally, including literal values and missing-versus-present leaves. Managed precedence, shadowed templates and null-section semantics remain covered by real-host tests.
5. **Standalone observations:** explicit top-level operator homes are honored in both package-backed and bundle-only paths. Nested observation metadata never selects a store; native callbacks remain bound to their active profile. Tests use a different ambient store and assert the intended write destination. The distribution-test env additions were child-environment fixtures, so those ownership assertions were retained.

### Nits

- Runtime homes remain canonical for store/cache identity; core project paths already canonicalized before this change. The native logical-cwd distinction and symlink equivalence are documented/tested, rather than changing #1511 semantics.
- Explicit blank/whitespace homes intentionally remain invalid. This compatibility tightening is now explicit in the docs: omit the argument/use `None` for defaults. No native blank binding falls back to ambient state.
- Named-user `~user` expansion is rejected before OS lookup; safe literal/current-user paths remain supported. Evidence child-home binding errors return bounded JSON without spawning.
- Missing optional `MemoryProvider`/`RecallStatus` symbols no longer hard-fail import; unrelated internal import faults still propagate.

### Verification

**Full local suite at `4339deca`: 11,568 tests run, OK, 3 platform/filesystem skips; unchanged SHA/tree and clean worktree. Complete opt-in native suite: 74 passed, no failures/retries.** Ruff, compileall, all eight documentation checks and the built/installed wheel smoke passed (identical final product sources). **[Hosted CI run 34703960318](https://github.com/rlaope/oh-my-hermes/actions/runs/34703960318) is now successful for this exact head:** all 13 workflow jobs and DCO passed, including Linux 3.11/3.12, Windows, quarantine, distribution and aggregate. Independent review and bounded re-review found no remaining Important findings after the probe-order fix. A final fixture-only change preserves the static CI planner instead of relaxing it; the planner passes.

No merge, installed-runtime update or live model/platform delivery was performed. Please re-review the current head; I have not marked your review threads resolved or treated local review as maintainer approval.

sionic-khope:
Round 2, against `4339deca`. **1 important, 0 nits.** Eight of the nine round-one findings are fixed at the line, and the new `tests/test_runtime_binding_review.py` pins behaviour rather than mirroring the implementation — thank you for the thorough, fast follow-up, and for the clear mapping in your reply. This is careful work and it lands.

The one carryover is narrow and CI could not show it to you: `_canonical_path` (`runtime_paths.py:89-96`) classifies only when `Path.resolve()` raises, and CPython 3.13 resolves a symlink loop without raising (it returns the link itself; 3.11 and 3.12 raise). On 3.13 the loop therefore passes the turn-start guard and resurfaces at `runtime_reader.py:207` as a bare `RuntimeError` outside any `try`; with awareness off the hook returns `None`. Your own `tests/test_runtime_reader_filesystem_faults.py` shows it: 5 failures under 3.13 on this head, OK under 3.12, OK on `main` under 3.13. Hosted CI runs 3.11/3.12 only, and `requires-python` is `>=3.11`.

Rather than ask for another round I have put the one-branch fix on top of your commits in #1514, which carries all eight of your commits unchanged and adds one: a root that is still a symlink after resolution is a `RuntimeBindingError` on every interpreter, with a test that pins both the real loop and the 3.13 resolve shape. Merging #1514 merges this PR with your authorship intact. The settled round-one threads are resolved below; the hook thread closes with #1514.

For the record, not a finding: the pre-tool `action: block` on a multiplexed profile with no `omh_home` setting and no `OMH_HOME` secret is reachable from configuration absence alone. It is documented and the message is self-explanatory, so that stays a maintainer decision.

Issue #1583, opened 2026-09-15
A memory-provider comparison routes to the telemetry card, not to its declared owner


sionic-khope:
## Problem

A memory-provider comparison question routes to the operations telemetry card, not to the skill that owns memory-provider posture.

The ownership is not ambiguous. `external-connector-readiness` declares `memory_provider_posture/v1` as one of its own outputs — "identity scope, automatic hooks, storage boundary, synchronization, failure, retention, deletion, export/import, backup/restore, and portability" (`src/skills/catalog_feature_surfaces.py:807`). And `memory-sync` explicitly hands provider questions over: "Send memory-provider lifecycle questions -- enabling, switching, pausing, retention, deletion, export, or failed synchronization -- to external-connector-readiness, which owns the provider posture this review cannot establish" (`src/skills/catalog_feature_surfaces.py:200`).

The designated owner does not place in the top three.

## Complete user goal

A memory-provider question — comparing, switching, or assessing providers — reaches `external-connector-readiness`, the skill the catalog has already designated as its owner.

## Intended surfaces

Hermes chat or wrapper.

## Observed evidence

Reproduced on `origin/main` at `dec82db8`:

```
$ omh recommend "compare memory providers on my own data for retrieval quality and latency"
1. ops-observability-card [high]
   Why: Matched operational telemetry language; prepare an observed-only status card instead of starting or advancing a loop.
2. performance-goal [high]
   Why: Matched workflow trigger language for this task.
3. memory-sync [medium]
```

`external-connector-readiness` is absent. The `Why` line names the cause: "latency" and "quality" read as operational telemetry language, and that signal outweighs "memory providers".

## Proposal

Add memory-provider comparison phrasings to `external-connector-readiness`'s triggers, each with its paired negative case so genuine telemetry questions still reach `ops-observability-card`.

## Alternatives considered

Moving provider posture to `ops-observability-card` — rejected; it contradicts two explicit ownership statements already in the catalog, including a deference line `memory-sync` carries specifically to send these questions away.

## Success criteria

- The query above returns `external-connector-readiness` in first place.
- A genuine telemetry question — token, cost, latency of operations rather than of a provider choice — still returns `ops-observability-card`, pinned as a negative case in `ROUTING_PRECISION_CASES`.
- The provider-comparison case is pinned in `ROUTING_INTERVENTION_CASES`.

## Boundaries and risks

[...]

Issue #1556, opened 2026-09-15
[Feature]: Warn when the model that answered is not the model that was requested


sionic-khope:
## Problem

A run can be routed to one model and answered by another. Today nothing compares the two, so an alias that silently resolves elsewhere, a provider substitution, or a version drift is invisible to the person and to every downstream report that names the model.

The observation needed to detect this is already in the tree and already read by OMH. `src/plugin_bundle/omh/hermes_delegation.py:1713-1726` documents it directly: `session_model_usage.model` is written per call from the model that actually answered, so one distinct value there names it. The plugin already reads that table read-only at `:1826`.

## Complete user goal

When the model that answered differs from the model that was requested, the person is told so at the point the difference matters, with the alias and version difference resolved — not left to compare two opaque strings. When the two agree, or when the session cannot be judged, the surface says that instead of implying agreement.

## Intended surfaces

Hermes chat or wrapper; all executors (executor-neutral).

## Observed evidence

`_observed_wire_model()` returns `sessions.model` (the recorded value) when present and does not query `session_model_usage.model` (the answering model) at all. The recorded-first ordering hides the discrepancy in exactly the case the discrepancy exists. This is read from source, not from a live run: no attestation has been observed because no comparison exists yet.

## Proposal

Add a comparison function in `src/plugin_bundle/omh/hermes_delegation.py` that resolves both sides through the existing alias machinery and reports agreement, disagreement, or `unknown`.

Reuse, do not reimplement, the alias resolution already complete in `mixture_category_for` (`:1455-1540`): `_dated_snapshot_base` for date-pinned snapshots, `_projected_model_alias` for reasoning-mode and service-tier projection, `EXACT_CONTRACT_POINTER_ALIASES` for vendor pointer to exact contract id, and the `-ultrafast` / `-highspeed` / `-fast` speed-tier handling at `:1525-1536`. The requested side comes from `read_delegation_route()` (`src/plugin_bundle/omh/delegation_routing.py:47`).

Surface the verdict as a HUD row field.

## Alternatives considered

Observing the model response or a gateway payload — unnecessary and out of boundary. Hermes already records the answering model per call; OMH reads its own read-only view of that table. No network call, no execution, no Hermes patch.

Reimplementing alias normalization in a new module — rejected; it would drift from `mixture_category_for` on the next model generation.

## Success criteria

- A session whose requested alias and answering model resolve to the same model reports agreement.
[...]

PR #1585, opened 2026-09-15
docs(skills): the body limit bounds the body, not the whole pack

### LINKED ISSUE #1584: ADDING-A-SKILL says the body limit bounds the whole pack; it bounds the always-loaded body only
## Problem

`docs/ADDING-A-SKILL.md` tells a skill author the wrong cost model, in the one section that exists to teach the cost model.

§6 opens: "`FULL_PROFILE_SKILL_BODY_CHAR_LIMIT` bounds what the whole pack costs" (`docs/ADDING-A-SKILL.md:115`).

It does not. `_full_profile_skill_body_chars()` returns `profile["skill_body"]["bytes"]` and nothing else (`src/maintenance/drift.py:202`). The cost payload reports `references` as a sibling key, and the reference files are measured entirely outside the ratchet.

## Complete user goal

An author reading the authoring doc knows which of their bytes are charged against the ratcheted budget and which are not, so the decision "body or reference file" is made on the real cost rather than on a sentence that flattens the two.

## Observed evidence

Measured 2026-09-15 on `origin/main` at `dec82db8`:

```
$ uv run python -P -m omh.cli docs skill-context-cost
## full profile
- Always-loaded skill body: 953716 bytes, 16152 lines, ~238429 tokens
- On-demand references: 85 files, 498849 bytes (~124713 tokens)
```

Two sibling figures. `budget_metrics()` ratchets the first and never reads the second, so roughly a third of what the pack actually ships is outside the limit the doc says bounds "the whole pack".

## Proposal

Correct the sentence to say what the limit bounds — the always-loaded body — and state the consequence that follows, because the consequence is the useful part: content placed in a `references/*.md` costs nothing against the ratchet, and content placed in the body costs the full-profile install a share of every context window.

Derive any figure in the corrected text from the producer rather than writing it beside the claim, per this repo's own rule about counts in prose. `docs claims --check` is the natural gate if the corrected sentence is worth pinning.

## Alternatives considered

Widening the budget to cover reference files so the sentence becomes true — rejected, and it should stay rejected. The separation is correct: an always-loaded byte is paid every turn and an on-demand byte is not, so one ratchet over both would price them identically and remove the incentive that currently pushes optional detail out of the body.

## Success criteria

- §6 states that the limit bounds the always-loaded body, not the whole pack.
- The body-versus-reference cost difference is stated where an author decides between them.
- No figure is written next to the claim; it is derived, or the reader is pointed at the command that prints it.

## Boundaries and risks

Documentation only. The risk of leaving it is not hypothetical: this triage round treated the asymmetry as the organizing fact for a whole wave of capability decisions, and an author who trusts §6 instead would move optional detail *into* the body to keep reference files small — exactly backwards.

Found while triaging 50 capability candidates against the budget (#1556-#1583).

sionic-khope:
## Capability

An author reading `docs/ADDING-A-SKILL.md` §6 now learns which of their bytes are charged against the ratcheted budget and which are not, and what to do about it.

## Motivation

§6 opened with "`FULL_PROFILE_SKILL_BODY_CHAR_LIMIT` bounds what the whole pack costs." It bounds one of the two figures the pack reports.

`_full_profile_skill_body_chars()` returns `profile["skill_body"]["bytes"]` (`src/maintenance/drift.py:202`) and nothing else. `docs skill-context-cost` prints on-demand references as a sibling figure that the ratchet never reads.

Measured on this tree:

```
full profile
  Always-loaded skill body:  953716 bytes  (~238429 tokens)   ratcheted
  On-demand references:      498849 bytes  (~124713 tokens)   not ratcheted
```

About a third of what a `full` install ships sits outside the limit the section said bounds "the whole pack". The two thirds inside it are paid in every context window; the third outside is paid only when a reader opens the file. Different prices for the same byte — and the section that exists to teach the cost model quoted one price for both.

The consequence is why this is worth a PR rather than a footnote: an author who believes the whole pack is bounded will compress a body to stay under the limit, when relocating the same content to a `references/*.md` takes it out of the count entirely. Compression buys back a fraction of one skill's body; relocation buys back all of it.

## Implementation

One hand-written documentation file. §6's opening paragraph now names what the limit bounds, names the producer that decides it, states the per-turn versus on-open price difference, and tells the author to relocate optional detail rather than squeeze it. The density paragraph that followed is unchanged in substance; its subject is restated because the inserted text separated it from its antecedent.

No figure is written beside the claim. The corrected text names the producer and the command that prints the numbers, per this repo's rule that counts in prose drift and then send a reader looking for a string that no longer exists.

No generated artifact, no producer, and no catalog data is touched.

## Observed verification

| Gate | Result |
| --- | --- |
| `docs workflows --check` | rc=0 |
| `docs roles --check` | rc=0 |
| `docs navigation --check` | rc=0 |
| `docs chain-table --check` | rc=0 |
| `docs claims --check --json` | ok |
| `release drift` | No drift. 18 checks passed. |
| `test_documentation_claims.py` + `test_router_content.py` | 140 tests, OK |

Full suite not rerun: this commit touches one hand-written doc and no generated artifact or producer.

## Risks

Documentation only; nothing executes differently. The risk of not landing it is the one described above — an author trusting §6 optimizes in the wrong direction.

Found while triaging fifty capability candidates against this budget (#1556-#1583); every one of those verdicts turned on the asymmetry the section denied.

Closes #1584


sionic-khope:
Adding one gate I had missed at PR time. `CLAUDE.md` on `origin/main` lists `docs skill-sources --check` (the watch-closure gate) and my session's snapshot of that file predated it, so it was not in the verification table above.

Run now on this branch:

```
$ uv run python -P -m omh.cli docs skill-sources --check
rc=0
```

Passes. No change to the branch; recording it so the table is complete rather than leaving the omission unstated.

sionic-khope:
**Your `test-windows (1)` failure is not your branch.** Filed as #1589 with the diagnosis.

The identical test failed on PR #1549 (different author, different branch, nothing in common with this one) with the identical shape:

```
#1549  ... 68836, 'capability_snapshot_digest' ...  !=  ... 68837, ...
#1585  ... 69926, ...                              !=  ... 69927, ...
```

On #1549 the failing commit is provably comment-only against a head that had passed 14/14 — six added lines, all comments, zero deletions — so no executable difference separates the green run from the red one.

**It is not a race.** `test_loop_tool` drives the same action through two adapters and compares the cards. The two loop ids differ by exactly one character:

```
'parity-cli-loop'   15
'parity-tool-loop'  16
```

so any field that is a **length computed over text containing the loop id** differs by exactly 1. `scrub()` replaces the ids inside strings, but a number derived from that text was computed before the substitution.

`_VOLATILE_KEYS` already carries one entry of exactly this class, with its reason attached — `goal_command_sha256`, "Derived from the loop's own handoff command, which names the loop id." A digest over text naming the loop id was recognised; a length over the same text was not.

It is intermittent because shard composition varies run to run (6052 tests in one failing run, 6037 in the other), so whether this test sees a card shape that includes the loop id in that total depends on what preceded it in the shard.

Nothing for you to do here beyond re-running the failed job. #1589 carries the fix direction.

sionic-khope:
This PR's only failing check is not this PR's defect, and it now has a fix in flight.

`test-windows (1)` failed on `test_goal_driver_observe_reaches_the_same_card` with `69926 != 69927` — one field, off by one. `test-windows (0)` passed in the same run, so this is not a fail-fast cancellation rendered as a failure. This branch changes one hand-written documentation file and imports nothing.

Diagnosed and filed as #1590, fixed in #1591: `_VOLATILE_KEYS` in `tests/test_loop_tool.py` lists `observed_at` but not `age_seconds`, which `src/workflows/loop_driver_contract.py:227` derives from that stamp *and* `datetime.now()` at render time. Both cards read the same fixed stamp, so the two ages differ by exactly the gap between the two renders — one of which is a CLI subprocess spawn.

**Deliberately not re-running CI on this branch.** The race is probabilistic, so a re-run has a good chance of coming back green without anything having changed. That would look like evidence and would not be any. This PR should be re-checked after #1591 lands, not before.

`aggregate` fails downstream of that one shard; the other twelve checks pass.

sionic-khope:
**Correction to my earlier comment: the mechanism I gave was wrong.** The failure is real and not your branch — that part stands — but the cause is a clock race, not the loop-id length.

The field is `status_card["driver"]["age_seconds"]` (`src/workflows/loop_driver_contract.py:227`): `int((datetime.now(timezone.utc) - observed_time(stamp)).total_seconds())`. Both cards read the same fixed `observed_at` and each calls `datetime.now()` once, so a second boundary falling between the two renders gives a difference of exactly 1. The large values are seconds since that fixed 2026-09-14 stamp, which is why they drift between runs.

The loop-id-length hypothesis I posted has been falsified by measurement: with the clock frozen, driving the flow with a 16-character and a 15-character loop id moved **0 of the 15 integer leaves** in the card. No length-derived field exists there.

The fix is open as **#1591**, which asserts a producer rule pairing each age with its `observed_at` rather than scrubbing alone. #1589 is closed in its favour.

Apologies for the wrong pointer — if you went looking at loop ids on my say-so, that was my error.

PR #1586, opened 2026-09-15
feat(setup): export a configured OMH profile and apply it elsewhere

### LINKED ISSUE #1558: [Feature]: Export and apply a configured OMH profile as a local, secret-free pack
## Problem

A person who has configured an OMH profile — executor, operating model, memory mode, model aliases, MCP recipe, capability policy — cannot move it. There is no export. Reproducing the setup on a second machine or handing it to a teammate means retyping `omh setup` flags from memory.

## Complete user goal

Serialize a configured profile into one local, secret-free artifact, and apply that artifact on another machine or profile so the second install matches the first, with a readable report of what applied, what differed, and what was deliberately not carried.

## Intended surfaces

Hermes chat or wrapper; all executors (executor-neutral).

## Observed evidence

`omh setup` (`src/commands/setup.py:4927-5051`) already configures every ingredient: `--profile`, `--default-executor`, `--operating-model`, `--memory-mode`, `--model-alias` / `--apply-model-config`, `--with-mcp` / `--mcp-host`, `--profile-pack`. What is missing is the other direction. `TeamProfilePack` (`src/profiles/team.py:11`) is hand-authored with `install_command` pointing back at `omh setup --profile-pack <id>` (`:52`); `omh setup profile list|inspect` (`:5214-5222`) only reads the built-ins. Nothing produces a pack from a machine.

The only import-shaped commands in the tree import something else: `omh ops rules-import` (`src/commands/rules_import.py:38`) takes third-party agent rule files, and `omh convert` (`src/commands/setup.py:5078`) takes a skills directory.

## Proposal

One serializer that writes `setup_profile/v1` plus the MCP recipe, model aliases, and capability policy to a local JSON artifact; one applier that reuses `omh setup`'s existing writers; a redaction denylist reusing the `metadata_safety` / `RAW_OR_HIDDEN_KEYS` primitives in `src/system/append_only_store.py`; and a manifest. CLI and library only — no catalog registration, no exact-count churn.

## Alternatives considered

Extending `TeamProfilePack` — rejected; those are authored product content, not a machine's state.

## Success criteria

- Exporting a configured profile and applying it to an empty profile produces the same configured state, and the report names each applied field.
- A secret present in the source config is absent from the artifact, and the export says which fields it withheld rather than silently dropping them.
- Applying a pack built by a newer OMH reports the version mismatch instead of writing a partial profile.
- A field the target cannot accept is reported as not applied, with the reason.

## Boundaries and risks

**Transport is out of scope and must stay out.** Moving the pack between machines is the person's own git or file transfer. OMH makes no network calls (`CONTEXT.md:15-16`); a pack that fetches or pushes itself would cross that line.

Recurring work is not part of a profile — it is a record under `.omh/hermes-ops/`, not config — so there is no cron to bundle. The original proposal named cron; that half does not exist to export.

sionic-khope:
### What Changed

- New `omh setup-profile export` serializes this machine's configured OMH profile — the `setup_profile/v1` payload (coding owner, operating model, memory mode, selected categories), the capability policy, the recorded MCP host recipe, and the Hermes model aliases — into one local JSON pack (`omh_setup_profile_pack/v1`) with a content-addressed manifest and an explicit `withheld` list.
- New `omh setup-profile apply --from <pack.json>` writes that pack back through `omh setup`'s own writers. Dry run by default; `--apply` writes.
- `parse_hermes_model_aliases()` in `src/maintenance/advisory.py` is promoted from private so the pack reads aliases through the existing parser instead of a second one.

Closes #1558.

### Why This Exists

`omh setup` already configures every ingredient of a profile and nothing serialized any of it back out. Reproducing a configured machine on a second box — or handing the setup to a teammate — meant retyping `omh setup` flags from memory. There was no export in the tree; the only import-shaped commands import something else (`omh ops rules-import` takes third-party agent rule files, `omh convert` takes a skills directory).

`TeamProfilePack` was deliberately not extended. Those packs are authored product content whose `install_command` points back at `omh setup --profile-pack <id>`, and `omh profile list|inspect` only reads the built-ins. A machine's own state is a different thing and gets its own schema and its own commands.

### User / Operator Impact

Before: a configured profile lived on exactly one machine.

After:

```sh
omh setup-profile export --output profile-pack.json     # configured machine
omh setup-profile apply --from profile-pack.json        # other machine: preview
omh setup-profile apply --from profile-pack.json --apply
```

The apply report names every field as applied or not applied, with the reason for each refusal and the exact `omh setup` command for the two surfaces it will not write unasked.

### How It Works

**Two boundaries that are the point of the design, not incidental:**

- **Transport stays out.** Export writes a local file; apply reads a local file. Moving one between machines is the person's own git or file copy. There is deliberately no `--from <url>` — a pack that fetched or pushed itself would make core OMH a network client (`CONTEXT.md`, no LLM/API/network calls).
- **Cron is not in a profile.** The original proposal named cron settings; recurring work is a record under `.omh/hermes-ops/`, not config, so there is nothing to bundle and none was invented.

**Redaction reuses the existing primitives.** `RAW_OR_HIDDEN_KEYS` from `src/system/append_only_store.py` and `is_secret_value_shaped` / `is_sensitive_metadata_text` from `src/system/metadata_safety.py` — no second denylist that could drift from the one every record store screens against. A field is withheld when its name marks credential material or a raw/hidden payload, when its value is credential-shaped, or when its value is a machine-local path or link. **Every withheld field is named with its reason** in the pack, the export report, and the apply report; a silent drop is treated as the defect, not as redaction working.

**Two refusals that behave differently.** A pack whose schema generation this install does not know is refused whole before the first write, and the message distinguishes "written by a newer OMH" from a merely unrecognized name — an unknown name is not evidence the other side is ahead. Past that gate, a field this install cannot accept (an executor or operating model a newer OMH shipped) is reported and skipped while everything else lands: one unknown name should not cost the person every other field they configured.

**Partial lists report honestly.** When some entries of `selected_categories` or `disabled_families` are acceptable and some are not, the row reports the write it made and names only the entries that fell out. `not_applied` beside a value that did land would send a reader looking for a write that already happened.

**Carried, reported, deliberately not written:**

| Field | Why not |
| --- | --- |
| `model_aliases` | The write goes through `omh setup --model-setup --apply-model-config`, which binds it to a Hermes config digest and an explicit confirmation a pack apply does not carry. The report hands back the fully-formed command. |
| `mcp.host_config` | Edits another product's config file (Codex `config.toml`, Claude Code `.claude.json`, …). Written only under `--with-mcp`, through the same `install_mcp_host_config` writer `omh setup --with-mcp` uses; `--mcp-config-path` targets an explicit file. |
| `org_rule_source_policy` | Names machine-local file paths the export withheld. |
| `parallelism` | `omh setup` accepts no parallelism flag. The row appears only when the source machine had actually edited the block, and names the differing values. |

- `src/profiles/pack.py` (new) — serializer, redaction walk, version gate, applier. Schemas: `omh_setup_profile_pack/v1`, `…_manifest/v1`, `…_mcp/v1`, `…_model_aliases/v1`, `…_export/v1`, `…_apply/v1`.
- `src/commands/setup_profile_pack.py` (new) — the `omh setup-profile export|apply` parsers and renderers.
- `src/commands/setup.py` — registers the new group (deferred import, same reason as `capability-policy`).
- `src/maintenance/advisory.py` — `_parse_hermes_model_aliases` → `parse_hermes_model_aliases`, with a docstring naming why it is public.
- `tests/test_setup_profile_pack.py` (new) — 34 cases.
- `docs/SETUP-PROFILE-PACK.md` (new), linked from `docs/README.md`; `CHANGELOG.md` Unreleased bullet.

No catalog registration, no generated-artifact change, no exact-count churn.

### Affected Surfaces

- [x] Hermes chat or wrapper
- [ ] Codex
- [ ] Claude Code
- [ ] Generic executor
- [x] Not executor-specific

## Validation

- [x] `git diff --check`
- [x] `omh --help`
- [ ] `omh … release hermes-smoke --install-path setup --include-command-smoke`
- [ ] Relevant workflow-learning review queue smoke — no learning contract changed
- [x] Relevant dry-run or smoke command: `omh setup-profile export`/`apply` round trip against temporary homes
- [ ] Manual Hermes/TUI check

### Observed Evidence

- `tests/test_setup_profile_pack.py`: 34 tests, OK. Each of the four acceptance criteria has its own case, plus the redaction screens, the partial-list semantics, the file reader, and the CLI surface.
- `tests/test_cli.py` + `tests/test_architecture_layout.py` + `tests/test_agent_board_kanban.py`: 405 tests, OK (the kanban file is the stale-editable-install canary).
- Nine docs byte gates through `-P`, all OK: workflows, roles, navigation, claims, chain-table, capability-families, ulw-inventory, ulw-site, agent-skills.
- `omh release drift`: No drift, 18 checks passed.
- `ruff check src tests`: all checks passed. `compileall -q src tests`: clean. `git diff --check`: clean.
- Manual round trip on temporary homes: a source profile carrying `github_token=ghp_…` and an `AKIA…` value exported with neither string anywhere in the artifact and both named in `withheld`; applying to an empty profile reproduced executor/operating model/memory mode/categories/disabled families exactly.

- **Full suite**: running at the time of this PR against a machine with 12 concurrent sibling worktree suites; the result is posted as a comment when it lands. The baseline on `origin/main` today is 12184 tests, OK (skipped=4).
- **Live Hermes install**: the model-alias section is exercised against a config-file fixture, never a real `~/.hermes`.
- **Host-default MCP config paths**: every test names an explicit `--mcp-config-path`, so no test can touch a developer's real `~/.claude.json` or `~/.codex/config.toml`. The default-path branch is `install_mcp_host_config`'s own, unchanged by this PR.
- **Windows**: the machine-local path screen is pattern-matched (drive letters and UNC prefixes included) and was not exercised on a Windows filesystem.
- **`release hermes-smoke --install-path setup --include-command-smoke`** and the manual Hermes/TUI check: this change adds no installed artifact, no skill, no plugin surface, and no TUI surface.

## Risk

- The redaction walk is conservative by shape, so a hand-added profile field whose value begins with `/` or `~` is withheld as machine-local even if it was portable. It is named in `withheld`, so it is visible rather than lost silently, and the person can set it on the target.
- `parse_hermes_model_aliases` is now public. Its one existing caller and its behavior are unchanged; only the name moved.
- `write_setup_profile` rebuilds the whole profile, so applying a pack resets any profile block the pack does not drive. The two such blocks today (`parallelism`, `org_rule_source_policy`) each get an explicit `not_applied` row naming what did not carry.

## Compatibility / Rollout

- Purely additive: one new command group, one new module, one function rename that is internal to `src/`. No existing payload, schema, or generated artifact changes.
- `omh_setup_profile_pack/v1` is the first generation; the applier already refuses a higher one by name, so a future `v2` degrades to a clear refusal on older installs rather than a partial write.
- Existing installs need `omh update` before `omh setup-profile` appears.

## Release / Claims

- [x] Release-channel impact considered (`stable`)

## Follow-Up

- If `parallelism` or the org rule source ever gains an `omh setup` flag, the two `not_applied` rows should become applied fields in the same change.


sionic-khope:
Additional gate, run after the PR was opened (it was missing from my original command list): the skill-source watch-closure gate.

```
$ uv run python -P -m omh.cli docs skill-sources --check
Skill-source closure: PASS
  rows=38 closed=0 held=0 not_applicable=38 receipts=0
```

Exit 0 on head `6ddf1217`. All 38 rows are `not_applicable / pre_receipt_baseline`, unchanged by this PR — it adds no skill and touches no `SKILL-SOURCES` row.

sionic-khope:
Independent read of the branch by the session that wrote #1558, recording what is already confirmed so a reviewer does not re-derive it. Not an approval.

The two boundaries the issue put out of scope hold:

- **Transport.** `grep` for `urllib|requests|http|socket|urlopen|fetch` across `src/profiles/pack.py` and `src/commands/setup_profile_pack.py` returns two hits, both prose saying OMH does not fetch (`pack.py:20`, and the `--from` help text: "OMH never fetches one; copy it here yourself"). Export writes a local file, apply reads one.
- **Cron.** Not present, not mentioned. Correct — recurring work is a record under `.omh/hermes-ops/`, not profile config, so there was nothing to bundle.

On the criterion most likely to be satisfied only on paper — a secret being absent *and* announced:

- Redaction imports `RAW_OR_HIDDEN_KEYS` and the `metadata_safety` predicates rather than restating them, so there is no second denylist to drift.
- `withheld` is not a bare flag. Four distinct reasons each name the property that makes the value unsafe or useless elsewhere (raw-or-hidden payload, credential-marked name, credential-shaped value, machine-local path), with a comment saying why a bare "withheld" would be inadequate: the reader cannot tell whether to go find the value by hand.
- The version gate refuses before the first write, and six tests assert the target file does not exist afterward — so "refused whole" is demonstrated rather than asserted.

Two things I did not check and a reviewer should: whether `parse_hermes_model_aliases` being promoted from private changes anything for its existing caller in `advisory.py`, and whether the set of fields reported not-applied (`parallelism`, `org_rule_source_policy`, aliases, MCP) is complete against what `omh setup` can actually configure — an omission there is silent by construction, which is the failure mode this feature exists to prevent.

sionic-khope:
**Only one job actually failed, and it found two real defects — neither is a flake.**

`gh pr checks` shows three `test` shards red. Only `test (3.11, 1)` failed; the other two were **cancelled** by fail-fast (the `test` matrix uses the default `fail-fast: true`; only `test-windows` sets it false, `ci.yml:177`). Separate `cancelled` from `failure` via `gh run view <id> --json jobs` before reading breadth.

The one failing job:

```
FAIL:  test_credential_fixture_policy.CredentialFixturePolicyTests.test_no_aws_access_key_literal_is_stored_in_the_tree
       AssertionError: Lists differ: ['tests/test_setup_profile_pack.py'] != []

ERROR: test_paired_run_execution_fan_in.PairedRunExecutionFanInTests.test_complete_authenticated_matrix_builds_deterministic_observed_decision

Ran 6082 tests — FAILED (failures=1, errors=1, skipped=10)
```

The first is the one to look at first. `tests/test_setup_profile_pack.py` now holds something the credential-fixture policy reads as an AWS access key literal. That gate exists to keep key-shaped strings out of the tree regardless of whether the value is real, so the fix is a placeholder that does not match the pattern, not an exemption.

I have not looked at your diff and am not reviewing the PR — this is only to save you diagnosing breadth that is not there. Unrelated to the `test-windows (1)` parity flake I commented about on #1585, which is #1589.

sionic-khope:
**Correction to my earlier comment: the mechanism I gave was wrong.** The failure is real and not your branch — that part stands — but the cause is a clock race, not the loop-id length.

The field is `status_card["driver"]["age_seconds"]` (`src/workflows/loop_driver_contract.py:227`): `int((datetime.now(timezone.utc) - observed_time(stamp)).total_seconds())`. Both cards read the same fixed `observed_at` and each calls `datetime.now()` once, so a second boundary falling between the two renders gives a difference of exactly 1. The large values are seconds since that fixed 2026-09-14 stamp, which is why they drift between runs.

The loop-id-length hypothesis I posted has been falsified by measurement: with the clock frozen, driving the flow with a 16-character and a 15-character loop id moved **0 of the 15 integer leaves** in the card. No length-derived field exists there.

The fix is open as **#1591**, which asserts a producer rule pairing each age with its `observed_at` rather than scrubbing alone. #1589 is closed in its favour.

Apologies for the wrong pointer — if you went looking at loop ids on my say-so, that was my error.

PR #1587, opened 2026-09-15
feat(ops): rehearse a planned batch against the local rules before it runs

### LINKED ISSUE #1557: [Feature]: Rehearse a planned command batch against local rules before unattended work
## Problem

Before a cron entry or an unattended batch starts, nobody can see which of its planned commands will be allowed, which will stop for approval, and which will be refused. The person finds out one command at a time, while the run is already going.

`automation-blueprint` plans the schedule and the failure policy but not the approval outcome, and its own safety rule (`src/skills/catalog_definitions.py:5338`) says a prepared failure policy is not enforcement.

## Complete user goal

Given a list of commands a scheduled or unattended run intends to issue, return a table saying, per command, whether the local rules refuse it, whether an approval bypass is currently active, and — where OMH cannot tell — that it cannot tell, before the run starts.

## Intended surfaces

Hermes chat or wrapper; generic executor.

## Observed evidence

`omh ops toolcall-rules-validate` (`src/commands/toolcall_rules.py:17-28`) validates the rules file's syntax only. There is no path that runs a list of planned calls through the rules. Read from source; no rehearsal has been run because none exists.

## Proposal

Add a batch rehearsal over the pure matcher that already exists. `toolcall_rule_directive(tool_name, tool_input, ...)` (`src/plugin_bundle/omh/toolcall_rules.py:156-185`) is side-effect-free and already returns a block verdict for one `(tool, args)` pair. Run it over N planned pairs and render the table, plus the bypass state from `_approvals_mode_off` (`src/plugin_bundle/omh/approval_bypass.py:129-160`) and `is_approval_bypass_active`.

Ship it as an `omh ops` subcommand. No skill registration.

## Alternatives considered

A new skill — rejected. The phrasing already routes to `automation-blueprint` at high/28.0, so a new skill buys a trigger collision with always-loaded body cost, for a check that belongs on the operator command surface.

## Success criteria

- A planned batch containing a command the local rules refuse reports that command as refused, naming the rule.
- A batch under an active approval bypass reports the bypass, because a rehearsal that ignores it describes a policy nobody is currently applying.
- A command whose approval tier cannot be determined reports `unknown` and the reason, and the summary never counts it as allowed.
- The rehearsal writes nothing and executes nothing.

## Boundaries and risks

**Scope limit, to be honored rather than worked around: only two of the three tiers are derivable today.** Refusal comes from the toolcall rules; bypass comes from `approvals.mode` plus the session row. "Needs approval" requires Hermes' per-tool approval declarations, and no reader exists — `approval_bypass.py:156` states outright that a deeper `mode:` such as `approvals.tools.mode` is a different key it does not read. Claiming all three tiers would be a claim-boundary violation. The third tier must render as `unknown`.

`APPROVAL_RULE_TABLE` (`src/system/approval_tier.py:67-107`) is about OMH's own operations, not Hermes tool calls; `resolve_approval_tier` (`:123`) is a reusable pure function but its six rows do not answer this question.

Safety verdicts must come from declared fields, never inferred from user message text — `src/quality/safety_preflight.py:11-16` forbids that pattern.

sionic-khope:
### What Changed

- New operator command `omh ops permission-rehearse --plan <file>`: it takes the
  `(tool, args)` pairs a scheduled or unattended run intends to issue and returns
  a per-call verdict table, the state of the local rules file, and the observed
  approval-bypass state — before the run starts. Nothing is executed and nothing
  is written.
- `matched_toolcall_rule()` split out of `toolcall_rule_directive()` in the
  plugin bundle. Behavior of the enforcing `pre_tool_call` path is unchanged;
  the directive builder is now a thin wrapper over the matcher.
- `docs/TOOLCALL-RULES.md` gains a "Rehearsing a planned batch" section, linked
  from the docs index.

### Why This Exists

Closes #1557.

`omh ops toolcall-rules-validate` validates the rules file's *syntax*. It does
not answer the question an operator actually has before arming a cron entry:
which of these commands will be refused? There was no path that ran a list of
planned calls through the rules, so the answer arrived one command at a time,
while the run was already going. `automation-blueprint` plans the schedule and
the failure policy, and its own safety rule says a prepared failure policy is
not enforcement.

### User / Operator Impact

Before a batch runs, the operator can see which planned calls a rule refuses and
which rule refuses each one; whether an approval bypass is currently active, so
they are not reading a policy nobody is applying; and — where OMH cannot tell —
that it cannot tell, with the reason. A wrapper can gate on the exit status.

### How It Works

Refusal verdicts come from `matched_toolcall_rule` in
`src/plugin_bundle/omh/toolcall_rules.py`, the same matcher the enforcing
`pre_tool_call` hook calls. Reuse is the point: a rehearsal with its own match
loop could disagree with the enforcement it claims to predict. The matcher was
split out so the rehearsal can read the matched rule object rather than parsing
the rule name back out of the block message.

The batch is rehearsed as one fresh session in plan order, so a
`repeat: "once"` rule refuses the first matching call and the rest proceed past
it — what the hook would do to that batch. The synthetic session id is unique
per invocation, so a rehearsal can never consume a real session's once-claims.

Bypass state comes from `effective_approval_bypass` (`approvals.mode` in
`config.yaml`, then the live TUI session row's persisted `/yolo` flag, then the
hook-observed ledger — which is where `is_approval_bypass_active` reaches a
process that is not the host).

**The scope limit, stated rather than worked around.** Only two of the three
tiers are derivable today. "Needs approval" lives in Hermes' per-tool approval
declarations and OMH has no reader for them — `_approvals_mode_off` says outright
that a deeper `mode:` such as `approvals.tools.mode` is a different key it does
not read. So the verdict vocabulary is closed at two values, `refused` and
`unknown`, and the summary counts `planned`, `refused`, and `unknown`. There is
no allowed bucket. That is the shape rather than an accident: an undeterminable
row cannot be miscounted as permission when no count can hold it. A negative
control pins that `approvals.tools.mode: off` does not read as the global
bypass.

The plan document fails **closed**, unlike the rules file it is rehearsed
against. The rules loader fails open because a broken rules file must not break
the hook that feeds the model; a plan OMH could not read whole is not a
rehearsal of that plan, so an oversized or malformed plan is refused, never
truncated. For the same reason the payload carries `rules.defect_count`: a rules
file that is present but defective makes the refusal tier answer from a policy
the operator did not write.

Exit codes, decided deliberately against `tests/test_exit_code_truthfulness_policy.py`
and with the reasoning written at the mapper:

| Code | Meaning |
| --- | --- |
| `0` | No rule refuses any planned call. Never a claim that they are allowed. |
| `1` | At least one planned call is refused — an unattended batch that would be blocked partway through should not start. |
| `2` | The command could not run (`OmhError`, e.g. an unreadable plan). |
| `3` | The rules file is present but defective. Not 2, because `main` already returns 2 for an `OmhError` and a wrapper must be able to tell the two apart. |

`unknown` deliberately does not move the status: every call no rule refuses is
unknown today, so a status that could never be 0 would teach a wrapper to stop
reading it. The payload carries the "not allowed" half instead.

No skill registration, per the issue: the phrasing already routes to
`automation-blueprint` at high/28.0, and a new skill would buy a trigger
collision plus always-loaded body cost.

- `src/workflows/permission_rehearsal.py` (new) — pure builder, plan parsing,
  `omh_permission_rehearsal/v1` and `omh_permission_rehearsal_plan/v1`.
- `src/commands/permission_rehearsal.py` (new) — the `ops` subcommand and
  `_permission_rehearsal_exit_code`.
- `src/commands/ops.py` — registration.
- `src/plugin_bundle/omh/toolcall_rules.py` — `matched_toolcall_rule` split out.
- `tests/test_permission_rehearsal.py` (new), `docs/TOOLCALL-RULES.md`,
  `docs/README.md`, `CHANGELOG.md`.

### Affected Surfaces

- [x] Hermes chat or wrapper
- [ ] Codex
- [ ] Claude Code
- [x] Generic executor
- [x] Not executor-specific

## Validation

- [x] `git diff --check`
- [ ] `omh --help`
- [ ] `omh --omh-home /tmp/omh-smoke --hermes-home /tmp/hermes-smoke release hermes-smoke ...`
- [ ] Relevant workflow-learning review queue smoke — no learning contract changed
- [x] Relevant dry-run or smoke command: the new command run end to end against a
      temporary OMH/Hermes home, with and without a rules file and with
      `approvals.mode: off`
- [ ] Manual Hermes/TUI check — no Hermes surface changed; the plugin-bundle edit
      is a refactor with unchanged enforcing behavior, covered by
      `tests/test_toolcall_rules.py`

### Observed Evidence

- `tests/test_permission_rehearsal.py` — 35 cases, OK. Covers both tiers, the
  rule-scoped negative case, `repeat` once/always across a batch, the
  `approvals.tools.mode` negative control, fail-closed plan parsing, a defective
  and an unparseable rules file, all exit-code lanes, and two no-write snapshots
  of both homes.
- `tests/test_toolcall_rules.py` (24) + `tests/test_exit_code_truthfulness_policy.py`
  + the new file: 64 tests, OK. The truthfulness gate re-derives the new mapper
  and passes.
- `tests/test_cli.py` — 359 tests, OK.
- `tests/test_agent_board_kanban.py` — 40 tests, OK (the stale-editable-install
  canary; the tree was resynced after the new modules were created).
- `uv run python -m omh.cli release drift` — `No drift. 18 checks passed.`
- `docs workflows/roles/navigation --check` — pass. `docs navigation --check`:
  PASS, roots=2 scope=72 reachable=65 exempt=7.
- `uv run python -P -m omh.cli docs skill-sources --check` — `PASS`, rows=38,
  closed=0, held=0.
- `ruff check src tests` — All checks passed. `compileall` — clean.
  `git diff --check` — clean.

- The local full-suite rerun was still running at commit time. Four
  implementations were running full suites on one 14-core machine
  (load average ~24), so the local run is contended and slow; **CI on this
  branch is the uncontended signal**. Baseline on origin/main measured today by
  the coordinating agent: 12184 tests, OK (skipped=4).
- `capability-families --check`, `harness validate`, `release checklist`,
  `release hermes-smoke`: no skill, capability family, harness contract, or
  release surface changed; `release drift` (18 checks) covers the generated
  artifacts this change could have moved, and it reports no drift.

## Risk

- The plugin-bundle refactor is the only edit to an installed artifact. It is
  behavior-preserving — `toolcall_rule_directive` produces the same directive
  from the same matcher — and existing rule tests pass unchanged. A live machine
  picks it up through `omh update` like any other bundle change; no current
  behavior depends on the split.
- The rehearsal models a fresh session for `repeat: "once"` rules. If the batch
  will run inside a session that has already fired one of those rules, the
  rehearsal will show a refusal the live session would skip. The payload and the
  docs both state the model.
- `rules.defect_count` re-reads and re-validates the rules file. The bound is the
  existing `MAX_RULES_FILE_BYTES`; a file over it counts as one defect, matching
  what the loader does with it.

## Compatibility / Rollout

- Additive. A new `ops` subcommand and a new pure module; no existing command,
  schema, or generated artifact changes shape.
- `omh update` is needed on a live machine only for the plugin-bundle refactor,
  and nothing observable changes if it is not run.

## Release / Claims

- [x] Release-channel impact considered (`local` operator surface)
- [x] Evidence contains no credentials, raw prompts, raw platform events, transcripts, or unrelated private data — the rehearsal row names the tool and the matched rule and never echoes the planned call's arguments

## Follow-Up

- The third tier stays `unknown` until a reader for Hermes' per-tool approval
  declarations exists. That is Hermes-side work and deliberately out of scope
  here; when it lands, the verdict vocabulary and `reason_codes` are the two
  places that extend.

Closes #1557


sionic-khope:
Independent read of the branch by the session that wrote #1557, recording what is already confirmed so a reviewer does not re-derive it. Not an approval.

The issue's defining constraint was that only two of the three approval tiers are derivable, and that claiming the third would be a claim-boundary violation. The branch does not merely satisfy that, it makes the violation unavailable:

- `PERMISSION_REHEARSAL_VERDICTS` is `(refused, unknown)`. There is no `allowed` verdict to mis-assign.
- The summary counts `unknown` as `len(calls) - refused` with the comment saying why — there is no bucket an unknown could be folded into.
- With no rules file, every row is `unknown` and `rules.present` is `false`. "Nothing refused" therefore cannot read as "safe to run unattended" on a machine that never opted in, which was the likeliest way this surface could have lied.
- `test_a_deeper_mode_key_is_not_the_global_bypass` pins the scope limit as a negative control, against `approvals.tools.mode` — the key `approval_bypass.py:156` says outright it does not read.

The exit-code mapper's docstring answers the question `test_exit_code_truthfulness_policy.py` exists to ask, including why 3 rather than 2 (`main` already returns 2 for an `OmhError`, so a wrapper reading 2 could not distinguish an unparseable plan from a rehearsal that ran).

Two things I did not check and a reviewer should: whether `MAX_PLANNED_CALLS` at 256 and `MAX_PLAN_FILE_BYTES` at 256 KiB are the right bounds for a real unattended batch, and whether the refused-outranks-defect ordering is right when both are present — the branch argues it and the argument is written down, but it is a judgment call, not a derivation.

sionic-khope:
CI settled: **14/14 pass**, zero non-SUCCESS rows, nothing cancelled — so the fail-fast pattern that can make two failures read as four does not apply here; every test-bearing job ran to completion and reported its own result.

Verified independently of the authoring session by `gh pr checks 1587 --json name,state`.

PR #1588, opened 2026-09-15
feat(hud): attest the model that answered against the model requested

### LINKED ISSUE #1556: [Feature]: Warn when the model that answered is not the model that was requested

sionic-khope:
### What Changed

- Every Hermes-native delegation row in the HUD payload now carries a
  `model_attestation` verdict: `agreement`, `disagreement`, or `unknown`,
  comparing the model the run was routed to against the model Hermes recorded
  as having answered it.
- `served_model_attestation()` is the comparison, in
  `src/plugin_bundle/omh/hermes_delegation.py`. It resolves both sides through
  the alias machinery the mixture-category label already uses.
- `_model_alias_candidates()` is extracted out of `mixture_category_for()` and
  shared by both readers, so alias resolution exists once.

### Why This Exists

Closes #1556. A run could be routed to one model and answered by another, and
nothing compared the two — an alias that silently resolves elsewhere, a
provider substitution, or a version drift was invisible to the person and to
every downstream report naming the model.

Hermes already records both facts independently:

- `sessions.model` is what was asked for. Read from Hermes' own source
  (`hermes_state_sessions.py`), it is written at session creation and on an
  explicit `/model` switch — never from a provider's answer.
- `session_model_usage.model` is written per call from the model that actually
  answered. OMH's plugin already reads that table read-only.

`_observed_wire_model()` returned the recorded value when present and fell back
to the usage value otherwise. The recorded-first ordering means the row's
`model` column shows what was requested, so the discrepancy was hidden in
exactly the case it existed.

### User / Operator Impact

A person reading the HUD payload (or any consumer of `read_omh_hud`) can now
see that a child labelled `gpt-6-astra` was in fact answered by
`claude-fable-5-1`, instead of seeing only the requested name. When the two
agree, the row says so; when the session cannot be judged, it says `unknown`
with the reason rather than implying agreement.

### How It Works

Requested side, per row:

1. the child's own `sessions.model` when Hermes recorded one
   (`requested_source: session_model`);
2. otherwise `read_delegation_route()` — `delegation.model` in `config.yaml`,
   what the next `delegate_task` dispatch resolves
   (`requested_source: delegation_route`). It is read once per payload and only
   when some child needs it.

Answering side: the single distinct `session_model_usage.model` for that child,
already aggregated by the existing query.

Both sides resolve through `_model_alias_candidates()` — `_dated_snapshot_base`
for date-pinned snapshots, `_projected_model_alias` for reasoning-mode and
service-tier projection, `EXACT_CONTRACT_POINTER_ALIASES` for vendor pointer to
exact contract id, and the `-ultrafast` / `-highspeed` / `-fast` speed-tier
handling. So `deepseek-flash` answered by `deepseek/deepseek-v4.1-flash`,
`gpt-6-astra` answered by `gpt-6-astra-2026-07-09`, and two speed tiers of one
model all read as agreement rather than a false alarm. Because the list is the
same one the category label matches chains against, a newly declared pointer
alias or dated snapshot moves both surfaces with no second copy to update — a
test proves this by declaring one more pointer row and watching the verdict
flip from `disagreement` to `agreement`.

When both sides agree on a name neither of them spells (an `-ultrafast` route
answered by a `-fast` one), the verdict names the shared identity as
`resolved`; when either side already spells it, no third name is added.

`unknown` is a first-class verdict, never a failure and never a silent
agreement:

- `multiple_usage_models` — the child changed models mid-run.
  `_sole_distinct_value()` returns an empty string for that case, exactly as it
  does for "no rows", and reading either as a match is the trap.
  `_distinct_value_count()` keeps the two apart.
- `no_usage_observation` — no `session_model_usage` row named a model.
- `no_requested_model` — nothing recorded what was asked for.

Coverage is stated in the output, not papered over: each verdict carries
`coverage: delegation_child`, and the payload carries the sentence explaining
that a main session is named by `sessions.model` alone, with no per-call
answering model to attest against. `read_omh_hud` carries that sentence
alongside the native rows it merges, and only when it merges them.

- `src/plugin_bundle/omh/hermes_delegation.py` — `_model_alias_candidates()`
  (extracted), `_distinct_value_count()`, `_shared_model_aliases()`,
  `served_model_attestation()`, `ATTESTATION_COVERAGE`,
  `ATTESTATION_COVERAGE_CLAIM`; the usage aggregate gains `model_count`; the
  row gains `model_attestation`; the payload gains `attestation_coverage`.
- `src/plugin_bundle/omh/runtime_reader.py` — carries
  `attestation_coverage` into `payload["subagents"]` when native rows merge.
- `tests/test_plugin_hermes_delegation.py` — new `ServedModelAttestationTest`
  (12 cases) plus two assertions on the existing `read_omh_hud` merge test.
- `CHANGELOG.md` — one `## Unreleased` bullet, appended at the end of the list.

No new dependency. Read-only throughout: the same `session_model_usage` table
the plugin already queries, plus the config the route writer already reads. No
network, no execution, no Hermes modification.

### Affected Surfaces

- [x] Hermes chat or wrapper
- [ ] Codex
- [ ] Claude Code
- [ ] Generic executor
- [x] Not executor-specific

## Validation

- [x] `uv run python -P -m omh.cli docs skill-sources --check`
- [x] `uv run python -m omh.cli release hermes-smoke` (plan mode only — see
      Not Tested)
- [x] `git diff --check`
- [ ] Manual Hermes/TUI check — see Not Tested

### Observed Evidence

- `PYTHONPATH=tests uv run python -m unittest discover -s tests` —
  **Ran 12196 tests in 1225.8s, OK (skipped=4)**. The baseline measured on
  `origin/main` today was 12184, OK (skipped=4); the difference is exactly
  the 12 cases this PR adds. No test changed status.
- `PYTHONPATH=tests uv run python -m unittest tests/test_plugin_hermes_delegation.py`
  — 86 tests, OK (12 new attestation cases; the file held 74 before).
- HUD suites (`test_hud`, `test_hud_payload_budget`,
  `test_hud_conversation_scope`, `test_subagent_graph_hud_bounds`,
  `test_hud_reader_security`) — 94 tests, OK.
- All `tests/test_plugin_*.py` — 305 tests, OK.
- Neighbouring delegation/model suites (`test_delegation_continuity`,
  `test_coding_delegation`, `test_model_contracts`, `test_tui_model_widget`,
  `test_provider_entitlements`, `test_delegate_route_tool`,
  `test_model_price_overrides`, `test_provider_detection`,
  `test_provider_truth`) — 248 tests, OK.
- Count-pinning suites (`test_hermes_ux_quality`, `test_release_smoke`,
  `test_routing_precision`) — 42 tests, OK. No routing case, skill, or demo
  card was added, so no exact-count fixture moved.
- `tests/test_agent_board_kanban.py` — 40 tests, OK. Run deliberately as the
  stale-editable-install canary before trusting any of the numbers above.
- `uv run python -m omh.cli release drift` — No drift, 18 checks passed.
- `uv run python -m omh.cli docs claims --check --json` — `ok: true`,
  5 supported / 1 not_run (model evaluation disabled).
- `docs workflows --check` (209 tap skills, no drift), `docs roles --check`,
  `docs chain-table --check`, `docs navigation --check` (PASS, 65 reachable /
  7 exempt), `docs capability-families --check` — all ok.
- `served_model_attestation()` run over this machine's recorded Hermes data
  (822 sessions), read-only and immutable — see
  https://github.com/rlaope/oh-my-hermes/pull/1588#issuecomment-5676847938
  for method and counts. Observed by the reviewing session, not by this
  branch's own test run.
- The premise this feature rests on — that `sessions.model` is never written
  from the model that answered — was independently checked against Hermes'
  own source: https://github.com/rlaope/oh-my-hermes/pull/1588#issuecomment-5676485184
- `uv run python -P -m omh.cli docs skill-sources --check` —
  `Skill-source closure: PASS`, exit 0.
- `uv run --group lint ruff check src tests` — All checks passed.
- `uv run python -m compileall -q src tests`, `git diff --check` — clean.
- Row size measured on a fixture disagreement row: 570 bytes, of which the
  attestation is ~180. `MAX_WIDGET_HUD_BYTES` is 60,000 and
  `ACTIVITY_ROW_LIMIT` is 8, so the addition stays well inside the widget
  budget; `tests/test_hud_payload_budget.py` passes unchanged.
- `UV_NO_SYNC=1 uv run python -P -c ...` against the installed editable
  generation returns the new verdict, so none of the numbers above are an
  artifact of a stale editable install.

- **No attestation has been observed against a real disagreement.** The
  function has since been run over this machine's recorded Hermes data —
  822 sessions, read-only and immutable, with the function receiving two
  strings and never touching the database. Numbers and method are in
  https://github.com/rlaope/oh-my-hermes/pull/1588#issuecomment-5676847938;
  they are not restated here so they live in one place. What that run
  establishes: the comparison runs against real recorded data without error,
  alias resolution does not over-report (hundreds of real sessions resolved to
  agreement with zero false disagreements, which is the failure mode a
  normalizer gets wrong), and both `unknown` bases occur in reality rather than
  only in fixtures — including `multiple_usage_models`, so the mid-run switch
  this PR pins is a real case and not a hypothetical one. What it does not
  establish: that the detector catches a true substitution. There were none in
  that corpus to catch, and zero disagreements on a healthy machine is the
  expected result, not a true-positive demonstration. That gap is still open.
- The tests themselves remain fixture `state.db` cases built by the existing
  harness; no end-to-end run through a live Hermes delegation was observed.
- `omh release hermes-smoke` ran in plan mode only, which by its own stated
  boundary is not evidence that Hermes installed, loaded, or used OMH.
  `--live --target-confirmed` was not run.
- `release checklist --json` not run — no release-lifecycle surface changed.
- The Modern-TUI widget is unchanged: it renders the model, not the verdict,
  so nothing on screen changes. This PR delivers the payload field the issue
  asked for; rendering it is a separate decision about screen space.

## Risk

- Low. The change is additive: one new key on the delegation row, one new key
  on the payload, and one new key inside the internal usage aggregate. No
  existing field changed value, and `_observed_wire_model()` is untouched, so
  the `model` column still reads exactly as before.
- The extracted `_model_alias_candidates()` is a pure move; `mixture_category_for`
  behaviour is pinned by the existing category tests, which pass unchanged.
- The `delegation_route` requested side is the current config value, which is
  rewritten per dispatch. A child that started under an earlier route is
  attested against its own `sessions.model` instead, which is why the config is
  consulted only for children whose session row names no model — and the row
  says which source it read.
- One extra `config.yaml` read per HUD poll, and only when some child lacks a
  recorded model. The read already degrades to empty on any error.

## Compatibility / Rollout

- Existing installs need `omh update` before the plugin bundle carries the new
  field; until then the HUD payload is unchanged, not broken.
- Consumers that enumerate row keys see one more key. Nothing in this repo
  asserts an exact row-key set.

## Release / Claims

- [x] Release-channel impact considered (`local` — plugin bundle only)
      observed
- [x] Prepared handoffs or plans are labeled `prepared_not_observed`, never as
      execution, review, CI, or merge evidence
- [x] Evidence contains no credentials, raw prompts, raw platform events,
      transcripts, or unrelated private data
- [x] Known manual Hermes checks are listed, including the
      `omh release hermes-smoke --live` gap

## Follow-Up

- Rendering the verdict in the Modern-TUI widget — a `disagreement` is the one
  case worth a glyph on screen, and that is a screen-space decision this PR
  deliberately leaves open.
- A main session carries no verdict by construction. Attesting one would need a
  different observation than the per-child usage grouping this reader does, and
  the payload says so rather than implying coverage it does not have.

Closes #1556


sionic-khope:
Independent verification of this PR's load-bearing premise, by the session that wrote #1556. Not an approval.

This feature has exactly one way to be silently useless: if `sessions.model` were ever written from the model that answered, the comparison would put a value against itself and could never fire. The branch changes the requested side from `read_delegation_route()` (what the issue named) to the child's own `sessions.model`, so that premise now carries the whole feature and is worth confirming from outside the branch.

Checked against Hermes Agent's own source. `sessions.model` has three write sites, and all three are the requested side:

| Site | Source of the value |
|---|---|
| `_insert_session_row` via `create_session` (`hermes_state_sessions.py:374`) | the model configured at session creation |
| `update_session_meta` (`:613`, `model = COALESCE(?, model)`) | `model_str = str(state.model)` from `acp_adapter/session.py:292` — ACP session state, not a response |
| `update_session_model` (`:652`) | an explicit mid-session `/model` switch |

No path writes it from a provider's answer. The premise holds and the comparison can genuinely fire.

The scope change itself is the right call and is better than what the issue specified: `read_delegation_route()` reads a config rewritten per dispatch, so it describes only the most recent one and would attest older children against a route that has since moved. Carrying `requested_source` on the row so a reader can tell which side was read is what makes the substitution safe to review.

One consequence worth stating in the PR body if it is not already: because the requested side now prefers a per-row value, a child whose `sessions.model` was never recorded falls back to the route and inherits that staleness. The row names it, so it is visible rather than silent — but a reviewer should know the two sources are not equally strong evidence.

Also confirming the full-suite arithmetic independently: baseline on `origin/main` dec82db8 measured 12184 OK (skipped=4) in this session before any branch existed; 12196 is that plus the 12 cases this PR adds, with no test changing status.

sionic-khope:
Closing part of the "not verified" gap in the PR body. The branch reports every case as a fixture `state.db` with no live delegation observed. There is recorded data on this machine, so `served_model_attestation()` can be run against it — and was.

Method: a read-only, immutable SQLite connection opened by hand (`mode=ro&immutable=1`), reading `sessions.model` and the distinct `session_model_usage.model` per session. No file was copied, nothing was written, and the branch's own code never touched the database — `served_model_attestation()` takes two strings, so it received values and nothing else. Metadata only below; no session identifiers and no content.

Over **822 recorded sessions**:

| Verdict | Count |
|---|---|
| agreement | 763 |
| unknown | 59 |
| disagreement | 0 |

Bases: `resolved_match` 763, `no_usage_observation` 58, `multiple_usage_models` 1.

What this does and does not establish, stated separately because they are easy to conflate:

- **It does** show the function runs against real recorded data without error, and that the alias resolution does not over-report: 763 real sessions resolved to agreement and produced **zero** false disagreements. That is the failure mode a normalizer gets wrong, and at this scale it did not.
- **It does** show both unknown bases occur in reality rather than only in fixtures — including `multiple_usage_models`, which fired once. The mid-run-switch case the feature exists to be honest about is real, not hypothetical.
- **It does not** show the detector catches a true substitution. There were none here to catch. Zero disagreements on a healthy machine is the expected result and is not a true-positive demonstration; only a session where a provider actually served something other than what was routed would be, and this machine has no such record.

So the remaining gap is narrower than the PR body states: not "no attestation has been observed against real data" but "no attestation has been observed against a real *disagreement*." Worth correcting in the body, since the two are different claims and the second is the one still open.

Issue #1589, opened 2026-09-15
test_loop_tool parity: a length derived from the loop id differs by one, and the scrub list is an entry short


sionic-khope:
## Symptom

`test_loop_tool.ToolCliParityTests.test_goal_driver_observe_reaches_the_same_card`
fails on `test-windows (1)` with a numeric field differing by exactly one:

```
PR #1549  @c919b2d0   ... 68836, 'capability_snapshot_digest': None ...
                  !=  ... 68837, 'capability_snapshot_digest': None ...
PR #1585  @20e58d5    ... 69926, ...  !=  ... 69927, ...
```

Two unrelated branches, different authors, same test, same shard, same shape.
Neither branch touches `src/workflows/loop_driver_contract.py` or the loop
tool. On #1549 the failing commit is **provably comment-only** against a head
that had passed 14/14 — six added lines, all comments, zero deletions — so no
executable difference separates the green run from the red one.

## It is not a race

The delta is not arbitrary. The two loop ids the test uses differ by exactly
one character:

```
'parity-cli-loop'   15
'parity-tool-loop'  16
```

So any field that is a **length or size computed over text containing the loop
id** differs by exactly 1 between the two adapters. `scrub()`
(`tests/test_loop_tool.py:100`) replaces `_CLI_LOOP` and `_TOOL_LOOP` with
`<loop>` inside strings — but a number derived from that text was computed
before the substitution, and no string replacement can repair it.

## The test already knows this class exists

`_VOLATILE_KEYS` (`tests/test_loop_tool.py:72`) carries exactly one entry of
this kind, with the reason attached:

```py
# Derived from the loop's own handoff command, which names the loop id.
"goal_command_sha256",
```

A digest over text naming the loop id was recognised and excluded. A *length*
over text naming the loop id was not. Same hazard, second instance, missing
from the list.

## Why it is intermittent

Both observed values are large (68,836 and 69,926) and differ between runs, so
the field accumulates across the session rather than describing one record.
Shard composition varies run to run — 6052 tests in one failing run, 6037 in
the other — so whether this test sees a card shape that includes the loop id in
that total depends on what preceded it in the shard. That is why it passes on
Linux and on most Windows runs while being deterministic in its mechanism.

## Target behaviour

1. Identify the field. It sorts immediately before `capability_snapshot_digest`
   in the card's `repr`, and the assertion elides it — run the comparison with
   `self.maxDiff = None` to name it rather than inferring it.
[...]

PR #1591, opened 2026-09-15
test(loop): stop the parity card racing the clock on age_seconds

### LINKED ISSUE #1590: Loop parity test races the clock: age_seconds is derived from a scrubbed stamp but is not scrubbed
## Problem

`test_loop_tool.ToolCliParityTests.test_goal_driver_observe_reaches_the_same_card` fails whenever its two card renders straddle a one-second boundary. It reddens unrelated pull requests and there is nothing an author can do about it from their branch.

The test renders the same loop card twice — once through the CLI adapter, once through the tool adapter — and compares them after `scrub()` removes the fields two independent loops must differ on. `_VOLATILE_KEYS` correctly lists `observed_at`. It does not list `age_seconds`, which is computed from `observed_at` **and the wall clock** at render time:

```py
# src/workflows/loop_driver_contract.py:227
"age_seconds": max(0, int((datetime.now(timezone.utc) - observed_time(stamp)).total_seconds())) if stamp else None
```

So the timestamp was recognised as volatile and neutralised, while the value derived from it — which is strictly more volatile, because it moves on every render rather than once per write — was missed.

## Complete user goal

The parity suite fails only when the two adapters actually disagree about loop state, and passes regardless of when in a second each render happens. The parity coverage `age_seconds` carries today is kept rather than dropped.

## Observed evidence

CI run 34939960807, `test-windows (1)`, on a branch whose only change is one hand-written documentation file:

```
FAIL: test_goal_driver_observe_reaches_the_same_card
  File "tests\test_loop_tool.py", line 500, in test_goal_driver_observe_reaches_the_same_card
    self.assert_card_parity(cli_result, tool_result)
  File "tests\test_loop_tool.py", line 295, in assert_card_parity
AssertionError: {'all[2687 chars] 69926, 'capability_snapshot_digest': None, ...}
             != {'all[2687 chars] 69927, 'capability_snapshot_digest': None, ...}
Ran 6037 tests in 1581.565s
FAILED (failures=1, skipped=142)
```

One field, off by one, immediately before `capability_snapshot_digest` in key order. `test-windows (0)` passed on the same run, so this is not a fail-fast cancellation presenting as a failure.

Read from source rather than reproduced: the race is a sub-second window, so a local re-run proves nothing either way. `_VOLATILE_KEYS` (`tests/test_loop_tool.py:72`) contains `observed_at` and not `age_seconds`; `loop_driver_contract.py:227` derives the second from the first plus `datetime.now()`.

## Proposal

Add `age_seconds` to `_VOLATILE_KEYS` with the reason written at the entry, the way `goal_command_sha256` already carries its own.

Do **not** stop there. Scrubbing alone would silently drop what the field was checking — that both adapters compute an age at all, and compute it from the same stamp. Replace the exact-equality it loses with a targeted assertion in `assert_card_parity`: the two ages are both present or both absent, and differ by at most one second. That keeps the coverage and removes the race, instead of trading one for the other.

## Alternatives considered

Injecting a fixed clock so both renders see the same `now` — better in principle and worse in scope: `loop_driver_contract.py` has no `now` seam, and adding one to fix a test race would change a production module for a test's benefit. Worth doing if a second clock-derived field appears; not worth it for one.

Widening the tolerance instead of scrubbing — rejected. A tolerance on the whole-card equality would loosen every other field to fix one.

## Success criteria

- The parity assertion no longer depends on when in a second either render happens.
- A card where one adapter emits `age_seconds` and the other does not still fails.
- A card where the two ages come from different stamps still fails.
- The reason `age_seconds` is volatile is written where the next reader of `_VOLATILE_KEYS` will see it.

## Boundaries and risks

Test-only. No production module changes under this proposal.

The risk to weigh is the one the proposal is shaped around: adding a key to a scrub list is the kind of fix that makes a red test green without anyone noticing what stopped being checked. The paired assertion is what makes this a fix rather than a suppression, and it should be reviewed as the load-bearing half.

Found while verifying an unrelated docs PR (#1585), whose CI this failure is currently blocking.

sionic-khope:
## Capability

`ToolCliParityTests` fails only when the two adapters actually disagree about loop state, instead of also failing when the two card renders happen to straddle a one-second boundary.

## Motivation

`test_goal_driver_observe_reaches_the_same_card` failed on `test-windows (1)` of run 34939960807, on a branch (#1585) whose only change is one hand-written documentation file:

```
AssertionError: {'all[2687 chars] 69926, 'capability_snapshot_digest': None ...}
             != {'all[2687 chars] 69927, 'capability_snapshot_digest': None ...}
Ran 6037 tests in 1581.565s
FAILED (failures=1, skipped=142)
```

One field, off by one. `test-windows (0)` passed in the same run, so this is not a fail-fast cancellation presenting as a failure.

`_VOLATILE_KEYS` lists `observed_at`. It did not list `age_seconds`, which `src/workflows/loop_driver_contract.py:227` derives from that stamp **and `datetime.now()`** at render time. The timestamp was recognised as volatile and neutralised; the value derived from it — strictly more volatile, because it moves on every render rather than once per write — was not.

Both cards read the same fixed stamp (`"2026-09-14T12:00:00Z"`, `tests/test_loop_tool.py:453`), so the two ages differ by exactly the wall-clock gap between the two renders. One of those renders is a CLI subprocess spawn.

## Implementation

Test-only. No production module changes.

`age_seconds` is added to `_VOLATILE_KEYS` with its reason at the entry, the way `goal_command_sha256` already carries one.

Scrubbing alone would have been a suppression, and the comment above that list says why — *"anything further would let a real divergence pass as noise."* So `assert_card_parity` now checks the rule the producer states rather than the two ages against each other: `_ages` carries each age together with the `observed_at` beside it, and a stamped record must be aged with a non-negative int while an unstamped one must not be aged at all.

## Observed verification

The first version of this assertion **did not have teeth**, and that is why the table below exists. It compared the two ages for symmetry, which both-`None` satisfies, so forcing the producer to emit `None` left all eleven tests green — a suppression that read like a fix. Measured against the stamp-paired assertion now in the branch:

| Producer mutated to | Result |
| --- | --- |
| `"age_seconds": None` (stamp still present) | **FAILED** (failures=1) |
| `"age_seconds": -1` | **FAILED** (failures=1) |
| `"age_seconds": "old"` | **FAILED** (failures=1) |
| unmutated | OK |

Other gates, all on this branch:

| Gate | Result |
| --- | --- |
| `ToolCliParityTests` | 11 tests, OK |
| `compileall` | clean |
| `ruff check src tests` | All checks passed |
| `docs workflows/roles/navigation/chain-table/skill-sources --check` | 5/5 OK |
| `release drift` | No drift. 18 checks passed. |

Full local suite is running under contention from sibling worktrees; CI is the uncontended signal and this PR's run is the one to read.

## Risks

Adding a key to a scrub list is exactly the kind of change that makes a red test green without anyone noticing what stopped being checked. The paired stamp assertion is the load-bearing half and should be reviewed as such — the mutation table above is the evidence that it is not decorative.

Not tested: the Windows race itself. It is a sub-second window, so no local run proves either direction; the mechanism is read from `loop_driver_contract.py:227` and the fixed stamp at `tests/test_loop_tool.py:453`.

A `now` seam in `loop_driver_contract.py` so both renders share one clock would be better in principle and was rejected on scope — it changes a production module to settle a test race, and one clock-derived field does not earn that. Worth revisiting if a second appears.

Closes #1590


sionic-khope:
Full local suite landed: **Ran 12184 tests in 1169.554s, OK (skipped=4)**.

That is the baseline count unchanged, which is the right number here — this branch adds no test case, it strengthens an assertion inside eleven existing ones. A count that had moved would have meant I added coverage somewhere I did not intend to.

sionic-khope:
The fix is right and I am not asking for a change to the diff. One premise in the motivation is false, and it is the load-bearing one, so it is worth correcting in the record rather than leaving for the next reader.

**There is no subprocess.** The body says:

> Both cards read the same fixed stamp (`"2026-09-14T12:00:00Z"`, `tests/test_loop_tool.py:453`), so the two ages differ by exactly the wall-clock gap between the two renders. One of those renders is a CLI subprocess spawn.

`run_cli` runs in-process. `tests/_cli_harness.py:33-35` calls `main(args)` directly under `redirect_stdout` / `redirect_stderr`, and there is no `subprocess`, `Popen`, or interpreter spawn anywhere in that harness.

Measured rather than read: patching `loop_driver_contract.datetime` with a stand-in whose `now()` advances one second per call, and then running `test_goal_driver_observe_reaches_the_same_card`, reproduces the CI failure byte-for-byte in shape —

```
AssertionError: {'all[2687 chars] 68401, 'capability_snapshot_digest': None ...}
             != {'all[2687 chars] 68402, 'capability_snapshot_digest': None ...}
clock.now() calls: 2
```

An in-process `mock.patch.object` cannot reach a child process. It reached both renders, so both renders are in this process, and the stand-in counted exactly two reads — one per card.

**Why it matters here and not only pedantically.** The subprocess claim is the argument for scrubbing rather than tolerancing:

> Rejected: allow the two ages to differ by up to a second | the gap is a subprocess spawn, not a rounding boundary, so the bound would flake or would not catch anything

The gap is in fact a rounding boundary — two clock reads microseconds apart in one process, which is also why this is rare on Linux and shows up on a contended Windows runner. So the stated reason is exactly inverted. The conclusion survives on a true premise, and I think a stronger one: a tolerance would paper over a producer that reads the clock twice for one comparison, and no bound makes two independent `now()` calls into one observation. That reaches the same fix without resting on a spawn that is not there.

Everything else here I checked and agree with. `age_seconds` is the field (`src/workflows/loop_driver_contract.py:227`), it sorts before `capability_snapshot_digest` because `assertDictEqual` formats through `pprint.pformat`, which sorts keys, and replacing the removed equality with the stamp-paired `_ages` rule rather than scrubbing alone is the part that keeps this from being a suppression. The three producer mutations are the right way to show it has teeth.

One further measurement, in case it saves anyone revisiting #1589's stated mechanism: with the clock frozen and the tool loop id driven at 16 characters and then 15, the status card has 15 integer leaves, **zero** of which move, and no key appears in one card and not the other. No length-derived field exists in this card, so the id-length asymmetry contributes nothing and equal-length ids would not have fixed the flake. The maintainer has corrected #1589 accordingly.

https://claude.ai/code/session_01R4Q3rKfZg4fvSwfut4QWaP

sionic-khope:
CI settled: **14/14 pass**, including `test-windows (1)` — the shard where the race fired on #1585.

Stating what that is and is not worth, because it is easy to over-read. The race is probabilistic: an unfixed branch also passes this shard most of the time, so one green run here is consistent with the fix working and equally consistent with the window simply not being hit. **It is not the evidence for this change.**

The evidence is the mutation table in the PR body, which does not depend on timing: forcing the producer to emit `None`, `-1`, or a string each fails the new assertion, and the unmutated tree passes. That is what shows the assertion has teeth, and it is reproducible on demand rather than once per lucky run.

The green shard adds one thing the mutation table cannot: it confirms the change does not break the shard it touches. Read it as that and no more.

Issue #1357, opened 2026-09-05
No filesystem confinement backend for fanout dispatch on Windows


sionic-khope:
## Summary

`omh coding fanout dispatch` confines the spawned owner CLI to its unit worktree on macOS (`sandbox-exec`) and, once verified, on Linux (`bwrap`). **Windows has no backend at all.** `backend()` returns `unsupported`, dispatch proceeds exactly as it did before, and the record says so with `no_os_confinement_backend_on_this_platform`.

That is honest, and it is deliberately not a failure: the feature degrades rather than blocking Windows users. But it does mean a Windows operator gets no write fence, while a macOS operator gets one, from the same command.

## What needs doing

Design and build a Windows backend, or establish that none is workable without a dependency and close this with that finding written down. The zero-dependency constraint is real: core `omh` ships with no runtime dependencies, so anything requiring a third-party package needs its own approval.

Directions worth evaluating:

- A restricted token plus a Job Object, which is the closest native analogue to what the other two backends do.
- AppContainer, which is stronger but has a much larger surface to get right.
- Whether either can express "writes outside this directory fail, reads are unrestricted", which is the exact boundary the other backends implement and the receipt attests.

Whatever lands must keep the existing contract: confinement is reported as observed only after a probe process in that run tried to write outside the root and was refused, and a host that cannot confine degrades with a named reason instead of a silent claim.

## Pointers

- `src/coding/fanout_confinement.py` — the probe and receipt contract any backend must satisfy
- `src/quality/cross_harness_adapter_sandbox.py:backend` — where platform selection happens
- `src/quality/safety_preflight.py` — the `runtime_filesystem_confinement` row and its capability flag

PR #1598, opened 2026-09-15
fix(coding): make fanout's write fence start and hold on Linux

### LINKED ISSUE #1356: Verify fanout dispatch filesystem confinement on Linux with a trusted bwrap
## Summary

`omh coding fanout dispatch` now places the owner CLI and the dispatcher's verification commands under an OS filesystem sandbox, so writes outside the unit worktree are refused. The macOS backend (`/usr/bin/sandbox-exec`) was built and proven with a real probe. **The Linux backend has never been executed.**

`src/quality/cross_harness_adapter_sandbox.py` already selects `bwrap` on Linux and `sandbox_command` already emits a bwrap invocation, because the cross-harness adapter lane uses it. What has not happened is anyone running the fanout confinement path on a Linux host: the machine this was developed on is macOS and has no `bwrap`, so every Linux claim in this feature rests on code reading.

## What needs doing

On a Linux host with a trusted `bwrap` (see `cross_harness_adapter_backend.trusted_bwrap` for what "trusted" means here):

1. Run the confinement probe and confirm the receipt reports `observed` with a real refusal, the same way the macOS probe does.
2. Confirm a confined process can still do useful work: `git --version`, `git status` inside the worktree, and a language runtime. The macOS policy needed broad reads for exactly this reason; the bwrap invocation binds a `--tmpfs /` plus explicit read-only binds, which is a different shape and may need its own adjustment.
3. Confirm the write fence holds against the same escape set the macOS probe covers: a child-process write outside the root, `rename` out, hardlink out, and a write through an in-root symlink pointing outside.
4. Confirm the scratch directory stays invisible to `git status` in a **linked** worktree, which is what dispatch creates.
5. Confirm a host without a trusted bwrap degrades exactly as before, with `sandbox_backend_unavailable` and no enforcement claim.

## Why it is filed rather than done

The rule this feature is built on is that confinement is only reported as observed when a probe process in that run was actually refused. Claiming Linux works from reading the code would be the exact thing the receipt design exists to prevent.

## Pointers

- `src/coding/fanout_confinement.py` — probe, receipt, and the fanout policy request
- `src/quality/cross_harness_adapter_sandbox.py` — `backend`, `backend_available`, `preflight`, `sandbox_command`
- `tests/test_fanout_confinement.py` — the macOS-shaped coverage

awss1i:
Closes #1356.

### What Changed

- On Linux, `omh coding fanout dispatch` now runs the owner CLI and its verification commands inside a bwrap write fence that its same-run probe attests. Before this change the Linux path had never executed, and on a real host it could not: the preflight failed and every dispatch ran unconfined.
- `sandbox_command` gains a broad-read bwrap layout, used only when the caller passes `allow_broad_file_read` (fanout does), and an `inherit_environment` option. The cross-harness adapter lane's strict layout is byte-identical; its pinned argv test is unchanged and passes.
- `preflight` accepts the same two options, so fanout's Linux preflight exercises the layout it will spawn under.
- `prepare_fanout_filesystem_confinement` uses that layout for the bwrap preflight, probe and command, creates `.omh/confinement-tmp` and redirects `TMPDIR` on Linux as it already did on macOS, and reports `sandbox_backend_unavailable` when no trusted bwrap exists. The sandbox-exec path is unchanged.
- The real-sandbox tests that ran only on macOS become a contract shared by the macOS class and a new Linux class; backend-independent tests move to a POSIX policy class; two exact-argv tests pin the new layout.
- `docs/FANOUT.md` no longer says dispatch provides no filesystem confinement, the `runtime_filesystem_confinement` statement covers both platforms, and the changelog records the fix.

### Why This Exists

#1356 asked for the Linux backend to be run on a real host. Observed on Fedora 44 (kernel 7.1.10, bubblewrap 0.11.0 at `/usr/bin/bwrap`, which `trusted_bwrap()` accepts), before this change:

| Check | Observed |
| --- | --- |
| Receipt | `status: prepared_not_observed`, `enforced: false`, `reason_code: sandbox_preflight_failed`; `command()` returned `None` |
| Preflight | `bwrap: execvp /usr/bin/true: No such file or directory`. The `--tmpfs /` root bound only each executable's directory, so the dynamic loader `/lib64/ld-linux-x86-64.so.2` was absent. With the loader bound, the probe's `/bin/sh` still failed, because `/bin` is a symlink that did not exist inside the sandbox. |
| Write refusal | With both of those bound, a write outside the worktree exited 0 into the private tmpfs root instead of being refused, so the probe could never pass. |
| Consequence | Dispatch spawned the owner unconfined, and every escape attempt listed below wrote to the host. |

Three more defects surfaced once the fence could run:

1. `--clearenv` plus the prepare-time `--setenv` dropped the spawn-time environment, so `OMH_MARK=present /bin/sh -c 'test "$OMH_MARK" = present'` failed as a confined verification command.
2. From a read-only host tree, `systemd-run --user --wait /usr/bin/touch <path outside the fence>` created the file on the host: the session bus in `/run/user/<uid>` stays reachable, and a read-only mount does not stop a socket connect.
3. A missing or untrusted bwrap reported `sandbox_preflight_failed`, not the `sandbox_backend_unavailable` that #1356 item 5 expects, because `backend_available()` answers for the platform only.

### User / Operator Impact

- A Linux operator with a trusted, working bwrap gets the write fence macOS already has, attested per run.
- Verification commands with environment overrides keep working under the fence.
- A Linux host without a trusted bwrap dispatches exactly as before, now recorded as `sandbox_backend_unavailable`.
- Tradeoffs under the Linux fence, observed on the same host and detailed under Risk: agent sockets in `/run/user/<uid>` are unreachable, and a nested user-namespace sandbox cannot start.

### How It Works

- **Layout** (bwrap, broad reads requested): `--ro-bind / /` is the Linux counterpart of sandbox-exec's `(allow file-read*)`, so tools find their loader, libraries, config and certificates, and a write outside the binds fails with `EROFS`. Then `--dev /dev`, `--proc /proc`, and `--tmpfs /run/user/<uid> --remount-ro /run/user/<uid>` when that directory exists (skipped otherwise, because bwrap cannot create a mount point on the read-only root). The worktree `--bind` and the owner-state `--bind-try` entries follow unchanged. The existing unshare flags stay, including `--disable-userns`.
- **Environment**: `inherit_environment=True` drops `--clearenv`/`--setenv`, matching sandbox-exec, which never clears the environment. The dispatcher already filters the child environment before spawn and adds per-command overrides at spawn time.
- **Same-run evidence**: the preflight and the probe run under the layout of the spawned command, so the receipt attests the policy the owner actually gets. sandbox-exec keeps its stricter probe policy exactly as before.
- **Backend selection**: with no trusted bwrap, prepare returns `sandbox_backend_unavailable` before any spawn.

- `src/quality/cross_harness_adapter_sandbox.py`: `sandbox_command` broad-read branch and `inherit_environment`; `preflight` keyword options; `_user_runtime_directory` / `_hidden_user_runtime_directory`. The strict branch is unchanged.
- `src/coding/fanout_confinement.py`: backend check, Linux preflight and probe policy, Linux `TMPDIR` scratch (the module-private constant is renamed `_FANOUT_TOOLCHAIN_TEMP_DIRECTORY`). The receipt schema `fanout_filesystem_confinement/v1` and its claim boundary are unchanged.
- `src/coding/action_gate.py`: the `runtime_filesystem_confinement` statement describes both platforms.
- `tests/test_fanout_confinement.py`: `_ConfinedSpawnContract` holds 8 real-sandbox tests used by `FanoutFilesystemConfinementTests` (macOS; test names and assertions unchanged apart from the refusal text moving to a class attribute with the same value) and `LinuxBwrapFanoutConfinementTests` (plus 3 Linux-only tests). `FanoutConfinementPolicyTests` (POSIX) holds 4 tests moved from the macOS-only class and 4 new ones. The file grows from 508 to 819 lines; the rest of its diff is moved methods.
- `tests/test_cross_harness_adapter_security.py`: two exact-argv tests for the broad-read layout, beside the strict-layout test.
- `docs/FANOUT.md`, `CHANGELOG.md`.

### Affected Surfaces

- [ ] Hermes chat or wrapper
- [ ] Codex
- [ ] Claude Code
- [ ] Generic executor
- [x] Not executor-specific (the fence wraps whichever owner CLI fanout dispatches)

## Validation

Check only commands that completed successfully. Leave non-applicable checks
unchecked and explain why under **Not Tested / Not Applicable**.

- [x] `PYTHONPATH=tests uv run python -m unittest discover -s tests -v` (12,320 tests OK, 27 skipped, with bun 1.3.14 on PATH as CI installs it; without bun the only failure is the distribution test that requires it, identical on the unmodified base)
- [x] `git diff --check`
- [x] `omh --help`
- [x] `omh --omh-home /tmp/omh-smoke --hermes-home /tmp/hermes-smoke release hermes-smoke --install-path setup --omh-command omh --include-command-smoke` (run with scratch directories as both homes)
- [x] Relevant dry-run or smoke command: a probe and escape matrix driven through `prepare_fanout_filesystem_confinement`, `FanoutFilesystemConfinement.command()` and `_run_verification_command` on Fedora 44, with linked worktrees under `/tmp` and under `/home` (see Observed Evidence)

### Observed Evidence

- Targeted tests, commands, CI checks, and manual behavior actually observed:
- The #1356 checks with this change, identical for a `/tmp` and a `/home` worktree:

  | #1356 item | Observed |
  | --- | --- |
  | 1. Probe | `status: observed`, `enforced: true`, inside write 0, owner-state write 0, outside write 1, refusal `Read-only file system` |
  | 2. Useful work | `git --version`, `python3`, `git status --porcelain` in a linked worktree (empty), reading `/etc/os-release`, writes to the owner state directory and its `.claude.json` literal, `mktemp` into `.omh/confinement-tmp`, a spawn-time variable reaching the child, and a confined `OMH_MARK=...` verification command all succeed |
  | 3. Escape set | refused: direct write, child process, `mv` out (`inter-device move failed`), hardlink out, write through an in-root symlink to outside, write into an unrelated repository, `mkdir` outside, exclusive create, chdir-then-relative, `$HOME`, `/tmp`, `/run/user/<uid>`; nothing appeared on the host |
  | 4. Scratch in a linked worktree | temp files land in `.omh/confinement-tmp`; host `git status --porcelain` stays empty |
  | 5. No trusted bwrap | missing binary, user-owned copy, and `backend_available()` false each give `sandbox_backend_unavailable` with `command()` returning `None` |

- `systemd-run --user --wait /usr/bin/touch <outside path>` under the fence: `Failed to connect to user scope bus via local transport: No such file or directory`, no file created. The system bus refused it with an interactive-authentication error, and a `/dev/shm` write did not reach the host.
- Negative controls: against the unmodified source, 11 of the new or reused Linux and policy tests fail (for example `'prepared_not_observed' != 'observed'`); with `inherit_environment` forced back to `False`, both verification-override tests and the spawn-environment test fail; with the runtime directory left visible, the hidden-runtime test fails because the directory is not empty.
- Without a trusted bwrap (patched to simulate a CI runner), the confinement, adapter, data-boundary, safety-preflight, dispatch and handoff-safety modules run 411 tests OK, and all 11 Linux real-sandbox tests skip with their reason.
- `docs chain-table`, `docs navigation`, `docs skill-sources`, `docs agent-skills`, `docs ulw-inventory`, `docs ulw-site` and `docs claims --check --json` all pass. `tests/test_agent_board_kanban` passes after clearing `build/` and resyncing.

- Skipped checks and the concrete reason each one does not apply:
- macOS: no macOS host. The sandbox-exec branch and its probe policy are unchanged, and the 8 shared tests keep their macOS names and assertions.
- Windows: not run; the new policy class is POSIX-only.
- Ubuntu 24.04 and later (AppArmor restricts unprivileged user namespaces) and other distributions: not tested. By construction the Linux class skips where bwrap cannot start a sandbox, and on that code path dispatch would record `sandbox_preflight_failed`; neither was observed.
- A real Codex, Claude Code or Hermes dispatch under the Linux fence: not run.
- CI does not run the Linux real-sandbox class, because its runners have no bwrap; it runs the policy and argv tests.
- `omh release hermes-smoke --live`, the manual Hermes/TUI check, and the workflow-learning smoke: not run; this change touches neither Hermes, the TUI, nor learning contracts.
- Scope of the claim: the fence confines direct filesystem writes. It does not stop a confined process from asking a reachable local service to write for it, beyond the per-user runtime directory hidden here.

## Risk

- Linux hosts that dispatched unconfined now dispatch confined. An owner that writes outside its worktree and state directories (for example a tool cache elsewhere in `$HOME`) is refused, as it already is on macOS.
- Agent sockets in `/run/user/<uid>` are hidden. On the test host that included the SSH agent (`/run/user/1000/gcr/ssh`) and the GPG agent (`/run/user/1000/gnupg`), so a confined owner cannot push over SSH through the agent or sign with gpg-agent.
- The existing `--disable-userns` flag now applies to real Linux dispatches: a nested `bwrap` or `unshare --user` inside the fence fails with `ENOSPC`, so an owner CLI that sandboxes its own tools with user namespaces cannot start that inner sandbox.
- The change sits in the shared sandbox module; the strict adapter layout is pinned by its existing argv test.

## Compatibility / Rollout

- No schema change: `fanout_filesystem_confinement/v1` fields are unchanged; Linux receipts can now carry `enforced: true`, and `sandbox_backend_unavailable` is an existing reason code.
- macOS and Windows behavior is unchanged. Linux without a trusted bwrap behaves as before apart from the reason code.

## Release / Claims

- [x] Release-channel impact considered (`stable`, `preview`, or `local`): no channel-specific code; the change reaches every channel that ships fanout
- [ ] Prepared handoffs or plans are labeled `prepared_not_observed`, never as execution, review, CI, or merge evidence: not applicable, no handoff or plan output changed

## Follow-Up

- The cross-harness adapter lane's strict layout fails its preflight on Linux the same way (observed: `bwrap: execvp .../python3.12: No such file or directory`); its read list would need the loader and library roots. Out of scope here.
- A CI job with bwrap on a runner that allows unprivileged user namespaces would run the Linux class on every PR.
- Whether fanout should keep `--disable-userns`, given owner CLIs that sandbox their own commands.
- Whether operators need an explicit allowance for agent sockets inside the fence.
- #1357 (Windows backend).


sionic-khope:
Merged as #1601 — thank you.

Both of your commits are in `main` with your authorship preserved. The fix stands as you wrote it; what I added on top is two wording corrections that review turned up, both truthfulness rather than mechanism:

- `_hidden_user_runtime_directory()` returns nothing when `/run/user/<uid>` does not exist, so the data-boundary statement and `docs/FANOUT.md` now say "when that directory exists" and point at the run's own `probe.command` rather than asserting it unconditionally.
- `docs/FANOUT.md` now states that the runtime-directory cover is one socket and not the class — the `docker` group case is wider than the `systemd-run` hole you closed, and abstract-namespace unix sockets cannot be hidden by any mount at all. The fence now reads as confining a child's own writes rather than containing one that is trying to escape.

I also corrected a comment of mine-adjacent scope: the preflight and the probe both take the broad-read layout, but only the probe matches the spawn argv, since `preflight()` takes no write roots and so omits every `--bind-try`. Strictly narrower, so it cannot over-attest — and the probe, which is what the receipt attests, is byte-identical.

Two things about your PR worth saying plainly. Finding this by running it on a real host is why it got found at all: the Linux path had never executed, and reading the code would not have shown that, because the failure looked like an orderly "no backend available". And disclosing the `systemd-run --user` escape you hit **after** getting the fence working — rather than shipping the fence and letting someone else find it — is the part I would point other contributors at.

Closing this in favour of #1601. Your commits are the substance of it.

awss1i:
Thanks for carrying it over with the commits intact. The two corrections are right: the runtime-directory cover only exists when `/run/user/<uid>` does, and closing the `systemd-run` route was one socket, not the class. `docs/FANOUT.md` reads more honestly for it.

PR #1681, opened 2026-09-18
fix(fanout): stop the read-root screen from dropping the write fence

### LINKED ISSUE #1602: read_roots_are_safe disables the fanout write fence instead of narrowing reads
## The inversion

`fanout_confinement.py` screens its read roots with `read_roots_are_safe()`, whose stated job is to protect data exposed to a child. In the fanout lane it cannot do that job, and the thing it does instead is the opposite.

Every fanout call site passes `allow_broad_file_read=True`, which discards `roots` in both backends — reads are the whole host tree either way. So the screen cannot reduce what a child reads. What it *can* do is return `unsafe_sandbox_read_root` → `_unconfined`, which means the run gets **no write fence at all**.

`_SENSITIVE_PARTS` (`src/quality/cross_harness_adapter_sandbox.py:34`) contains `.claude`, `.hermes`, `.omh`, `.config`, `.ssh`, `.aws`, and `roots` includes each resolved executable's parent directory. So an owner CLI installed at `~/.claude/local/claude` — a standard Claude Code install location — puts `.claude` into a read root, trips the screen, and disables the fence for that dispatch.

**The failure mode: the more sensitive the executable's location, the less confined the run.** The screen grants the read it was screening for and removes the write protection it was not.

## Pre-existing, not introduced by the Linux fence work

Verified on `origin/main`, both halves verbatim: the screen → `_unconfined` path, and `allow_broad_file_read=True` at the `command()` call site. On macOS this has been live and fence-disabling since the feature shipped. On Linux it was moot only because the preflight failed first and every run was unconfined regardless — which #1601 fixes, making this reachable there too.

## Target behaviour

Decide what the screen is for in this lane, and make it do that:

- If it is about **reads**, it has nothing to act on while `allow_broad_file_read=True` and should not run — or the broad-read flag should not be unconditional.
- If it is about **executables in sensitive locations**, disabling the write fence is the wrong response. Refusing the dispatch, or confining it and reporting the read exposure, are both defensible; silently removing confinement is not.

Either way the current outcome — sensitive path detected, therefore less protection — must not survive.

## Acceptance criteria

- A case pinning the `~/.claude/local/<cli>` shape, asserting whatever the new behaviour is, and failing under today's.
- The reason recorded at the code, because the current behaviour reads as deliberate until you trace `allow_broad_file_read` to the call sites.
- No change to the strict cross-harness-adapter lane, where the screen does narrow reads and works as intended.

## Notes

Found during review of #1598 / #1601. That PR states the read exposure in `docs/FANOUT.md`; this issue is about the write fence being removed, which is a different problem in the opposite direction.

costajohnt:
### What Changed

- `prepare_fanout_filesystem_confinement` no longer screens its read roots with `read_roots_are_safe`. The screen's failure path returned `_unconfined`, which removed the write fence for that dispatch.
- A comment at the call site records why the screen does not apply in this lane.
- New test `test_owner_cli_under_a_sensitive_read_root_keeps_the_write_fence` pins the `~/.claude/local/<cli>` shape: `read_roots_are_safe` still rejects those roots (negative control for the strict lane), the receipt is `observed` and enforced, and the spawn policy carries the worktree write subpath with no home path in it. Fails on main with `'prepared_not_observed' != 'observed'`.

### Why This Exists

Closes #1602. In the fanout lane `command()` spawns with `allow_broad_file_read=True`, so the child already reads the whole host tree on both backends and `roots` only shape the sandbox-exec preflight and probe (omh's own `/usr/bin/true` and `/bin/sh` script). The screen therefore had nothing to narrow. What it did do was drop the write fence: `_SENSITIVE_PARTS` contains `.claude`, `roots` includes each executable's parent, so an owner CLI at `~/.claude/local/claude` tripped the screen and that dispatch ran unconfined. The more sensitive the executable's location, the less confined the run.

### User / Operator Impact

- Dispatches whose executable lives under a sensitive path (the standard Claude Code `~/.claude/local` install) now keep the write fence instead of running unconfined. No new prompts, no config change.

### How It Works

- The four-line screen and its `_unconfined(..., "unsafe_sandbox_read_root", ...)` return are removed from the fanout lane. Preflight and probe keep their strict read policy over the same `roots` (the probe still passes `allow_broad_file_read=selected == "bwrap"`), which is a strict subset of the spawn's broad read and changes nothing about what the receipt attests.
- The strict cross-harness-adapter lane (`src/quality/cross_harness_adapters.py`) keeps its screen; there the roots do narrow reads.
- Rejected: refusing dispatch on an unsafe read root (option b in the issue). Every install under `~/.claude/local` would stop dispatching.

- `src/coding/fanout_confinement.py`: remove the screen, drop the now-unused import, add the comment.
- `tests/test_fanout_confinement.py`: one new test, `read_roots_are_safe` imported for the negative control.

### Affected Surfaces

- [ ] Hermes chat or wrapper
- [ ] Codex
- [x] Claude Code
- [x] Generic executor
- [ ] Not executor-specific

## Validation

- [x] `git diff --check`
- [ ] `omh --help`

### Observed Evidence

- `PYTHONPATH=tests uv run python -m unittest discover -s tests -p 'test_fanout_confinement*'`: 30 tests, OK (11 skipped, the bwrap cases on macOS).
- Full suite: 12839 tests ran, 17 skipped, one failure: `test_cache_retention_count_is_machine_readable`, which fails identically on `main` when the shell exports `FORCE_COLOR=3` (node colours the `console.log` output) and passes with the variable unset. Unrelated to this change.
- `ruff check src tests`: all checks passed. `compileall`, `docs workflows --check`, `git diff --check`: clean.

- Generated-docs checks other than workflows, harness validate, release checklist, hermes-smoke: no generated file, skill, role, capability family, or release surface is touched.
- Manual Hermes/TUI check: not run; the change is in the sandbox-policy construction and is covered by the policy assertions in the new test.

## Risk

- Low. Behaviour changes only on the path that previously returned `unsafe_sandbox_read_root`; every other dispatch builds the same policy as before.

## Compatibility / Rollout

- No config, receipt schema, or CLI change. The `unsafe_sandbox_read_root` reason code is no longer emitted by the fanout lane.

## Release / Claims

- [ ] Known manual Hermes checks are listed, including any `omh release hermes-smoke --live` gap

## Follow-Up

- None.


costajohnt:
Closing, #1684 landed the same fix (the `read_roots_are_safe` screen is gone from the fanout lane, plus the claim-boundary text and the strict-lane test) and #1602 is closed. Nothing left here worth carrying. Thanks.

Issue #1690, opened 2026-09-19
19 skills render `coding-handling` as their preferred harness, 15 of them in non-coding categories


sionic-khope:
## Problem

21 of 127 skills have no entry in `_PRIMARY_HARNESSES` (`src/skills/catalog_harnesses.py`). 19 of those render the fallback line "Preferred harness: `coding-handling`" into their body. 15 of the 19 are not coding skills: `achievements`, `buzz`, `capability-toggle`, `live-incident-response`, `meta-router`, `running-work-board`, `rust`, `todo-checklist` and more.

Those bodies instruct the model to record a coding-handoff harness run for work that is not coding. It is the single highest-leverage defect a catalog-wide audit found, because it is wrong text in always-loaded context on every one of those skills.

## Observed evidence

Measured on the producers (`builtin_skill_templates()`, `builtin_definitions()`), never on the committed `skills/*/SKILL.md`. Full table of all 127 rows: weakness score, body size, authored share, unique-block share, reference count, harness, schema-token count, payload density.

## Proposal

Give each of the 21 a correct primary harness, or make the renderer omit the line when no harness applies rather than defaulting to a coding one. The second is smaller and cannot be wrong.

## Success criteria

- No non-coding skill renders `coding-handling`.
- A test re-derives the set from the producer and fails when a skill without a primary harness renders a harness line.
- Regenerated `skills/`, `agent-skills/`, `docs/WORKFLOWS.md`; every `docs ... --check` gate green.

PR #1751, opened 2026-09-19
What keeping a plan costs: one item per write, and a shorter block per turn

### LINKED ISSUE #1730: The plan rule is restated in full on every turn with no decay, and `omh_todo` has no single-item update
## Two costs of keeping a plan, both measured by calling the real hooks (audit, 2026-09-19)

**1. The reconciliation sentence never decays.** Every other OMH injection has an off-switch: code-mode guidance latches once per session, engagement nudges cap at two, the board card claims once, the route hint claims per fingerprint, the repeat streak expires after 300 s. The plan line and its 357-char reconciliation rule fire on every turn for as long as a plan is open. A 40-turn session with one plan pays about 37,000 chars of hook text, of which about 14,000 is that one identical sentence.

The 2026-09-11 owner decision stands: the plan reminder must DRIVE continuation to an explicit stop criterion, and observe-only was rejected. This issue does not touch that. The drive is the plan line and the open-item count, which stay on every turn. The proposal is only that the long completion-claim guard is sent in full while it is news (the first few turns of a plan, and again whenever the plan record changes) and as its first clause otherwise.

**2. `omh_todo` rewrites the whole list to tick one item.** `action=set` is the only write (`src/plugin_bundle/omh/tools/todo_tool.py`). A 10-step plan costs 1 set to declare plus 10 sets to advance, each re-sending all 10 items with text, state and phase, about 1.5 KB of arguments per call, 16 to 24 KB of bookkeeping for one task. It also makes every advance a chance to drop or reword an item by accident.

Proposal: `action=advance` taking an item index (or id) and a state, plus the existing optional fields (`blocked_reason`, `deferred_reason`). Same record, same stop criteria, same fields read.

## Acceptance

- Hook measurement in the PR: chars injected on turn 1, turn 5, and turn 5 after a plan change, before and after.
- The stop criteria and the drive wording on the plan line are unchanged; the existing precedence tests (blocked > deferred > message-opened > drive) pass untouched.
- `advance` on an out-of-range index, on a finished plan, and with a state that needs a reason but has none: each refused with a message naming the field.
- A plan advanced only through `advance` produces a record byte-equal to the same plan advanced through `set`.

sionic-khope:
> **Stacked on #1747.** The base of this PR is `claude/harness-line-truth`, not
> `main`, so that CI has something to run: a stacked branch aimed at `main`
> reads as conflicting and GitHub creates no check run for it. The lane now
> sits on top of `main` and carries #1743 and #1739, so this branch is exactly
> its own two commits and the diff is only this change. Retarget it to `main`
> when the lane merges.

### What Changed

- `omh_todo` gains `action=advance`: change one item's state with the item
  number, the start of that item's current text as a guard, and the new state.
  `blocked_reason` rides with it and `deferred_reason` behaves exactly as it
  does on a `set`.
- Every write to a todo record now goes through a per-record OS lock, taken by
  `write_todo` as well, because the new path is the first one that reads a
  record in order to write one.
- The per-turn plan block lost 106 to 146 characters per turn on the variants
  that drive, without the drive or its stop criterion losing a word of force.
- The `todo-checklist` skill's claim that "there is no partial update" was true
  when written and is not now; it and the routing policy that repeated it name
  the new action instead.
- `set` gains a failure it never had: a contended record is refused with
  `"status": "contended"`, not `invalid_todo`, and nothing is written.

Closes #1730

### Why This Exists

- Ticking one item cost a whole-list write. A ten-step plan paid one
  declaration plus ten re-sends of every item's text, state and phase, about
  1.5 KB of arguments per advance, 16 to 24 KB of bookkeeping for one task, and
  one chance per advance to drop or reword an item nobody meant to touch.
- The plan block is read at the start of every turn a plan is open. #1739
  capped the reconciliation rule's tail; the rest had never been measured
  sentence by sentence against the job each one does, and three places said one
  thing twice.

### User / Operator Impact

- A session walking a plan sends one item's worth of arguments per advance
  instead of the whole list, and cannot silently drop an item it forgot to
  re-send.
- A stale item reference is refused rather than applied, so a plan that was
  re-scoped between two turns cannot have the wrong item ticked.
- Less of every turn's context is the same sentence again. Nothing here claims
  that changes what a model does; it changes what the model is told.

### How It Works

**The item reference is an index plus a guard, not an id.** The record carries
no item id and needs none for anything else. Adding one would move the on-disk
schema, the digest a deferral lapses on (`_DIGESTED_ITEM_KEYS`), and every
surface that projects an item, to carry a handle whose only reader would be
this action. What a bare index cannot do is refuse when the list moved
underneath it, so `item_text` is required and must be a prefix of the text
stored at that position. `item` is 1-based, the way the checklist reads.

**It is the same write by a narrower route, enforced rather than agreed.** The
new list is the stored list with one entry's `state` and `blocked_reason`
replaced; it then goes through `build_todo_record` — same title, source,
session and deferral handling, same `validate_todo_items`, same stamp. There is
no second validator and no second schema. A plan walked entirely through
`advance` produces a record byte-equal to the same plan walked through `set`
under a fixed clock, which is the acceptance criterion and a test.

**The refusals name the field**, and each one reads a record rather than any
wording: a reference that is not an integer or is out of range names `item` and
the item count; a missing or non-matching guard names `item_text` and says what
it is for; a state outside the enum names `state`; a finished plan and an
absent record each name `action=set` as the way on. A refused call leaves the
record byte-identical.

**The lock.** The store had no lock before this, and did not need one:
`write_todo` is atomic against a reader through `os.replace`, and nothing read
a record in order to write one. `advance` does, so `set` and `advance` now take
the same per-record lock and the read-modify-write is one step against any
other writer. A compare-and-set on `updated_at` was rejected: the window it
leaves open is the whole of the failure, since the stamp would be checked in a
different instant than the `os.replace` that acts on it. The lock is
`awareness_delivery`'s, the bundle's one two-backend implementation, already
taken by `tool_bursts`, `approval_bypass` and `memory_open_reminders`. This
change makes its wait a parameter rather than a constant, defaulting to the
value every existing caller had: that default is sized for telemetry that
would rather drop a counter than delay a turn, and a plan write is the
opposite trade, so the store passes 2s. Extending it beat forking it, and a
policy gate says so — `tests/test_journal_lock_portability.py` fails any
module outside a named list that calls `fcntl.flock` itself, because a copy
is how a Windows host once ended up taking no lock at all while reading as
though it had one. A timeout is a refusal, never a silent unlocked pass; a
host with neither backend takes no lock and still keeps a plan, which is what
every writer here did before the lock existed. Lock files are pruned on the
existing stale bound and only once the record they guarded is gone.

**What `set` sees when it is refused for contention**, since that is new.
Before this, `set` could not fail this way at all: it wrote through
`os.replace` and took no lock. A writer that waits more than two seconds
(`_LOCK_TIMEOUT_SECONDS`) for the record gets back
`{"status": "contended", "error": "todo record is held by another writer after
2s and was not written: <lock path>. Nothing changed; send the same call
again."}`, alongside the current projection the tool always returns. The record
is byte-untouched; the refusal is raised before anything is written, and
`test_a_contended_set_says_so_and_is_not_called_invalid` compares the record
before and after.

The separate status is the point rather than decoration. `invalid_todo` tells
a writer its payload was wrong, and a writer that believes that rewrites the
list it just sent — the whole-list rewrite this change exists to stop — when
the right response to a busy record is the identical call. `TodoContendedError`
subclasses `TodoStoreError`, so every caller written before the lock existed
keeps catching it without learning a new failure. `omh runtime todo set`
reports the same status and still exits 1, because either way nothing was
written. Nothing counts or surfaces refusals: no reader has anywhere to put a
count, so none was added.

**The plan block**, sentence by sentence and what each one does:

| sentence | job | verdict |
| --- | --- | --- |
| `n/m done · active: <text>` | state, from the record | kept |
| `Open items mean this plan is not finished` | the premise | joined to the drive with a colon |
| `unless something is blocking it, advance the next item in this turn rather than ending on a status report` | the drive | kept verbatim |
| `This stops when … blocked_reason -- not when a turn has produced an answer` | the stop criterion | trailing clause dropped, `only` added |
| `reconcile the checklist … mark completed items done, keep exactly one item active` | the record invariant | kept on every turn |
| `either finish the remaining items or say which stay open and why` | what to do instead of claiming done | moved to the budgeted half |
| `A completion claim … is a visible contradiction` | the reason | already budgeted, unchanged |
| `Todo updates are declarations, never execution evidence` | the evidence boundary | already budgeted, stays there |
| `The checklist standing still is an observation, not evidence that the work failed: say where the plan stands` | stall framing | kept |
| `and if nothing is blocking it, move it` | the drive again | dropped |

The trailing clause of the stop criterion said, in the stop criterion, what
"rather than ending on a status report" says one clause earlier in the drive;
the exclusivity it carried moved into "only", which is its stronger form. The
stall framing's last clause is the drive repeated about forty characters after
the drive, in a block that renders both. And "either finish the remaining
items or say which stay open and why" is what the drive already asks, in prose
rather than in record terms, so it moved to the budgeted side — nothing was
deleted to move it, and the full rule is within one character of what it was.

**The evidence-boundary sentence stays on the budgeted side.** It guards one
misreading — that writing `done` on an item is proof the item ran — and that
misreading is only available to a turn that is writing todo updates, which is
the turn right after the record moved. That is exactly the window
`TODO_RECONCILIATION_FULL_TURNS` keeps open. The boundary is also carried at
the two other moments it can be acted on: `omh_todo`'s own result payload
(`TODO_CLAIM_BOUNDARY`) and the turn-end directive
(`PLAN_CONTINUATION_BOUNDARY`).

**`TODO_ANSWER_FIRST_RULE` and `TODO_DEFERRED_RULE` were measured and left
alone.** The answer-first rule's four sentences are four distinct jobs written
against four halves of one observed failure, and the only fold available in it
is worth nine characters and deletes a phrase `test_person_turn_precedence.py`
pins against the incident wording. The deferred rule's last sentence duplicates
the tool schema, but it is read at a different moment: while a deferral is
already live and could be mistaken for a block.

- `src/plugin_bundle/omh/todo_store.py` — `advance_todo_item`, the per-record
  lock wrapper, the raw-record read, lock-file pruning. `omh_todo/v1` on disk
  is unchanged: no new field, no new key.
- `src/plugin_bundle/omh/awareness_delivery.py` — the shared lock's wait
  becomes a parameter, defaulting to what it was. No other behaviour moves and
  no existing caller changes.
- `src/plugin_bundle/omh/tools/todo_tool.py` — the `advance` action, its four
  parameters, and the `contended` status on both writes.
- `src/commands/runtime.py` — the same status on `omh runtime todo set`.
- `src/plugin_bundle/omh/todo_reconciliation.py` — the rule constants and where
  the reconciliation budget splits.
- `src/skills/catalog_definitions.py`, `src/skills/render.py`,
  `src/routing/recommend.py` — the false "no partial update" claim.
- `src/quality/schema_overlap.py` — reviewed verdicts for the new `action`
  enum text and the `item` bound.
- `src/maintenance/release.py` — two ratchets, each re-derived from its
  producer on the rebased tree with its reason appended beside the lane's:
  `FULL_PROFILE_SKILL_BODY_CHAR_LIMIT` 977649 to 977846, and
  `FULL_CAPABILITY_SKILL_SECTION_CHAR_LIMIT` 422526 to 422450. The second one
  FALLS across a change that adds to it: the section measured 422364 on this
  branch's base because #1690 removed a rendered harness line from fourteen
  skills while the limit still sat at the pre-#1690 value, so the new number
  is the producer's reading of both changes together and ratchets the leftover
  slack down with it. Both readings were measured; neither was computed.
- Regenerated: `docs/WORKFLOWS.md`, `skills/omh-todo-checklist/**`,
  `tests/fixtures/agent_skills_hermes_digests.json` (one line).

### Affected Surfaces

- [x] Hermes chat or wrapper
- [ ] Codex
- [ ] Claude Code
- [ ] Generic executor
- [ ] Not executor-specific

## Validation

- [x] `git diff --check`
- [ ] `omh --help`
- [x] Relevant dry-run or smoke command: `omh release skill-content-smoke`
      (through `skill_content_smoke()`), `omh.maintenance.drift.drift_report()`

### Observed Evidence

- The lane moved twice while this was in flight, so the runs below name the
  base they were on. Every `--check` gate, `drift_report()["ok"]`,
  `harness validate`, `release checklist --json`, `release hermes-smoke` and
  `ruff` were re-run at the current head (`33a1df97`, on lane tip
  `450d1ab1`) and all pass. The two unittest batteries ran at the previous
  head, whose authored files are byte-identical to this one -- the second
  rebase applied with no conflicts and changed only the base. A 19-file
  battery has since re-run at the CURRENT head as well: 840 tests, `OK`,
  covering `test_skill_reference_mechanisms`, `test_skill_harness_line_policy`,
  `test_agent_skills_projection`, `test_todo_advance`,
  `test_person_turn_precedence`, `test_injected_context_pressure`,
  `test_todo_reconciliation`, `test_plan_deferral`,
  `test_plan_continuation_driver`, `test_plan_nudge_progress_budget`,
  `test_phase_todo`, `test_schema_overlap`, `test_skill_structure_lint`,
  `test_router_content`, `test_release_smoke`, `test_hud`, `test_cli`,
  `test_hermes_ux_quality` and `test_routing_precision`.
- On lane tip `f46b6792`, the tests the review named, `OK` (252):
  `test_skill_reference_mechanisms`, `test_skill_harness_line_policy`,
  `test_agent_skills_projection`, plus `test_todo_advance` (35),
  `test_person_turn_precedence`, `test_injected_context_pressure`,
  `test_todo_reconciliation`, `test_plan_deferral`,
  `test_plan_continuation_driver`, `test_plan_nudge_progress_budget`,
  `test_phase_todo`, `test_schema_overlap`, `test_skill_structure_lint`,
  `test_exit_code_truthfulness_policy`, `test_broad_exception_policy`.
- `todo-checklist`'s rendered body is 6,954 bytes against the 26,000-byte
  per-skill ceiling, 19,046 of headroom, measured through
  `skill_structure_lint_payload()`; the lint reports no violations. #1743's
  pointer line did not bring that body near the ceiling.
- Targeted battery before the rebase, 502 tests, `OK`: `test_todo_advance` (31 new),
  `test_todo_reconciliation`, `test_phase_todo`,
  `test_person_turn_precedence`, `test_plan_continuation_driver`,
  `test_plan_deferral`, `test_plan_nudge_progress_budget`,
  `test_injected_context_pressure`, `test_hud`, `test_plugin_capabilities`,
  `test_capabilities`, `test_hook_manifest`, `test_hook_integrity`,
  `test_plugin_bundle_standalone`, `test_engagement_nudges`,
  `test_awareness_engagement`, `test_dispatch_outcomes`,
  `test_tui_widget_pack`, `test_interpreter_spawn_policy`,
  `test_exit_code_truthfulness_policy`, `test_broad_exception_policy`,
  `test_agent_board_kanban`.
- On the same base, `OK` (728): `test_router_content`, `test_release_smoke`,
  `test_cli`, `test_hud`, `test_hermes_ux_quality`, `test_routing_precision`,
  `test_plugin_capabilities`, `test_hook_manifest`,
  `test_plugin_bundle_standalone`, `test_engagement_nudges`,
  `test_awareness_engagement`, `test_dispatch_outcomes`,
  `test_tui_widget_pack`.
- Before any rebase, separately `OK`: `test_router_content`, `test_release_smoke`,
  `test_agent_skills_projection`, `test_schema_overlap`, `test_cli`,
  `test_hermes_ux_quality`, `test_routing_precision`,
  `test_hud_reader_security`, `test_runtime_paths`, `test_runtime_artifacts`,
  `test_runtime_reader_external_effect_receipts`.
- Every `--check` gate passes: workflows, roles, chain-table, navigation,
  skill-sources, capability-families, ulw-inventory, ulw-site, agent-skills,
  plus `docs claims --check --json` (`"ok": true`) and
  `drift_report()["ok"] == True`.
- Plan-block characters, measured by driving `pre_llm_call` against a temp
  home, per variant, turns 1-3 carrying the full reconciliation rule:

  | variant | t1-3 before | t1-3 after | t4+ before | t4+ after | 40 turns before | 40 turns after |
  | --- | --- | --- | --- | --- | --- | --- |
  | drive | 694 | 654 | 541 | 435 | 22,099 | 18,057 |
  | answer-first | 866 | 865 | 713 | 646 | 28,979 | 26,497 |
  | stalled | 887 | 807 | 734 | 588 | 29,819 | 24,177 |
  | deferred | 828 | 827 | 675 | 608 | 27,459 | 24,977 |

- Schema characters, `json.dumps(OMH_TODO_SCHEMA, sort_keys=True)`: 4,777
  before, 5,570 after. 571 of the 793 is the four new parameters; the rest is
  the action enum and its routing sentence.
- Mutation battery, each mutation applied to the committed tree and reverted:
  the `item_text` guard made a no-op, the item cap removed from
  `validate_todo_items`, the record lock made a no-op, the reference read as
  0-based, an omitted `blocked_reason` made sticky, the `contended` status
  folded back into `invalid_todo`, the contention timeout made to fall
  through and write anyway, the drive's imperative removed, and the stop
  criterion removed. All nine caught, twice. Results are in the PR thread.

- The full local suite was not run: machine load was 26.8 with several
  worktrees running suites, and a contended run is not a signal. CI is the
  signal for the full suite. The targeted battery above covers every surface
  this change touches.
- The Windows lock backend (`msvcrt`) has no local run on macOS. CI's
  `test-windows` job is the signal.
- `omh --help` and the installed-command smoke were not run: nothing here
  changes the CLI surface or the install manifest.
- No live Hermes session and no TUI check: the tool and the hook are exercised
  through their real entry points in tests, and a live check would need an
  install this branch is not reflected on.

## Risk

- `write_todo` now takes a lock on every write, including the CLI's, so `set`
  gains a failure mode it did not have. A host with neither `fcntl` nor
  `msvcrt` takes no lock and behaves exactly as before. A contended record
  refuses after two seconds rather than waiting forever, says so under its own
  status, and leaves the record untouched.
- Lock files are a new artifact beside each record. They are pruned on the
  existing stale bound and only when the record they guarded is gone, so an
  active writer's lock is never removed underneath it; `test_hud` pins both
  halves.
- The block is shorter. What may not be shortened away is pinned per variant
  and past the reconciliation budget by `PlanLineForceTest`, so a future trim
  that drops the drive or its stop criterion fails rather than lands.

## Compatibility / Rollout

- `omh_todo/v1` on disk is unchanged. A record written before this reads and
  advances; a record written by this is byte-identical to one `set` writes.
- The `advance` action is additive: `set`, `clear` and `show` are unchanged,
  and a model that never calls `advance` behaves exactly as before.
- No new dependency. No network. No change to any install manifest.

## Release / Claims

## Follow-Up

- Three premises in the issue and the brief did not survive measurement, and
  each is worth its own look rather than being folded in here. `omh_todo`
  enforces no "exactly one active item" invariant — it is instruction text in
  the schema and the skill, and no validator reads it; `advance` therefore
  enforces exactly what `set` does. No item state requires a reason, so the
  refusal the issue asks for has no referent; `blocked_reason` is a field
  independent of state, and what stands in its place is a missing or invalid
  `state`. And the stop criterion does not ride every turn today: it belongs to
  the drive, so the answer-first and deferred variants replace it along with
  the drive, which is deliberate and now pinned per variant.


sionic-khope:
## Mutation battery

Each mutation applied to the branch head (`3437e9eb`), the named suite run, the tree reverted. Every one was caught.

| # | mutation | suite | failing cases |
| --- | --- | --- | --- |
| M1 | `_check_item_guard` returns immediately | `test_todo_advance` | 4 |
| M2 | `validate_todo_items` stops enforcing the item text cap | `test_todo_advance` | 1 |
| M3 | `_todo_record_lock` yields without taking a lock | `test_todo_advance.LostUpdateTest` | 11 |
| M4 | `_validated_item_reference` returns the index unshifted (0-based) | `test_todo_advance` | 65 |
| M5 | an omitted `blocked_reason` keeps the stored one | `test_todo_advance` | 1 |
| M6 | the drive loses its imperative | `test_person_turn_precedence.PlanLineForceTest` | 7 |
| M7 | the stop criterion is removed | `test_person_turn_precedence.PlanLineForceTest` | 7 |

M2 is the one that says the validator is shared rather than copied: breaking the item text cap inside `validate_todo_items` breaks a test that only ever calls `advance`, which it could not do if the single-item path had its own rules.

M3's eleven are one whole-suite failure plus ten of the forty interleave attempts in `test_a_whole_list_write_is_never_overwritten_by_a_list_read_before_it`. Ten in forty is the lost update showing up at the rate an unsynchronised read-modify-write produces it on this machine; with the lock it is zero in forty, every run.

M4 is loud (65) because a 0-based read shifts every reference in the file, including the byte-equality walk. The cases that matter are the out-of-range pair, which still name `item` and the item count.

sionic-khope:
## Mutation battery, second run

Re-run after the rebase, on the branch head, now including the two mutations for the `contended` status. Each mutation applied, the named suite run, the tree reverted. All nine caught.

| # | mutation | suite | failing cases |
| --- | --- | --- | --- |
| M1 | `_check_item_guard` returns immediately | `test_todo_advance` | 4 |
| M2 | `validate_todo_items` stops enforcing the item text cap | `test_todo_advance` | 1 |
| M3 | `_todo_record_lock` yields without taking a lock | `test_todo_advance.LostUpdateTest` | 25 |
| M4 | `_validated_item_reference` returns the index unshifted (0-based) | `test_todo_advance` | 66 |
| M5 | an omitted `blocked_reason` keeps the stored one | `test_todo_advance` | 1 |
| M6 | the `contended` branch removed, so a lock timeout reports `invalid_todo` | `test_todo_advance.ContentionRefusalTest` | 2 |
| M7 | the contention timeout falls through and writes anyway | `test_todo_advance.ContentionRefusalTest` | 3 |
| M8 | the drive loses its imperative | `test_person_turn_precedence.PlanLineForceTest` | 7 |
| M9 | the stop criterion is removed | `test_person_turn_precedence.PlanLineForceTest` | 7 |

M2 is the one that says the validator is shared rather than copied: breaking the item text cap inside `validate_todo_items` breaks a test that only ever calls `advance`, which it could not do if the single-item path had its own rules.

M3's failures are one whole-suite failure plus the interleave attempts in `test_a_whole_list_write_is_never_overwritten_by_a_list_read_before_it` that lose a write. It landed on ten of forty attempts in the first run and twenty-four of forty in this one, the difference being machine load; with the lock it is zero of forty, every run.

M6 and M7 are the pair that keeps the new refusal honest. M6 proves the status is a real branch and not decoration, and M7 proves the timeout refuses before anything is written rather than writing and then reporting.

M4 is loud because a 0-based read shifts every reference in the file, the byte-equality walk included. The cases that matter are the out-of-range pair, which still name `item` and the item count.

sionic-khope:
## Mutation battery after the lock was shared

Re-run on the current head. Twelve mutations, each applied, the named suite run, the tree reverted. All caught.

| # | mutation | suite | failing cases |
| --- | --- | --- | --- |
| M1 | `_check_item_guard` returns immediately | `test_todo_advance` | 4 |
| M2 | `validate_todo_items` stops enforcing the item text cap | `test_todo_advance` | 1 |
| M3 | the store takes no lock at all | `test_todo_advance.LostUpdateTest` | 28 |
| M4 | the item reference is read as 0-based | `test_todo_advance` | 68 |
| M5 | an omitted `blocked_reason` keeps the stored one | `test_todo_advance` | 1 |
| M6 | the `contended` branch removed, so a lock timeout reports `invalid_todo` | `test_todo_advance.ContentionRefusalTest` | 2 |
| M7 | the shared lock returns a mechanism instead of raising past its deadline | `test_todo_advance.ContentionRefusalTest` | 3 |
| M8 | the drive loses its imperative | `test_person_turn_precedence.PlanLineForceTest` | 7 |
| M9 | the stop criterion is removed | `test_person_turn_precedence.PlanLineForceTest` | 7 |
| M10 | the store calls `fcntl.flock` itself again | `test_journal_lock_portability` | 1 |
| M11 | the lock wrapper relabels the caller's own `OSError` as a store error | `test_todo_advance.SharedLockTest` | 1 |
| M12 | the store drops back to the shared 0.1s telemetry deadline | `test_todo_advance.SharedLockTest` | 1 |

Worth recording how M3 and M7 got here, because the first run of this battery reported both as SURVIVED and that reading was wrong. Their patterns were written against the forked two-backend lock this revision deletes, so after the rewrite they matched nothing and the unmutated suite passed. A mutation that fails to apply looks exactly like a mutation nothing catches. Re-aimed at the code that now exists — M3 makes `_todo_record_lock` yield without entering the shared helper, M7 makes the helper return a mechanism instead of raising past its deadline — both fail loudly.

M2 is still the one that says the validator is shared rather than copied: breaking the item text cap inside `validate_todo_items` breaks a test that only ever calls `advance`.

M10 is the gate this revision exists for. M11 and M12 are the two halves of sharing a lock rather than forking one: the wrapper may not relabel a failure that is not its own, and the deadline it passes has to be its own rather than the telemetry default.

Issue #1742, opened 2026-09-19
OMH has two families of config.yaml writers and only one of them checks for a concurrent change


sionic-khope:
## Problem

OMH writes the person's Hermes `config.yaml` from two families of code that do not know about each other.

- The delegation route writer (`src/plugin_bundle/omh/delegation_routing.py`) reads, compares a file signature, and replaces atomically; with PR #1737 it also runs inside a lock and, from lifecycle hooks, without a person present.
- `src/install/config_adapter.write_config` rewrites the whole file with no lock and no signature compare. It is reached by `omh setup`, `omh theme`, `omh memory`, self-update's external-dirs registration, and `system/targets`.

Read from the code during the independent review of #1737, not reproduced as a lost update: a route write refuses when the file changed under it, but the `write_config` callers do not, so one of them can read the file, have a route land, and write back its stale copy. The route is erased, and the restore record then sees a value it did not write and drops the person's baseline as a foreign edit. Before #1737 this overlap needed a person running a CLI command during a routed dispatch. With restores now running at turn and session boundaries, the window is met more often.

## Direction

One discipline for every OMH write of `config.yaml`: read, check the signature again immediately before the replace, refuse or retry on a change, replace atomically. Whether the CLI-side writers should also take the route lock is a design choice to make with the lock's location in mind (it lives in the OMH home, and CLI commands may target a home the plugin is not bound to); state the choice.

## Acceptance

- A test interleaves a route write between a `write_config` caller's read and its write: the second writer refuses or retries, and no update is lost.
- Every `write_config` call site is either converted or listed with the reason it cannot race.
- No behaviour change for a single writer.

PR #1764, opened 2026-09-19
feat(uninstall): reverse every config key setup wrote, not one of seven

### LINKED ISSUE #1725: `omh uninstall` reverses one of the config keys setup writes; the rest keep pointing at a bundle that is gone
## Problem

Setup's apply step writes several Hermes `config.yaml` keys in one pass (`src/commands/setup.py`, `_apply_result`): `skills.external_dirs`, `auxiliary.compression.fallback_chain`, `plugins.enabled`, `display.interface`, `display.skin`, `memory.provider`.

`cmd_uninstall` writes the config exactly once, through `_remove_managed_external_dirs`. Measured by reading every config write on the uninstall path:

- `clear_memory_provider` exists in `src/install/config_adapter.py` and its only callers are the `omh memory` command and its own module. Uninstall does not call it.
- `ensure_plugin_enabled` has no inverse anywhere in `src/`.
- `uninstall_skin` and `uninstall_tui_widget` remove files; they do not write `config.yaml`.

So after `omh uninstall --all` a machine keeps `plugins.enabled: - omh` and `memory.provider: omh` naming a plugin that no longer exists, plus any `delegation.*` a route left behind. Of the six keys, only the branded-TUI pair was consent-gated at install.

Not measured: what Hermes does at startup with `memory.provider: omh` and no such provider. Worth one observed run in a temp `HERMES_HOME` as part of this work.

## Direction

Uninstall reverses each key OMH set, and only when the current value is still the one OMH wrote (a person who changed `memory.provider` to something else keeps it). Setup already records what it changed in its result payload; uninstall needs the same facts persisted, not re-derived by guessing.

## Acceptance

- Install into a temp home, uninstall, diff `config.yaml` against the pre-install bytes: only keys a person changed in between differ.
- A key whose value is no longer OMH's is left alone, and the uninstall report says so by name.
- `--registration-only` keeps its current narrow meaning.
- `--dry-run` lists every key it would reverse.

sionic-khope:
### What Changed

- `omh uninstall` reverses every `config.yaml` key `omh setup` writes, not only the `skills.external_dirs` registration — in the primary Hermes home **and in every bot profile**.
- `omh setup` (and every `omh update` that refreshes the registration) records what its apply pass added, keyed by config path, under `hermes_config_writes` in `$OMH_HOME/runtime/state.json`.
- A key is reversed only while its current value is still the recorded one. A key the person has since changed is left alone and named in the uninstall output.
- A value consent let OMH migrate the person off (`display.interface`, `display.skin`) is put back, not removed.
- Every remover refuses the same document shapes its forward writer refuses.
- `--registration-only` keeps its narrow meaning. `--dry-run` lists every key it would reverse and leaves the file byte-identical.

### Why This Exists

Setup's apply step writes seven keys: `skills.external_dirs`, `auxiliary.compression.fallback_chain`, `plugins.enabled`, `display.interface`, `display.skin`, `display.sections` (since #1750) and `memory.provider`. `cmd_uninstall` wrote the config exactly once, through `_remove_managed_external_dirs`. `clear_memory_provider` existed but only `omh memory` called it; `ensure_plugin_enabled` had no inverse anywhere in `src/`; `uninstall_skin` and `uninstall_tui_widget` removed files, not config keys.

**Observed**, Hermes 0.21.3, temp `HERMES_HOME`, bundle absent, host read-only at `~/.hermes/hermes-agent`:

| leftover key | what Hermes does |
| --- | --- |
| `memory.provider: omh` | Agent init calls `load_memory_provider("omh")` → `None`, then `recover_at_startup("omh")`, which looks the provider up in the plugin catalogue and logs `Memory provider 'omh' is configured but not installed and not in the plugin catalog.` `hermes doctor` reports `⚠ omh plugin not found run: hermes memory setup`, outside its numbered issue list. |
| `plugins.enabled: - omh` | Silent. `Plugin discovery complete: 58 found, 52 enabled`; a name with no directory produces no warning at all. |
| `display.skin: omh` | Silent. `hermes skin list` does not list it and the TUI resolves the built-in default. |

Three corrections to that lookup, from the independent review, re-checked against the installed host. The URL is `hermes-agent.nousresearch.com/docs/api/plugin-catalog.json` (`LIVE_CATALOG_URL`); the first version of this section quoted the 301 redirect target instead of the configured value. The response is cached under `HERMES_HOME/cache` for six hours (`LIVE_CATALOG_TTL_SECONDS`), so the cost is at most one request per six-hour window per home, not one per start. And `recover_at_startup` returns before any lookup when `security.allow_lazy_installs` is false (default true).

So the recurring cost of the memory slot is a warning on every affected start plus an occasional catalogue request, for a provider that will never exist. The other two keys are invisible, which is why they survived.

### User / Operator Impact

Uninstall leaves the person's `config.yaml` as it was before the install. Anything they changed in between stays theirs and is named, e.g. `Config key kept: memory.provider (memory.provider is honcho, not omh; leaving it alone)`.

Installs made before this record get the honest subset, stated in the output rather than guessed:

| key | no record | with record |
| --- | --- | --- |
| `skills.external_dirs` | reversed (the path is OMH's by construction; unchanged behaviour, every scope) | same |
| `plugins.enabled` | reversed — the name `omh` is OMH's | reversed |
| `memory.provider` | reversed — Hermes runs one external provider and `set_memory_provider` refuses to take the slot from another product, so `omh` here can only be OMH's | reversed |
| `display.skin` | reversed when the value is `omh` or an `omh-*` theme; a foreign skin is kept | reversed |
| `display.interface` | **kept**, reported as unrecorded — `tui` is one of Hermes' own values | reversed, or restored to the value consent replaced |
| `display.sections.{thinking,tools,subagents}` | **kept**, reported as unrecorded — Hermes' default is `expanded`, so `collapsed` says somebody chose it and nothing says who | reversed per key; a child the person changed is kept and named |
| `auxiliary.compression.fallback_chain` | **kept**, reported as unrecorded — derived from the person's own providers, so it carries no mark saying OMH wrote it | reversed only while the block still reads byte for byte as written |

**Bot profiles go through the same function as the primary.** `_sync_hermes_profiles` calls the apply pass once per profile home, so setup writes all seven keys to each; uninstall now takes them back the same way, with each profile's own record, its own config path and its own rows in the report. A profile that still holds a key is reported as `partially_cleared` with the keys named, not as `cleared` over a config that still says `omh`.

### How It Works

Two facts are kept deliberately separate.

**What OMH wrote** is recorded at write time as the difference between `managed_config_reading()` before the writer chain and after it — the same test #1750 applied to `display.sections`, extended to every key. The difference is what makes a key OMH's: a value the person already had is preserved by the writers and stays theirs. The record merges forward across runs, because the apply pass runs on every `omh update` too and on the second run every writer reports "already set"; a record holding only the latest run's writes would erase itself on the first update.

**Whether it is still OMH's** is re-read at uninstall time and compared against the record, per key.

The record is **keyed by config path**, and it has to be. `_profile_clone` sets `clone.omh_home = str(primary.omh_home)`, so every bot profile shares the primary's store and one `omh setup` writes the same `runtime/state.json` once per Hermes home. A single slot would end the run holding the last profile's record, and the primary — the home somebody is most likely to uninstall — would find none of its own and fall back to the no-record subset. Each entry's `config_path` is checked against its key as well as against the caller's, so a hand-edited map cannot make one profile's record answer for another's. The whole map is read **once**, at the top of `cmd_uninstall`, before any file removal: `uninstall_skill_pack` deletes the OMH home and the profile loop runs after it, so a profile reading its own record at that point would find no `state.json` at all.

Three further rules:

- `restores` holds the value consent replaced. Only `display.interface` and `display.skin` can be replaced at all (`activate_*` under explicit consent); everything else OMH writes is unset-only, so its inverse is a removal. The most recent replacement wins: `ensure_tui_interface` refuses anything the person has set, so a second entry means they chose something new and then re-consented, and handing back the value from before the first install would restore a choice they had already abandoned.
- A managed container the reversal emptied is dropped, with its appended blank-line separator. This holds with no record: everything in it was OMH's, or it would still have a child. A container the person already kept empty is compared before and after and left exactly as it was.
- Every remover takes the shared refusal (`section_edit_guard`), and a row's status is read off the config rather than matched against its own English message. Both are review findings; see Risk.

- `src/install/config_adapter.py` — inverse writers under the same guard as their forward twins: `remove_plugin_enabled`, `revert_display_scalar`, `remove_display_sections`, `remove_memory_provider`, plus `config_container_paths` / `childless_containers` / `remove_childless_containers` and the now-shared `section_edit_guard`.
- `src/install/compression_defaults.py` — `compression_fallback_chain_lines` (verbatim) and `remove_compression_fallback_chain` (verbatim match or refuse), beside the writer that derives the chain.
- `src/install/config_reversal.py` (new) — the record, the per-key decision, and the reversal rows.
- `src/commands/setup.py` — the apply pass records; `_unregister_and_reverse` is the one path the primary and every profile take; `_print_uninstall_config_keys` names what was left behind.
- `src/commands/language.py` — five new strings in en/ko/ja/zh.
- `docs/INSTALLATION.md` — a new section saying what a full uninstall takes back, and the profile paragraph corrected to include the config reversal.

`config_message` keeps its exact prior text, because the summary line renders it through `_config_change_label`, which maps the whole message to a translation key. Everything the reversal did is reported through the new `config_keys` rows.

### Affected Surfaces

- [ ] Hermes chat or wrapper
- [ ] Codex
- [ ] Claude Code
- [ ] Generic executor
- [ ] Not executor-specific

## Validation

- [x] `PYTHONPATH=tests uv run python -m unittest discover -s tests` (one pre-existing unrelated failure, see Observed Evidence)
- [x] `git diff --check`
- [ ] `omh --help`
- [ ] Relevant workflow-learning review queue smoke: not a learning contract
- [x] Relevant dry-run or smoke command: `omh uninstall --dry-run` against a temp home with a bot profile, both configs byte-identical afterwards
- [x] Manual Hermes/TUI check: Hermes 0.21.3 startup behaviour observed against a temp `HERMES_HOME` (table above), plus the three corrections read off the host source

### Observed Evidence

- `tests/test_config_reversal.py` (new): 38 cases, `Ran 38 tests ... OK`.
- `tests/test_cli.py`: 8 new cases, `Ran 367 tests in 127.771s ... OK`. Among them: install into a temp home → uninstall → byte-diff against the pre-install config; the same with a bot profile, where **both** configs come back byte-identical; a profile whose `memory.provider` the person changed, reported `partially_cleared` with `config_keys_kept: ["memory.provider"]`.
- **Full local suite**, run once machine load dropped below the contention threshold: `Ran 13446 tests in 1223.380s` / `FAILED (failures=1, skipped=16)`. The one failure is `test_test_resource_cleanup_policy.TestResourceCleanupPolicyTests.test_no_unregistered_temporary_directory_sites`, naming two `TemporaryDirectory` sites in `tests/test_toolcall_rules.py` (lines 409 and 539). **Was pre-existing and not this branch's**: neither file differs from `origin/main`, and the same failure with the same two sites reproduced on a clean `git archive origin/main` checkout at the time. It was main's own, from the gate #1754 added meeting the two test classes #1748 added, and #1766 has since fixed it (`1fa4cee0`). This branch is now rebased onto that, and `tests/test_test_resource_cleanup_policy.py` passes here: `Ran 5 tests ... OK`.
- Mutation battery, 22 mutations, all caught. Eleven on the original mechanism (unchanged from the first push) and eleven on the review fixes:

  | mutation | caught by |
  | --- | --- |
  | memory remover loses its guard | the dotted-`memory.provider` damage case |
  | plugins remover loses its guard | the anchored-list damage case |
  | compression remover loses its guard | the anchored-document case |
  | the shared guard drops its dotted-key clause | the dotted-`memory.provider` damage case |
  | the quoted inline list is rewritten anyway | the quoted-inline case |
  | status goes back to matching the message text | the deleted-section-child case |
  | the plugins loop discards a refusal again | the anchored-list damage case |
  | profiles stop reversing | 2 CLI cases |
  | the record is re-read after the store is deleted | 2 CLI cases |
  | `partially_cleared` collapses back to `cleared` | the changed-profile case |
  | the partially-kept line is not printed | the terminal case |

  Two of these needed their fixture corrected rather than their guard. A BOM defeats the compression chain *reader*, so that case passed with or without the guard; replaced with an anchored document where the chain reads back verbatim. And the round-2 review reproduced the status-from-text mutation surviving both suites: fixing the guard parity made the round-1 example unreachable through its fixture, since a dotted `memory.provider` now refuses at the guard with a message containing none of the six fragments, and the other cited case calls `config_names_key` directly. The fixture that does separate them is now pinned — OMH wrote `display.sections.tools`, the person deleted that line, and the remover answers "no display.sections child to remove", which reads like absence and is not.
- All nine docs byte gates pass, plus `docs claims --check --json`.

- `harness validate`, `release checklist`, `release hermes-smoke`: no harness, release or smoke contract changed.
- Windows: not run locally. Measured instead — `read_config` opens in universal-newline mode on every platform, so a CRLF `config.yaml` reaches the line editors as LF and no removed or restored line can splice a CR; `write_config` opens with `newline=""` and emits LF. The consequence, which predates this change, is that any OMH write converts a CRLF config to LF. `tests/test_config_reversal.py::NewlineHandlingTests` pins the whole path end to end.

## Risk

The reversal edits a person's `config.yaml` with a hand-rolled block-YAML editor. Four things bound it:

1. **Every remover takes the same refusal.** `section_edit_guard` is one list — BOM, non-mapping or multi-document YAML, a sequence at the section, quoted mapping keys, YAML node properties, duplicate or noncanonical sections, and the dotted form of the key — parameterised by section. The first version of this PR claimed that and did not do it: the list was `_display_edit_guard`, and the memory, plugin and compression removers consulted none of it. Two of the listed shapes were not merely untouched by those removers, they were damaged, and both are now refusal tests:
   - a dotted `memory.provider: omh` beside a `memory:` block made the owner check read the dotted value and the mutator delete the block's `provider: honcho`, leaving OMH's marker and losing the person's provider;
   - `enabled: &plist` was emptied under its own anchor, so every `*plist` alias in the file resolved to null.
2. **No key is reversed unless its current value equals the recorded one**, and the mutation battery proves each compare is load-bearing.
3. **Nothing is deleted that OMH did not put there.** A `config.yaml` that reversal leaves empty is left in place, whoever created it. It is usually a file setup created, but not always: a person-authored `display:` with only `skin: omh`, or `memory:` with only `provider: omh`, empties too even with no record, because those two values name OMH by construction. Either way OMH does not delete a file in the person's Hermes home.
4. **The report says what it did, from the file.** A row's status is derived from whether the key is still named in the config, not by searching its own English message for six fragments. The old test reported a key that was present and deliberately untouched as absent, and coupled the absent/kept distinction to message wording.

The residual risk is the container cleanup, which is the one rule that acts without a record. It removes a managed container only when the reversal itself emptied it, which means everything in it was OMH's.

Two known imprecisions. A person holding `memory.provider: ''` (explicitly empty) before installing gets the line removed rather than restored to `''`, because the reader cannot distinguish an empty value from an absent key; the two are identical to Hermes. And the config write lands before every file-removal step, as it always has, so a failure in one of those leaves a reversed config beside a bundle still on disk — the safer of the two directions, now written down beside the write.

## Compatibility / Rollout

No migration. An install made before this change has no record and gets the subset in the table above, reported rather than guessed. The record appears on the next `omh setup` or on the next `omh update` that refreshes the registration.

## Release / Claims

- [x] Release-channel impact considered (`local`; no release surface changed)

## Follow-Up

Raised by the round-2 review and deferred, with reasons:

- **The widened guard over-refuses, and its message names the wrong construct.** `_contains_potential_quoted_mapping_key` fires on any quote after a `,` or `{`, so one `servers: ["a", "b"]` anywhere in the file makes every managed key report `kept`. The direction is safe and each key is named, and this was already true for `display.skin` before this PR; what changed is that three more keys join it. Two wording problems follow: the message says "quoted YAML mapping keys" when the trigger can be a quoted list *item*, and the new docs section does not list "the config is in a shape OMH will not edit" as a third reason a key is not reversed. Both clauses predate this change.
- **The no-record path leaves two null containers OMH created.** On a pre-#1725 install, `skills.external_dirs` and `plugins:` are left as null-valued keys. Two distinct causes: `empty_before` is computed after the external-dirs removal has already emptied that container, so it looks like one the person kept empty; and `plugins` is not yet childless when the emptied set is computed, so `remove_childless_containers` is never handed it. With a record both are in `containers_created` and cleaned, which is why the primary path looks clean. Not data loss, and limited to installs predating the record, but a real gap — worth a fast follow rather than widening this PR at merge time.

Deferred from the round-1 review, with reasons:

- **Localized `detail` (FU6).** The four new template strings are translated; the `{detail}` they interpolate is an untranslated `ConfigChange.message`. Translating it means ~15 remover messages across four languages, which is a change with its own review, and `tr` raises on a missing key so the existing `_config_change_label` cannot be reused as a partial fallback. The key and the status a person acts on are localized today.
- **`_parse_inline_list` (FU8, root cause).** This PR refuses a quoted inline `plugins.enabled` rather than mangling it. The parser itself still splits on commas before stripping quotes, and `ensure_plugin_enabled` reaches it on the write side. Worth its own issue against the parser.
- **`--remove-files` (FU10).** That legacy scope deletes `~/.omh` while keeping the plugin bundle and every config key, because `remove_all` is false. Consistent with the scope's own definition, and the gap between it and the default is now seven keys wide rather than zero.
- `write_config` still rewrites the whole file with no lock and no signature compare, which is #1742. The PR stacked on this one converts every call site, including the ones this PR adds.

Closes #1725


sionic-khope:
Force-pushed once after opening: the commit was amended to key the write record by config path. Every Hermes bot profile shares the primary's OMH home (`_profile_clone` sets `clone.omh_home = str(primary.omh_home)`), so one `omh setup` runs the apply pass once per Hermes home and every run writes the same `runtime/state.json`. With a single slot the run ended holding the last profile's record and the primary silently fell back to the no-record subset. Found while converting the call sites for the stacked PR, pinned by `test_one_setup_run_keeps_a_record_per_hermes_home_in_one_store` and a CLI case with a bot profile, and mutation-proved. bd30aa2a -> c4cc3a27.

sionic-khope:
Force-pushed with both must-fixes and six of the ten follow-ups. c4cc3a27 -> abe02e4c.

**MF1.** `_uninstall_hermes_profiles` now goes through the same `_unregister_and_reverse` as the primary, with each profile's own record, its own config path and its own rows. Reproducing your case: temp home with `profiles/bot/`, `setup --with-plugin --yes`, then `uninstall` — both configs come back byte-identical to pre-install. A second defect surfaced while fixing it: `uninstall_skill_pack` deletes the OMH home and the profile loop runs after it, so a profile reading its own record at that point found no `state.json` and fell back to the no-record subset. The map is now read once at the top of the command and passed down; `record` is a required keyword so there is no path that re-reads it. Both are mutation-proved.

**MF2.** `section_edit_guard` is now one list parameterised by section, and `_display_edit_guard` is a one-line call to it, so the messages and order are unchanged for the display path. `remove_memory_provider`, `remove_plugin_enabled` and `remove_compression_fallback_chain` all take it, including the dotted-key clause. Both damage cases are refusal tests with your fixtures verbatim. The Risk section is restated to say what is now true and to name both reproductions.

**FU8 treated as must-fix**, as you suggested: it is a person-value rewrite. `remove_plugin_enabled` refuses an inline list carrying quotes rather than rewriting it, so `["a,b", omh]` is left alone instead of coming back as two entries. The root cause in `_parse_inline_list` is listed as follow-up, not fixed here.

**FU4** fixed, and it turned out to be load-bearing for MF2 rather than cosmetic: with the new guards in place, `_reads_absent` classified every guard refusal as `absent` and printed nothing. A row's status is now read off the config through `config_names_key`, which is deliberately looser than the writers' readers — a dotted `memory.provider: omh` is present even though `display_interface_selection`-style readers return `""` for shapes they refuse. The plugins loop also discarded refusals by seeding itself with "not in plugins.enabled"; fixed and pinned.

**FU3** profile rows carry `config_keys`, and a profile that kept one reports `partially_cleared` with `config_keys_kept`, not `cleared`. **FU5** the terminal prints `Config key partly kept: display.sections — your value(s) for tools stayed.`; `kept` is a field on the row rather than a phrase inside `detail`, for the same reason FU4 exists. **FU7** the ordering is written down beside the write. **FU9** counts corrected: 37 unit cases, 8 CLI cases. **FU1** all three specifics corrected against the host — `LIVE_CATALOG_URL` is `hermes-agent.nousresearch.com` and I had quoted the 301 target; the six-hour `HERMES_HOME/cache` TTL means at most one request per window per home, not one per start; and `security.allow_lazy_installs: false` skips the lookup entirely. **FU2** `docs/INSTALLATION.md` gains a section on what a full uninstall takes back, and the profile paragraph now says the config reversal is included, which MF1 makes true.

**FU6 and FU10 deferred with reasons**, in the PR's Follow-Up section.

Full local suite now run, which closes your "no green full-suite evidence" note: `Ran 13446 tests in 1223.380s`, one failure — `test_test_resource_cleanup_policy` naming two `TemporaryDirectory` sites in `tests/test_toolcall_rules.py`. Pre-existing: neither file differs from `origin/main` here and it reproduces unchanged on a clean `git archive origin/main` checkout. Reported, not fixed.

Mutation battery is now 22, all caught. One of the new eleven survived its first fixture and the fixture was wrong, not the guard: a BOM defeats the compression chain *reader*, so that case passed either way. Replaced with an anchored document where the chain reads back verbatim.

sionic-khope:
Round 2 addressed. f16bce2e -> d4cfd020.

**MF-A.** You were right that the mutation table claimed something the suites did not hold. Added the fixture you verified: record `{"display.sections": ["tools"]}` against `display:\n  sections:\n    thinking: collapsed\n`. The remover answers `no display.sections child to remove`, which reads like absence about a key the config still carries — current classifier says `kept`, the deleted six-fragment matcher says `absent`. Re-ran the mutation with the exact matcher from your report: it now fails on `test_a_key_whose_own_refusal_message_reads_like_absence_is_still_kept` and nothing else. The table row now names that case instead of claiming two, and the survivor note in the body records why the two cited cases could not catch it — fixing MF2 made the round-1 example unreachable through its own fixture.

**MF-B.** The provenance clause is gone. Both of your counterexamples reproduce here (`display:\n  skin: omh\n` and `memory:\n  provider: omh\n` each reverse to `''` with no record), so the sentence now says the emptied file is left in place whoever created it, names the case where it is not setup's, and keeps the part that matters. The commit's Rejected line changed with it.

**FU-N1, FU-N2, FU-N3 listed as follow-ups with reasons**, not fixed here. FU-N2 is the one I would call a real gap rather than wording: on a pre-record install, `skills.external_dirs` and `plugins:` are left as null-valued keys, for the two distinct causes you identified — `empty_before` computed after the external-dirs removal already emptied that container, and `plugins` not yet childless when the emptied set is computed. Both look cheap (compute `empty_before` on the pre-removal text; have `remove_childless_containers` reconsider a parent after removing a child). I did not do it because it widens the change at merge time and the population is installs predating the record. It is written up with both causes so the follow-up does not have to rediscover them.

Test count is now 38 in `test_config_reversal.py`; the body says 38 and I measured 38.

PR #1765, opened 2026-09-19
fix(config): one write discipline for every OMH write of config.yaml

### LINKED ISSUE #1742: OMH has two families of config.yaml writers and only one of them checks for a concurrent change

sionic-khope:
> **Stacked on #1764** (`claude/uninstall-reverses-setup`). Its commit is the parent here; review and merge that one first. The diff below is the second commit only.

### What Changed

- Every OMH write of a person's Hermes `config.yaml` goes through one function, `update_config`: read, run the caller's mutation against the text just read, re-check the file signature immediately before the replace, retry when it moved, refuse after three rounds.
- The uninstall mutations carry the whole decision — unregistration and the per-key reversal from #1725, for the primary home and for each bot profile — so a retry re-derives what to write rather than replaying a decision made against an earlier read.
- Where the OMH home is known and exists, the CLI writers take the **same** lock the delegation route writer holds, so the two families serialize rather than merely detecting each other.
- The replacement carries the destination's file mode. The old path did not.
- `--dry-run` takes no lock at all: a preview writes nothing, so it has nothing to serialize against.
- Every refusal names its own cause: `ConfigUnwritable` for an unreadable or symlinked config, `ConfigConcurrentUpdate` for a real race, `ConfigMutationFailed` for a mutation that timed out, all under one `ConfigWriteRefused`. Any other exception from a mutation belongs to the caller and propagates unchanged.
- Four source-derived gates keep a new call site from skipping the discipline.
- The route writer's behaviour is unchanged. It gains two public one-line aliases (`config_file_signature`, `read_config_snapshot`) and one composition helper (`route_write_lock`) so the other family compares the same stamp and takes the same lock.

### Why This Exists

`install/config_adapter.write_config` rewrote the whole file with no lock and no signature compare. The route writer read, compared and replaced atomically, and since #1737 held a lock and ran from lifecycle hooks with no person present. So a `write_config` caller could read the file, have a route land, and write back its stale copy. The route is erased, and the restore record then sees a value it did not write and drops the person's baseline as a foreign edit. Before #1737 this needed somebody running a CLI command during a routed dispatch; with restores at turn and session boundaries the window is met far more often.

### User / Operator Impact

No behaviour change for a single writer: the same keys are written, with the same messages and the same `--dry-run` semantics. Under a real concurrent write the command now either retries and keeps both updates or refuses with `omh: ...` and exit 2, instead of silently erasing the other writer's change.

One change nobody asked for and everybody wanted: a `config.yaml` the person had chmodded to 0600 no longer widens to the umask default on every OMH write.

### How It Works

`update_config(path, mutate, *, omh_home, dry_run, attempts)`. `mutate` receives the text just read and returns a `ConfigChange`. It runs again on every retry, and that is what makes retrying safe rather than merely polite: the second attempt derives its change from the **other** writer's file, so neither update is lost.

Two mechanisms, not redundant:

- **The signature compare is the contract.** It holds for every writer, including one in a process that never heard of the lock. Implemented by `atomic_replace_text` in `system/local_store`, which runs the guard with the replacement already on disk as a temp file in the same directory **and already carrying the destination's mode**, so the only thing left after the guard returns is the rename. The mode is read before the temp file exists; an earlier version did the `stat` and the `chmod` between the guard and the rename while describing that window as one `os.replace`, which was not true.
- **The lock is the addition.** `route_write_lock(omh_home)` is the same `_awareness_delivery_lock` on the same restore record the route writer takes. Taken when `omh_home` names an existing OMH home.

**The lock decision, stated.** The lock lives in the OMH home and the plugin is bound to one home, while CLI commands take an explicit `--hermes-home` / `--omh-home` and may target a home the plugin is not bound to. Every CLI call site already resolves an `OmhPaths`, so the home is known and the lock is cheap to take — taken. Two things it does not cover, both deliberate and both marked in the code:

1. A home that does not exist yet. No OMH home means no plugin bound to one, so there is no route writer on the other side. `_Home` absent → `lock_enforced: False`, compare only.
2. `system/targets` writing a config that belongs to another Hermes home. Passing `paths.omh_home` there would take a lock keyed on the wrong home and give false comfort, so it passes `None` on that branch and the compare stands alone.

**Dry runs take no lock.** A preview writes nothing, so it has nothing to serialize against, and taking the lock regardless made a read-only command create `routing/` and a lock file inside it, force that directory to 0700, and acquire a five-second wait plus a way to exit 2 — none of which "the same `--dry-run` semantics" allows. A dry run reads without the lock and reports a snapshot, which is what a preview is.

**The other direction of the asymmetry**, now recorded at `route_write_lock`. The route writer waits only the lock's own 0.1 s budget and returns an error rather than blocking a dispatch, so a CLI command holding this lock longer than that drops a concurrent route write. Re-measured, mean of four runs, against that 100 ms budget:

| command | 163-line config | 800-line synthetic |
| --- | --- | --- |
| `omh apply` | 2.0 ms | 7.6 ms |
| `omh uninstall` | 3.4 ms | 43.0 ms |

`omh uninstall` is the worst case, not `omh apply`: in the full scope it runs the registration removal and then the seven-key reversal. The first version of this docstring named `apply` and quoted its numbers as the bound, which sends a future reader to measure the cheaper half. The margin is also not fixed — roughly 30x on a real config and closer to 2x at four times the size — so the recorded rule is now that adding another whole-file pass to a command that holds the lock erodes it as surely as a slow call does.

And the limit of the compare itself, which is the interesting one: it guards the window between the text `update_config` hands the mutation and the replace, and nothing earlier. A mutation that IGNORES that text and returns bytes derived from an earlier read is writing a stale copy by construction and no signature can tell, because the file has not moved since this writer read it. That is why every call site was restructured so the decision happens inside the mutation, and why one of the gates refuses a mutation that drops its argument.

| Call site | Converted | Lock | Note |
| --- | --- | --- | --- |
| `commands/setup.py` `_apply_result` | yes | `paths.omh_home` | the seven-writer chain became the mutation; the intermediate `ConfigChange`s are stashed for the payload |
| `commands/setup.py` `cmd_uninstall` | yes | `paths.omh_home` | removal and the #1725 reversal both run inside the mutation, through the same `_unregister_and_reverse` the profile loop uses |
| `commands/setup.py` `_uninstall_hermes_profiles` | yes | `profile_paths.omh_home` | the mutation is that profile's whole `_unregister_and_reverse`, so a retry re-derives its reversal too; one profile's refusal fails its own row, as its existing `except (ValueError, OmhError)` already does |
| `commands/theme.py` `_apply_theme` | yes | `paths.omh_home` | the reported `previous_skin` is re-read on a retry, so it describes the file that was actually replaced |
| `commands/memory.py` `cmd_memory_provider` | yes | `paths.omh_home` | the status-only path takes no lock, because it writes nothing |
| `system/targets.py` `record_target_observation` | yes | conditional | see the lock decision above |
| `install/self_update_state.py` bootstrap migration | yes | `paths.omh_home` | the remove/ensure chain became the mutation |

`write_config` survives for tests that build a fixture config from nothing, where there is no other writer to race. A gate fails on any use of it in `src/`.

New: `system/local_store.atomic_replace_text`, `install/config_adapter.update_config` / `ConfigUpdate` / `ConfigConcurrentUpdate` (an `OmhError`, so every CLI surface already reports it as a one-line refusal and exit 2), `tests/test_config_write_discipline.py`.

### Affected Surfaces

- [ ] Hermes chat or wrapper
- [ ] Codex
- [ ] Claude Code
- [ ] Generic executor
- [ ] Not executor-specific

## Validation

- [x] `PYTHONPATH=tests uv run python -m unittest discover -s tests` (one pre-existing unrelated failure, see Observed Evidence)
- [x] `git diff --check`
- [ ] `omh --help`
- [ ] Relevant workflow-learning review queue smoke: not a learning contract
- [x] Relevant dry-run or smoke command: `omh uninstall --dry-run` and `omh setup --yes` against temp homes through the new writer
- [ ] Manual Hermes/TUI check: not applicable; no surface Hermes renders changed

### Observed Evidence

- `tests/test_config_write_discipline.py` (new): 14 cases, `Ran 14 tests ... OK`. The acceptance case is `test_a_route_landing_between_the_read_and_the_write_is_not_erased`: a real `write_delegation_route` lands from inside the mutation, the writer retries, and afterwards the file carries both `memory.provider: omh` and the route, with `attempts == 2`.
- Mutation battery, 5 mutations, all caught:

  | mutation | caught by |
  | --- | --- |
  | a call site reverts to a bare `write_config` | `test_no_module_writes_the_config_outside_update_config` |
  | a call site drops `omh_home` | `test_every_update_config_call_names_an_omh_home` |
  | a mutation ignores the text it is handed | `test_no_mutation_discards_the_text_it_was_handed` |
  | the guard is dropped from `atomic_replace_text` | the interleave case and the refusal case |
  | the refusal after N rounds becomes a silent give-up | the refusal case |

  Eight more over the review fixes, all caught:

  | mutation | caught by |
  | --- | --- |
  | `--dry-run` takes the lock again | the dry-run case |
  | the mode is read back inside the guard-to-rename window | the mode-ordering case |
  | the timeout handler covers the whole loop again | the mutation-timeout case |
  | read errors name a race again | the symlink case |
  | a keyword `mutate=` evades the shape gate | the shape gate |
  | a renamed `write_config` import evades the call-site gate | the alias gate |
  | a write one level lower evades the call-site gate | the lower-level gate |
  | the lock is dropped entirely | the route-writer contention case |

  The last three were planted as real modules under `src/commands/`, using the reviewer's own evasion spellings.

  Seven more over the round-2 fixes, all caught:

  | mutation | caught by |
  | --- | --- |
  | the mode is read back inside the guard-to-rename window | the mode-ordering case |
  | `--dry-run` takes the lock again | the dry-run case |
  | the timeout handler is re-widened over the loop | the mutation-timeout case |
  | the backoff is removed | the injected-sleep case |
  | `ConfigUnwritable` is re-parented under `ConfigConcurrentUpdate` | the symlink case |
  | the gate resolves mutation names module-wide again | the readability gate |
  | the singular collapses back to the plural | the one-attempt case |

  Two of these had no guard at all before this round, which is why the round-2 review reproduced them surviving: the gate could not see `cmd_uninstall`'s mutation, and nothing held the backoff. Re-running the reviewer's own two mutations on `setup.py` now reports exactly one site each, the right one — dirtying only `cmd_uninstall`'s mutation reports line 2045, dirtying only the profile loop's reports line 635, where before both reported both or neither.

  Separately, the compare itself is neutralised in `test_the_compare_is_load_bearing` (the guard is answered with the signature the attempt read) and the route is then erased, which is the defect reproduced.
- The third mutation above is not hypothetical. Converting `cmd_uninstall` reintroduced exactly that shape — the new text was precomputed outside and returned from a `lambda _text: ...` — and it passed every other test in this file. The gate was written because of it.
- `tests/test_cli.py`: `Ran 364 tests in 127.232s ... OK`.
- Batch of 15 (config adapter/reversal, theme command and picker, targets, staged self-update, delegation route restore, plugin bundle standalone, journal lock portability, broad exception policy, memory provider, identity conflicts, interpreter spawn policy, architecture layout): `Ran 466 tests ... OK`.
- Batch of 12 adjacent (probe, tui widget pack, model setup flow, runtime artifacts, setup profile pack, hermes ux quality, release smoke, wrapper contract, plugin distribution, menubar app, chat router, coding retarget): `Ran 675 tests ... OK`.
- **Full local suite on the previous head** (`4dc30732`, before the round-2 fixes and the re-stack): `Ran 13466 tests in 1260.199s` / `FAILED (failures=1, skipped=16)`. The one failure was `test_test_resource_cleanup_policy`, naming two `TemporaryDirectory` sites in `tests/test_toolcall_rules.py` — main's own, from the gate #1754 added meeting the two test classes #1748 added, and reproducing unchanged on a clean `git archive origin/main` checkout at the time. #1766 has since fixed it (`1fa4cee0`); this branch is rebased onto that and the gate passes here (`Ran 5 tests ... OK`). The count is exactly #1764's 13446 plus this PR's 20 new cases.
- **After the rebase onto the reviewed #1764 tip**, re-run on this commit: `tests/test_cli.py` + `tests/test_plugin_distribution.py` `Ran 455 tests in 173.062s ... OK`; config write-discipline / reversal / adapter `Ran 84 tests ... OK`; theme, targets, staged self-update, delegation route restore, plugin bundle standalone, journal lock portability, broad exception policy, memory provider, identity conflicts, interpreter spawn policy, architecture layout, probe `Ran 376 tests ... OK`; ruff, compileall, all nine docs byte gates, `git diff --check`.
- End to end through the converted writer: `omh setup --with-plugin --yes` then `omh uninstall` on a temp home carrying a bot profile returns **both** configs byte-identical to their pre-install bytes, and the profile row reads `cleared`.
- All nine docs byte gates pass.

- (Superseded: the full suite was re-run on this commit, see Observed Evidence. Both runs predate the rebase onto `1fa4cee0`, which is what clears their one failure.) It was also run on the parent (#1764): `Ran 13446 tests in 1223.380s`, one failure, `test_test_resource_cleanup_policy` naming two `TemporaryDirectory` sites in `tests/test_toolcall_rules.py`. That failure is pre-existing and not either branch's — neither file differs from `origin/main` here, and it reproduces unchanged on a clean `git archive origin/main` checkout. Reported, not fixed.
- `harness validate`, `release checklist`, `release hermes-smoke`: no harness, release or smoke contract changed.
- Windows: not run locally. The replace goes through `_with_windows_retry`, the same jittered backoff `atomic_write_text` already uses for the WinError 5/32 sharing window, and the mode chmod goes through it too because chmod opens the file and races the same window. The temp file is created in the destination's own directory, so the rename never crosses a volume. `tests/test_journal_lock_portability.py` passes; no direct `fcntl.flock` was added, the lock is the existing `_awareness_delivery_lock`.

## Risk

The change is concentrated in one new function plus one new primitive, and every call site's decision logic moved inside a closure without being rewritten. The two behavioural changes a reviewer should weigh:

1. **A command can now refuse.** `ConfigConcurrentUpdate` is an `OmhError`, so it exits 2 with one line. That is the intended outcome — the alternative is the lost update — but it is a new way for `omh setup` to fail. Three attempts with the shared lock held make it very unlikely in practice.
2. **The mutation may run more than once.** Every converted mutation is pure with respect to the config; the ones that record something (`theme`'s `previous_skin`, `setup`'s intermediate changes, `uninstall`'s reversal rows) record the last attempt's values, which are the ones that landed. `_sync_hermes_profiles` calls `install_plugin_bundle` outside the mutation, so nothing with a side effect on disk runs twice.

`route_write_lock` creates `<omh_home>/routing/` as a side effect of taking the lock, which is why the lock is skipped when the OMH home does not exist.

## Compatibility / Rollout

No migration, no config or state format change, no new dependency. `write_config` keeps its signature for test fixtures.

## Release / Claims

- [x] Release-channel impact considered (`local`; no release surface changed)

## Follow-Up

Raised by the round-2 review and deferred, with the reason:

- **The lower-level write gate sees one spelling.** It matches when the destination expression names `hermes_config_path`, which is how every write in this repo spells it at the point of the call. A path bound to a local first — `config_path = Path(str(target["hermes_config_path"]))`, then `atomic_write_text(config_path, ...)` — escapes it, because answering that needs dataflow rather than a name. Stated in the test's own docstring and in the CHANGELOG rather than overclaimed; a reviewer still owns that spelling.

- `system/targets` writing a foreign Hermes home's config is the one call site with compare-only protection. Giving it the lock means resolving that home's OMH home the way the plugin does, which can raise inside a host; it is not attempted here.

Closes #1742


sionic-khope:
Force-pushed with all three must-fixes and follow-ups 4 to 11. 8fa0f179 -> 28d3a4b7. (#1764 had already moved to abe02e4c for its own review; this is rebased onto that with `--onto`, so only this commit replayed.)

**MF1.** `--dry-run` takes no lock. Your reproduction now shows an OMH home with nothing added after two dry runs, and with the lock held elsewhere the dry run exits 0 in 0.37 s instead of waiting 5 s and exiting 2. Pinned twice: no `routing/` created, and no wait when somebody holds it.

**MF2.** The destination mode is read before the temp file exists, so the guard-to-rename window really is the rename. Made observable rather than only restated: a destination that disappears after the mode is read still hands its mode to the replacement, which the old ordering could not do because its `stat` came after the guard and saw `FileNotFoundError`. Mutation-proved.

**MF3.** 3a closed — the mutation is collected from `mutate=` as well as positionally, and a call where neither is present fails the readability gate. Your keyword evasion, planted verbatim as a real module under `src/commands/`, now fails. 3b is not closable syntactically, so the claim is narrowed instead: the class docstring, the `update_config` docstring and the CHANGELOG all now say the gate proves the mutation NAMES its argument, which is the spelling that was reintroduced, and say plainly that a mutation which mentions its argument and still returns precomputed bytes is the same lost update and is a review question.

**FU5.** The lock test no longer compares a value to itself. It wedges the real `write_route_with_baseline` inside its own lock, in a thread, and asserts the CLI writer waits and then refuses naming that record's file. (Patched at `delegation_route_restore.write_delegation_route`, not at the defining module — the restore module binds the name at import.) Dropping the lock entirely now fails this test.

**FU7.** The `except TimeoutError` covers the acquisition only, via `ExitStack`. A mutation raising `TimeoutError("something else entirely")` now propagates as itself.

**FU8.** `ConfigUnwritable` for an unreadable or symlinked config, `ConfigConcurrentUpdate` for a real race, both under `ConfigWriteRefused`. While doing this I dropped the redundant re-raises in `theme` and `memory`: both classes are already `OmhError`, and catching to re-raise the same class only discarded the subclass that names the cause. `memory` still wraps a bare `OSError`, as it did before.

**FU9.** One paragraph at `config_file_signature`: `st_ino` carries the compare because both writer families rename, and `st_mtime_ns` matters only for a same-size third-party in-place rewrite on a coarse-timestamp filesystem.

**FU4, decided and fixed cheaply.** Documented at `route_write_lock` with your measured numbers — 1.0 ms for `omh apply` on a small config, 4.8 ms on a 400-key one, against the route writer's 100 ms — plus the rule that follows: a CLI writer must not hold this lock across a network call, a prompt or a subprocess. I did not raise the route writer's timeout, because that is #1737's behaviour and this PR does not touch it.

**FU6, decided and fixed cheaply.** Two more gates: no renamed `write_config` import, and nothing hands the Hermes config path to `atomic_write_text`, `write_text`, `write_bytes` or `open`. Both of your evasions, planted as real modules, now fail.

**FU10 and FU11** also applied: a jittered pause between rounds (documented, not timing-asserted), the exhaustion message reports the rounds actually run, and `theme`'s seed `read_config` is gone since the mutation always replaces it.

Mutation battery is 13 now, all caught, including the three planted evasions. Full suite run on this exact tree: `Ran 13466 tests in 1260.199s`, one failure — the same pre-existing `test_test_resource_cleanup_policy` one that reproduces on a clean `origin/main` checkout. 13446 + 20 new cases = 13466.

sionic-khope:
Round 2 addressed. 4dc30732 -> 2b8d0e17, re-stacked with `--onto` on #1764's new tip d4cfd020.

**MF1, the gate blind to `cmd_uninstall`.** Mutation names now resolve in their own scope: the collector carries the scope chain down, and a name binds to the innermost enclosing definition before falling back to the module. Re-ran your two mutations on `setup.py`. M7 (only `cmd_uninstall`'s mutation ignores its text) now fails, reporting `src/commands/setup.py:2045` and nothing else. M6 (only the profile loop's) reports `:635` and nothing else — the false positive on the untouched site is gone too. Pinned by a self-contained resolver test with two same-named nested functions, so the guard does not depend on `setup.py` keeping that shape; reverting the resolver to a module-wide lookup fails it.

**MF2, the wrong worst case.** Re-measured rather than adjusting the sentence, mean of four runs each:

| command | 163-line config | 800-line synthetic |
| --- | --- | --- |
| `omh apply` | 2.0 ms | 7.6 ms |
| `omh uninstall` | 3.4 ms | 43.0 ms |

`omh uninstall` is the worst case, as you found. My 800-line number is much worse than your 9.4 ms — likely a different config shape, mine is 400 provider entries so ~800 lines — and it changes the conclusion, so the docstring no longer says "more than an order of magnitude". It gives both columns and says the margin is roughly 30x on a real config and closer to 2x at four times the size. The rule that follows is now two-sided: holding the lock across anything slow breaks it, and so does adding another whole-file pass to a command that already holds it. The 163-line figure comes from a copy of a real config, read into a temp home and never written back.

**FU5, fixed rather than documented.** You were right that the CLI consequence changed. A `TimeoutError` from the mutation is now `ConfigMutationFailed`, a `ConfigWriteRefused`, so it is one `omh: ...` line and exit 2 again, and the message says the mutation rather than the lock. Deliberately narrow: only `TimeoutError`, because that is the one the old lock handler swallowed. Any other exception belongs to the caller and propagates unchanged, which `omh setup`'s `ValueError` mapping depends on; both halves are pinned.

**FU4.** The sleep is an injectable seam and a test reads the delays: three pauses for four rounds, each within its jittered window, none after the last. Deleting the sleep now fails that test; before, all twenty other cases stayed green.

**FU6.** `attempts=0` reads "during 1 update attempt".

**FU3 stated as a limit**, in the test's own docstring and in the CHANGELOG: the gate matches a destination expression naming `hermes_config_path`, which is how every write in this repo spells it at the call, and a path bound to a local first needs dataflow rather than a name. A reviewer owns that spelling.

Mutation battery on the round-2 work is 7, all caught, including the two that had no guard at all before (the gate's scope resolution and the backoff). Re-ran ruff, compileall, `git diff --check`, all nine docs gates, a 16-file adjacent batch (`Ran 514 tests OK`) and `test_cli.py` + `test_plugin_distribution.py` (`Ran 455 tests OK`). Full suite not re-run after the re-stack; it was green on the previous head apart from the failure that was main's and is now fixed underneath.

sionic-khope:
Windows failure fixed. 4133b083 -> 18e6b01f.

**Diagnosis, reproduced on macOS rather than reasoned about.** Your reading was right and here is the composition traced to the end. `read_config_snapshot` refuses a symlink by opening with `O_NOFOLLOW` and catching the `OSError`. `os.O_NOFOLLOW` is Unix-only, so on Windows `getattr(os, "O_NOFOLLOW", 0)` makes the flag a no-op, the open follows the link and succeeds, and the refusal branch is never reached. The compare's two sides then stop describing the same file: the read stamps `os.fstat` on the followed descriptor (the target) and the guard stamps `path.lstat()` (the link).

Forcing `os.O_NOFOLLOW = 0` locally reproduces it exactly, including your traceback's message:

```
lstat(link)  inode 563332520  size 84
stat(target) inode 563332519  size 11
signatures agree: False
ConfigConcurrentUpdate: config.yaml kept changing during 3 update attempts
```

So it was not the 8.3 short path; it was `fstat`-after-following against `lstat`, and the symlink refusal not firing is what let it reach the compare at all.

**Fix.** `update_config` checks `path.is_symlink()` before the read, on every round, and raises `ConfigUnwritable("config.yaml is a symlink; refusing to rewrite it")` — the same wording the reader uses, so both writer families say the same thing. Both shapes now give the identical outcome, link intact and target untouched:

```
POSIX (O_NOFOLLOW present)        -> ConfigUnwritable: config.yaml is a symlink; refusing to rewrite it
Windows shape (O_NOFOLLOW absent) -> ConfigUnwritable: config.yaml is a symlink; refusing to rewrite it
```

**On "make both sides use the same stat call"** — I did not, and the reason is recorded in the code. Aligning them by following the link on both sides would weaken the guard: a file swapped for a symlink between the read and the replace would stop being detected. The two calls only disagree for a symlink, so removing symlinks from the compare is what makes them agree without giving anything up.

**On the parent components** — only the final one is checked. The temp file is created in `path.parent` and renamed within it, so a symlinked parent cannot produce the link-destruction this prevents, and symlinked dotfile directories are a real setup. `O_NOFOLLOW` guards only the final component too, so the two families still refuse exactly the same thing.

**Pinned three ways**, all running on every platform: the POSIX case, the Windows case modelled by forcing the flag to 0, and an ordering case asserting the reader is never reached. Removing the refusal fails the Windows case and only that one — which is the point, since the POSIX case passes without it.

Re-ran ruff, compileall, `git diff --check`, all nine docs gates, `tests/test_cli.py` (`Ran 367 tests OK`) and a 14-file adjacent batch (`Ran 487 tests OK`).

PR #1772, opened 2026-09-20
fix(install): drop null skills/plugins containers on pre-record uninstall (#1767)

### LINKED ISSUE #1767: Uninstall on a pre-record install leaves skills.external_dirs and plugins as null-valued keys
## Problem

After #1764, `omh uninstall` reverses every `config.yaml` key setup wrote. On an install made BEFORE that PR (no `hermes_config_writes` record in `runtime/state.json`) the reversal is the by-construction subset, and it leaves two containers behind as null-valued keys:

```yaml
skills:
  external_dirs:

plugins:
```

Reproduced during the independent review of #1764 (round 2, follow-up N2) on a config that had no record. Two causes, both diagnosed in the PR body of #1764 so a follow-up starts from them:

1. `empty_before` is computed on the post-removal text rather than the pre-removal text, so a container OMH's own removal emptied is not recognised as OMH's to drop.
2. `remove_childless_containers` does not reconsider a parent after removing its child, so `skills:` stays once `external_dirs:` is gone.

Installs made after #1764 carry the record and are not affected.

## Direction

Compute the before-state on the pre-removal text, and make the container cleanup iterate until no childless managed container remains. Pin with the pre-record fixture from the review (no record, `skills.external_dirs` plus `plugins.enabled: [omh]`), asserting the file is byte-identical to a config that never had those sections.

## Acceptance

- The pre-record fixture leaves no null-valued `skills:` / `plugins:` keys.
- A container the person kept empty on purpose is left exactly as it was (existing test).
- Existing byte-diff round trips for recorded installs unchanged.

Frank-zhu0404:
### What Changed

- Uninstall on pre-`hermes_config_writes` installs no longer leaves null-valued `skills.external_dirs:` / `plugins:` containers.
- `reverse_managed_config` snapshots emptiness from the **pre-removal** document and iterates childless-container cleanup until stable.
- `_unregister_and_reverse` passes the pre-removal text as `before_text` into reversal.

### Why This Exists

- After #1764, pre-record uninstalls emptied managed children then treated those containers as “person-kept empty,” leaving null YAML keys (follow-up N2 from #1764 review).
- Parents were not reconsidered after a child was removed, so `skills:` could remain once `external_dirs:` was gone.

### User / Operator Impact

- Pre-#1764 Hermes homes uninstall cleanly (no null `skills`/`plugins` leftovers).
- Containers the person already kept empty stay exactly as they were.
- Recorded-install byte-diff round trips unchanged.

### How It Works

1. Capture which managed containers were empty **before** key removal.
2. Remove managed keys as today.
3. Drop childless managed containers that were **not** empty before, repeating until no further removal.
4. Leave any container that was already empty in the pre-removal document untouched.

- `src/install/config_reversal.py` — pre-removal emptiness + iterative container cleanup
- `src/commands/setup.py` — pass pre-removal text into `reverse_managed_config`
- `tests/test_config_reversal.py` — pre-record null-container fixture + person-kept-empty guard

### Affected Surfaces

- [ ] Hermes chat or wrapper
- [ ] Codex
- [ ] Claude Code
- [ ] Generic executor
- [x] Not executor-specific

## Validation

- [x] `uv run --group lint ruff check src/commands/setup.py src/install/config_reversal.py tests/test_config_reversal.py`
- [x] `PYTHONPATH=tests uv run python -m unittest tests.test_config_reversal -v` (39 tests OK)
- [ ] `uv run python -m compileall -q src tests`
- [x] `git diff --check`
- [ ] `omh --help`

### Observed Evidence

- Targeted tests, commands, CI checks, and manual behavior actually observed:
- `tests.test_config_reversal` 39 OK including `test_a_pre_record_install_leaves_no_null_skills_or_plugins_containers` and `test_a_container_the_person_already_kept_empty_is_left_exactly_as_it_was`
- ruff on touched paths: All checks passed
- `git diff --check upstream/main...HEAD`: exit 0
- Prior full suite on this knife: Ran 13556 tests; 4 pre-existing failures (3 bwrap `outside_write_exit_code`, 1 injected-context budget) reproduce on clean tree without this change

- Skipped checks and the concrete reason each one does not apply:
- Docs / harness / release / hermes-smoke CLI checks: no catalog, harness, or release-channel surface changed
- `compileall` / `omh --help`: not required for this narrow install-reversal fix; focused unittest + ruff cover the contract
- Windows; a live pre-#1764 Hermes home (pinned by the unit fixture instead)

## Risk

- Narrow uninstall / config-reversal path only. Mis-passing post-removal text as `before_text` would regress null containers again (directive called out in commit).
- Person-kept-empty containers remain protected by pre-removal emptiness comparison.

## Compatibility / Rollout

- No schema or release-channel change. Affects uninstall reversal for installs without `hermes_config_writes` records; recorded installs keep existing byte-diff behavior.

## Release / Claims

## Follow-Up

- None planned.

## Additional Notes

Joe waived Cline Muse / Codex review for this knife (2026-09-20); local preflight only.

Closes #1767


sionic-khope:
Thank you for this — and apologies for the collision. We were working on the same issue at the same time without either of us knowing: your commit is stamped `00:46:05Z` and mine `00:50:49Z`, and the two PRs were opened 38 seconds apart. Your diagnosis and mine are the same two causes, found independently, which is a good sign about the issue text.

We are landing #1770 and closing this one. Three reasons, and the third is the one that matters.

**1. Coverage.** #1770 also fixes `--registration-only`, which reaches the same helper and leaves the same litter, adds an end-to-end CLI case that drives `setup` → strip the record → `uninstall`, corrects the `docs/INSTALLATION.md` sentence that says that scope "removes the registration and nothing else", and carries a CHANGELOG entry.

**2. The loop and the widened candidate set are equivalent.** We checked rather than argued: 244,350 constructed documents through both strategies, zero differences in resulting text. The reason is structural — `config_container_paths` stops at depth two, and `remove_childless_containers` already sorts deepest-first and re-reads the mutated lines each step, so for a two-level tree one pass in that order *is* the fixpoint. Your loop is correct and terminates on every input; it just has nothing left to do on the second round. Nothing to fix here.

**3. Both of us had the same defect, and it deletes a person's data.** This is the part worth carrying away. `empty_before` answers "was this container empty when the uninstall started?" — but the question that decides whether OMH may drop it is "did OMH create it?", and on a recorded install `containers_created` answers that exactly. They disagree on one shape: a container the person already had, **empty**, before `omh setup` ran. Setup fills it, so at uninstall time it has children and is not in `empty_before`; the reversal empties it; and both of our implementations then conclude it was OMH's and delete it. Measured on a recorded install with a pre-setup config of `version: 1\n<section>:\n`:

| section | main | your PR / our first commit | after the fix |
| --- | --- | --- | --- |
| `skills` / `display` / `plugins` | kept | **deleted** | kept |
| `memory` | **deleted** | **deleted** | kept |
| nested `display:` / `sections:` | **deleted** | **deleted** | kept |

The last two rows were already wrong before either of us touched it. The fix is to stop unioning the guess on top of the record: with a record, `containers_created` is the whole answer; the guess is what the no-record lane has, and that lane is what #1767 was about.

So: the fix that ships carries your diagnosis, and the review that found the rest of it only happened because two people arrived at the same place. If you would like to keep contributing here, the issues with a reproduction in the body are the best ones to pick up, and `AGENTS.md` and `CLAUDE.md` describe what a change is expected to carry — a test that fails before it and passes after, and a PR body that reports only what was observed.

PR #1777, opened 2026-09-20
Ask a person's turn to open with what it is about to do

sionic-khope:
### What Changed

- A turn a **person** opened now carries one injected block asking the reply to
  open by saying what it understood and what it will do first. The exact string
  the model receives, in full, so it can be vetoed without reading the diff:

  ```
  [OMH Turn Opening]
  Open this reply with one or two short lines: what you understood the request to be, and the first thing you will do about it. Then continue in the same reply without waiting for an answer.
  ```

  207 characters. It rides `pre_llm_call`, the one hook that fires exactly once
  per turn, and lands inside the existing `<omh-context>` fence.

- The block is bounded: **three person-opened turns per session, then never
  again**. The count is durable, keyed by session, in the store the engagement
  nudges already write.

### Why This Exists

The Modern TUI's spinner cycles verbs — deliberating, mulling — so you can see
that the model is working and never what it is working on. A long turn is
opaque until it ends. The request was for one or two lines up front: *hold on,
for your request about X I'll go look at Y.*

The first question was whether OMH was suppressing a rule like this. It was
not, and the sweep is worth stating precisely rather than as "nothing found".
Searching `src/plugin_bundle/omh/**` and `src/skills/` for preamble, terseness
and narration instructions returns **exactly one hit**, and it is not a
suppression: the interview skill body in `src/skills/render.py` fixes the shape
of a clarifying *question* — "one sentence, no preamble, no restating what the
user just said" — inside a document the model must already have chosen to load.
Nothing always-on says anything in either direction. The behaviour was
unasked-for, not suppressed.

That one hit is also the only place this rule can collide with an OMH text, and
the collision is narrow: the skill governs the question sentence, this governs
how the reply opens, and a reply can carry an orientation line and then the
skill's header and question. **No carve-out is written for it** — a third
clause here is paid on every turn this fires, to settle a case that arises only
inside one skill body. Flagging it rather than hiding it; if you want the
carve-out, say so and it is two lines.

### User / Operator Impact

On the first three turns you type in a session, the reply should open by naming
what it understood and what it is about to do, then keep going without stopping
to ask. After that it stops being asked. Nothing changes on a turn Hermes
opened for itself, inside a delegated child, or when a caller passes
`include_omh_awareness=False`.

### How It Works

`src/plugin_bundle/omh/turn_intent_line.py` is the new module. Four gates, all
of them records, none of them the message's wording:

| Gate | Record read | Why |
| --- | --- | --- |
| `turn_opened_by_person(user_message, display_kind)` | text present + the host's own `display_kind` stamp | Hermes opens turns for its own `role="user"` rows — a background-process completion, an async-delegation batch, a model switch — and a reply to a notice has no request to restate. Presence of text cannot tell them apart; the stamp can. |
| `session_is_delegated(session_id)` | `subagent_start`'s `child_session_id` | A child answers its orchestrator, not a person. Its opening line would be read by nobody. |
| session id present | — | No id, no budget; the budget is the only thing that makes this affordable, so an unnamed session gets nothing rather than getting it forever. |
| `engagement_count(...) < TURN_INTENT_LINE_TURNS` | per-session durable counter | The cost bound. |

**The budget — shape, number, and why it can't latch.** The shape and the
number are `TODO_RECONCILIATION_FULL_TURNS`'s, next door: a small per-record
turn budget that then goes quiet. The budget is not a refinement of this
feature, it is the reason it is shippable — the audit behind
`tests/test_injected_context_pressure.py` measured one un-budgeted sentence at
14,280 characters of a single 40-turn session, and a line asking for a
*per-turn habit* is exactly the shape that would repeat that.

Three, rather than one or forever, for a reason that is specific to this
channel: Hermes replays each turn's `api_content` verbatim, so the block stays
in context after it is sent. By the fourth person turn the model is reading
three copies of the rule *and* whatever it actually did on those three turns,
which is a stronger prior than a fourth copy would be.

It **cannot latch**, and that is the difference from every neighbour here. The
plan nudge latches on a declared plan and the delegation nudge on a routed
lane, because each has a record saying the thing happened. The only thing that
would end this one is the model actually opening a reply that way — prose, which
OMH can neither read nor claim. So it decays, and the number is a cost decision
rather than a measurement. Nothing in the payload reports that a briefing was
given; `test_no_field_claims_a_briefing_happened` pins that.

It is **durable** for the measurement already recorded in `nudge_budget`: a
per-process counter hands a session a second full budget across a plugin-host
restart, which is the 2+2 delegation-nudge split measured on 2026-09-19. It is
also the cheapest durable field there — at most three writes a session, on a
hook that already writes the approval-bypass record and the awareness-delivery
claim.

**No "is this turn big enough" gate**, stated plainly because the brief asked
for one if it existed. It does not. No record available at `pre_llm_call` time
says a turn will be long; the only thing that correlates is the message's
wording, which is the inference #1549 removed from the plan surface for being
wrong in both directions, and which the first-turn primer 100 lines away
already declined for the same reason — *"judging 'big enough' is the same guess
that is already failing here, and the largest request in that set is one of the
misses."* So a one-line question does pay for this, at most three times in a
session, and the bound is the answer instead of a heuristic.

**Localisation: English, and here is what I found.** Every string this bundle
injects outbound is English. The locale packs in `routing/localization` are
inbound-only — `prepare_routing_text` normalizes the *person's message* so
triggers match in ko/ja/zh — and `src/plugin_bundle/omh/degradation.py` carries
the policy comment. This text is addressed to a model, not to a person, so it
gets no four-language treatment.

**Placement:** last inside the awareness block in `pre_llm_call`, after the plan
line, the dispatch outcomes, the active workflow and the continuation claim.
Those say what the turn's *work* is; a rule asking for a first step has to read
after them or it asks for a step chosen before the blocks that would change it.
`test_the_line_reads_after_the_blocks_that_say_what_the_work_is` pins the order.

- `src/plugin_bundle/omh/turn_intent_line.py` — new. The text, the budget
  constant, and the four gates.
- `src/plugin_bundle/omh/hooks/llm_hooks.py` — one call at the end of the
  awareness block.
- `src/plugin_bundle/omh/hooks/nudge_budget.py` — `INTENT_LINE_TURNS_FIELD`,
  added to `DURABLE_ENGAGEMENT_FIELDS`. No new map, no new store, no new reset
  seam: that module's docstring argues that a second bounded per-session map is
  a second thing a test has to remember to clear.
- `tests/test_injected_context_pressure.py` — 14 new tests; two existing
  quiet-turn contracts now spend the opening-line budget first, so "a turn with
  nothing to report costs zero" is still asserted and still true.
- `tests/test_engagement_nudges.py` — two exact-set fixtures over
  `DURABLE_ENGAGEMENT_FIELDS` and the stored row gain the fifth field.
- `tests/test_awareness_engagement.py`, `tests/test_hook_manifest.py` — three
  tests asserted `pre_llm_call(...) is None` for a turn the awareness gate
  withholds from. That is no longer the question they are asking, so they
  assert equality against `fence_omh_context([TURN_INTENT_LINE])` instead — a
  *stronger* form of what they meant: not merely that no awareness/primer
  arrived, but that nothing else did either. One is renamed
  (`..._still_gets_nothing` → `..._still_gets_no_awareness`) so the name does
  not outlive its claim.
- `CHANGELOG.md` — entry under `## Unreleased`.

No generated artifact, no skill, no routing trigger, no docs page. Every byte
gate is unchanged and re-run below.

### Affected Surfaces

- [x] Hermes chat or wrapper
- [ ] Codex
- [ ] Claude Code
- [ ] Generic executor
- [x] Not executor-specific

## Validation

- [x] `git diff --check`
- [ ] `omh --help`
- [ ] Relevant workflow-learning review queue smoke — no learning contract changed.
- [ ] Manual Hermes/TUI check — see **Not Tested** below.

Also run, each as its own command line: `docs claims --check --json`,
`docs chain-table --check`, `docs navigation --check`, `docs skill-sources
--check`, `docs ulw-inventory --check`, `docs ulw-site --check`,
`docs agent-skills --check`. All pass.

### Observed Evidence

- `tests/test_injected_context_pressure.py` — **52 tests**, 1 failure, and that
  one is `test_two_sessions_do_not_spend_each_others_budget` failing
  `'absent' != 'first_clause'`, the clock-sensitive fixture already fixed in
  PR #1773. It fails identically on `origin/main` at `f3b75107` (38 tests,
  same single failure), measured before the first edit on this branch.
- `tests/test_engagement_nudges.py`, `tests/test_plugin_bundle_standalone.py`,
  `tests/test_architecture_layout.py`, `tests/test_broad_exception_policy.py` —
  62 tests, OK.
- `tests/test_awareness_engagement.py` (17) and `tests/test_hook_manifest.py`
  (35) — OK.
- Every one of the 28 test files that exercise `pre_llm_call`, run
  individually — OK. (The first pass subsetted this and missed
  `test_hook_manifest`; the full suite caught it, and then the list was
  derived from `grep -rln pre_llm_call tests/` rather than picked.)
- **Full suite: 13,569 tests, 1 failure**, and that failure is the
  pre-existing clock fixture named above. The run before the
  `test_hook_manifest` fix had 2, which is how that one was found.
- **Failing before, by mutation** against the committed tree — each gate proved
  by removing it and naming the tests that then fail:

  | Mutation | Tests that fail |
  | --- | --- |
  | hook never appends the block | 9 of the 14 new tests, plus 1 error; the 5 that survive are the absence assertions, correctly trivial when nothing is emitted |
  | `turn_opened_by_person` → bare "text is present" | `test_a_host_written_opener_is_not_asked_to_restate_a_request` (3 subtests), `test_a_turn_nobody_opened_does_not_spend_the_budget`, and the pre-existing `test_the_ledger_file_is_not_created_by_a_turn_that_claims_nothing` |
  | budget check removed | `test_the_budget_is_spent_and_then_the_line_stops`, `test_the_budget_survives_a_plugin_host_restart`, `test_no_wording_decides_any_of_it`, and both quiet-turn cost contracts |
  | `session_is_delegated` removed | `test_a_delegated_child_is_not_asked_for_an_opening_line` |
  | field dropped from `DURABLE_ENGAGEMENT_FIELDS` | `test_the_budget_survives_a_plugin_host_restart` |

- **Measured per-turn cost**, asserted in `test_what_one_turn_of_this_costs`
  rather than quoted: 207 characters for the block; **326** for the entire
  injection on a person-opened turn that would otherwise have been quiet
  (fence + system note + block); under 1,000 for a session's worst case.
- **No wording decides anything**: `test_no_wording_decides_any_of_it` feeds
  the hook a message that is explicitly about briefings and opening lines
  ("before you start, brief me: say what you understood and what you will do
  first, in one or two lines") alongside an unrelated Korean question, and
  shows identical presence and identical budget spend.

- **Whether a model actually opens a reply this way.** No OMH surface can
  observe that; it would mean reading the reply's prose. Every assertion here is
  about what the model is *told*. This is the honest limit of the feature.
- `omh --help` and the packaged-install smoke: no CLI surface, no installed
  artifact path, and no packaging change — the new module ships with the bundle
  directory, which is copied whole.
- Manual Hermes/TUI check: not run. Reaching a live session needs `omh update`
  on the owner's machine plus a TUI restart, which is the owner's call. The
  behaviour is a model behaviour, so a live check is the only thing that can
  tell whether the wording *works* as opposed to arriving — worth doing once
  this is merged and reflected.

## Risk

- **The wording is the risk, and it is a one-line revert.** If the model reads
  "say what you understood" as an invitation to write a paragraph, or stops to
  confirm, the ask has backfired. Both are held off in the string ("one or two
  short lines", "without waiting for an answer") and neither can be verified
  from here.
- **The interview-skill collision** described above: narrow, reported, not
  carved out.
- The budget is the only thing keeping this off every turn forever. The
  directive trailer on the commit says so, and any future edit here needs its
  own bound.

## Compatibility / Rollout

- One new durable field in `$OMH_HOME/runtime/engagement-nudges.json`. Rows
  written by an older generation simply lack it and read as zero, which is a
  full budget — the fail-toward-the-previous-behaviour direction.
- No schema version bump: `omh_engagement_nudges/v1` already stores an open set
  of durable counters and the reader filters to the known field names.
- Reaches a live machine through `omh update` plus a TUI restart, as usual.

## Release / Claims

- [x] Release-channel impact considered — `local`/`stable` behaviour only; no
      release machinery touched.
- [x] Runtime/native capability claims backed by evidence or marked not observed
- [x] Prepared handoffs unaffected; nothing here is presented as evidence
- [x] Evidence contains no credentials, raw prompts, transcripts, or private data
- [x] Manual Hermes check listed above as a gap

## Follow-Up

- If the wording lands badly in a live session, the string is one constant
  (`TURN_INTENT_LINE_RULE`) and the budget one integer.
- The interview-skill carve-out, if you want it.


sionic-khope:
Closing on my own re-verification, before review.

The premise this was built on is wrong. Hermes forbids exactly this behaviour in its shipped coding brief — `agent/coding_context.py:130-131`, the last sentence of `CODING_AGENT_GUIDANCE`: "Be concise: lead with the change or answer, not a preamble." That brief is in the **stable** system tier (`agent/system_prompt.py:699-701`) on every turn of a coding-posture session, which is every TUI session rooted in a code workspace under the default `agent.coding_context: auto`.

An OMH line asking for an opening restatement would argue against a system-prompt rule from inside the user message — and from a position OMH's own fence declares subordinate: `OMH_CONTEXT_FENCE_NOTE` says the injected text is "NOT the person's words; their message outranks it." Two rules pulling opposite ways, with OMH's the one the person cannot see in the transcript.

Hermes also has an official seam for this: `agent.coding_instructions` (`agent/coding_context.py:370-380`), whose own docstring says it exists "so a user can pin workflow rules without editing the shipped brief", and which lands after the brief rather than fighting it.

The branch stays for the record. Nothing here is merged and no follow-up PR is planned; the finding goes back to the requester instead.

PR #1807, opened 2026-09-21
fix(plugin): enforce the final serialized Kanban readback ceiling

Joeywrz:
### What Changed

Enforce the Kanban readback ceiling on the **complete serialized output**:
label, newline, JSON escaping and annotation metadata included. Oversized
readbacks receive a disclosed core-field projection instead of escaping the
limit. Keep Kanban records out of subsequent diff padding, which can corrupt
JSON containing Unicode line separators. Remove quadratic front-row shifting.

### Why This Exists

Clipping source fields to 2,000 characters and subtracting dropped rows did not
ensure a final 24,000-character envelope. Quotes, backslashes and control
characters expand during JSON encoding; retained fields, unknown root extensions
and final metadata could still exceed the ceiling. A second internal issue was
that the later diff-presentation transform could split valid Unicode separators
inside JSON strings, insert literal newlines and invalidate a bounded result.

### User / Operator Impact

Large board reports no longer exceed OMH's readback contract merely because their
text is escape-heavy. When the usual reductions cannot fit, the result clearly
says that only core fields remain. Task and latest-run identity/status/outcome
remain intact for ordinary native records. Completed still means **reported done**,
not independently verified. Small payloads retain their existing behavior.

### How It Works

1. Apply existing field clipping and oldest-row reduction.
2. Render the complete label-plus-JSON envelope and measure its actual characters.
3. If it is still oversized, project fixed scalar fields, preserving the latest
   run and marking omissions. Core values are at most 256 serialized characters;
   list rows share a 12,000-character budget. IDs, states and paths that cannot fit
   are omitted explicitly rather than clipped into a different identity.
4. Exclude the three Kanban readback tools from OMH's later diff padding, including
   when the hook sees its own already-labelled result again.

The 512-character reserve remains a normal-path heuristic, not the final proof.
Rows are measured individually and removed in one slice; the implementation does
not repeatedly serialize the entire payload or shift a long list for each drop.

- `src/plugin_bundle/omh/kanban_readback.py`: complete-output measurement, bounded
  last-resort projection, truthful counters and linear row deletion.
- `src/plugin_bundle/omh/hooks/result_transforms.py`: Kanban exclusion from the
  downstream diff-presentation pass.
- `tests/test_kanban_readback_ceiling.py`: escaping, exact boundary, metadata,
  omissions/identity, Unicode diff-lookalikes, counters, idempotency and complexity.
- `tests/host_fixtures/kanban_ceiling.py`: optional native Hermes integration fixture.
- `docs/KANBAN-READBACK-CEILING.md`, `docs/README.md`: contract and reproduction.

### Affected Surfaces

- [x] Hermes chat or wrapper
- [ ] Codex
- [ ] Claude Code
- [ ] Generic executor
- [ ] Not executor-specific

## Validation

Check only commands that completed successfully. No remote CI or merge is claimed.

- [x] `git diff --check`
- [ ] `omh --help`
- [x] Relevant dry-run or smoke command: native-host fixture below.
- [ ] Manual Hermes/TUI check, or explicit reason it was not run: no live-model or
  installed-profile run authorized; real host APIs are exercised with synthetic data.

### Observed Evidence

OMH base: `d7271b208038895f50c74e9774616b1a0a26a0dd` (v2.0.4).
Native Hermes test host: `bfcd2099064715edf81786fb189d58e4f0cad176`.

- **190 scoped tests pass**, including existing readback behavior, new ceiling
  regressions, hook integrity, standalone bundle, exception policy and remote-wait
  composition checks.
- **36 native-host cases pass**, repeated by the parent verifier. Real SQLite
  create/claim/complete/comment operations and the actual `kanban_show` handler
  feed PluginContext/lifecycle hooks, direct registry dispatch and `tool_call`.
  Adversarial unknown-root/list/attachment cases explicitly substitute only
  synthetic handler output. Network and model construction are blocked.
- The same final native fixture gives **26 failures and 10 passes on the unchanged
  base**, with no setup errors. Native quote/control examples produce 29,312 /
  51,350 characters on base. Largest measured accepted candidate envelope in the
  initial native corpus is 23,641 characters; exact-24,000 cases are tested at
  the public composed hook.
- **249 tests pass on an isolated A + N1 + C combination**, and all 36 C host cases
  also pass there. A's existing equivalent unknown-read guard was retained; docs
  navigation hunks were excluded from the source integration overlay. Counts
  overlap and must not be added. This is not a merged/installed branch claim.
- Independent read-only review found one real integration gap: post-bound Unicode
  diff padding. Parent captured six failing composed-hook subcases, corrected the
  downstream exclusion and added six native regressions. Existing actual patch
  diff-padding tests still pass. No second independent review of that last guard
  is claimed.
- A + N1 + B + C also pass the selected joint regression set and all 36 C
  native cases; B's native completion cases pass on that combination as well.
  B remains independently reviewed and published, not implied accepted by C.
- Lint, compile, isolated build, offline distribution smoke and documentation
  checks for navigation/workflows/roles/capability-families/agent-skills/chain-table/
  claims/ulw-inventory/ulw-site passed.
- **Full frozen suite: 13,718 tests, 24 skipped, two failures** (exit 1, no
  timeout, 1,315.682 seconds including runner overhead). Both failures are the
  previously baseline-reproduced macOS/scratch-root assumptions:
  `test_adapter_io_accepts_trusted_macos_var_alias` and
  `test_symlink_escape_is_denied_and_symlinked_sensitive_root_rejected`.
  Their unchanged-base failures are identical; no safety guard was relaxed.
  No additional error and no distribution/cache failures occurred. Source diff
  hashes match before/after. This is explicitly not a green whole-suite claim.
  An earlier pre-review-fix run was cancelled and is not acceptance evidence.

Reproduce the core boundary:

```sh
PYTHONPATH=tests uv run python -m unittest \
  tests/test_kanban_readback.py tests/test_kanban_readback_ceiling.py -v
```

`docs/KANBAN-READBACK-CEILING.md` gives the optional native fixture procedure and
explicit source selection. Use isolated homes and no provider credentials.

- This is a **character** ceiling, not a UTF-8 byte/token limit or parser memory
  sandbox. Initial parsing/measurement still scales with input size.
- Malformed/non-object JSON and foreign tool names retain fail-open behavior.
- No live LLM, installed plugin discovery/full-bundle admission, sequential AIAgent
  conversation, production board/dispatcher, Windows local execution or remote CI.
- Release/install/generic harness validators were not run for this source-only
  correction. Workflow-learning contracts are unchanged, so its queue smoke does
  not apply.
- No universal cross-plugin cap: the pinned host invokes both result listeners on
  the same raw string and uses the first string response. Tests show OMH-first
  delivers its bounded candidate; foreign-first can deliver that plugin's larger
  result even though OMH computed a bounded candidate. Host arbitration is not
  changed by this PR.

## Risk

- Last-resort projection deliberately loses non-core detail. This is explicit in
  the label, `projection=core_fields_only`, truncation flags and omission counters.
- Oversized identity/state/path fields are omitted and named, never rewritten as
  plausible alternate identifiers. Ordinary native identifiers remain unchanged.
- Diff-looking text inside Kanban data is no longer cosmetically padded; actual
  patch/diff tools keep their existing presentation path.

## Compatibility / Rollout

No profile, schema-store migration, host-core patch or installation change.
A/#1802 and N1/#1805 remain independent PRs; the local combined checks establish
compatibility of the tested source candidates, not future upstream states.
Recheck upstream and host-plugin ordering before rollout.

## Release / Claims

- [x] Release-channel impact considered (`stable`, `preview`, or `local`): source
  contribution only, no release or installation.
- [x] Runtime/native capability claims are backed by evidence or marked as not observed.
- [x] Prepared handoffs or plans are labeled `prepared_not_observed`, never as CI or merge.
- [x] Public evidence uses synthetic records, no credentials/private prompts/transcripts.
- [x] Manual Hermes and live-model gaps are explicitly listed above.

## Follow-Up

Preserve the bound if adding another OMH transform after this seam. Cross-plugin
arbitration, HUD task selection/usage attribution and broader workflow changes
remain separate concerns; this PR does not silently implement them.


rlaope:
Merged as #1820.

Your commits were carried onto `main` with `git cherry-pick -x`, so authorship is preserved on every one of them. This was necessary rather than a rebase because the branch is on a fork with `maintainerCanModify` disabled, so a maintainer could not update it in place.

What was resolved in the carry, so nothing is hidden:
- One row in `docs/README.md`, kept both sides.

A CHANGELOG entry was added in a separate commit, since the original had none and the release notes would otherwise have omitted the change. That commit is the only one not authored by you.
Thank you for the work.

